PR #30, opened 2020-09-24
EOS-13618: Set M0D_MAX_RPC_MSG_SIZE to 512K

yeshpal-jain-cs:
Current RPC max message used is 256K, and due to this there is a limit on meta-data object
sizes.
With s3 list index count as 30, meta data KV size can't be greater than 8K.

To support larger values of meta-data size, it is proposed to change RPC msg size to 512K
and test it with LDR R1 stack for stability and performance impact.

Update following parameters on the setup to 512K,
$# cat /etc/sysconfig/motr | grep RPC
MOTR_M0D_MAX_RPC_MSG_SIZE=262144

$# cat /opt/seagate/cortx/s3/conf/s3config.yaml | grep RPC
S3_MOTR_MAX_RPC_MSG_SIZE: 262144 # Maximum rpc message size for motr, default is 131072 (128k)
$#


yeshpal-jain-cs:
> Please add explanation in the commit message about why this change is needed

added appropriate commit message, can you review it again

huanghua78 (approved):

PR #34, opened 2020-09-24
EOS-13258: m0addb2dump: fixed an iterator initialization error

alexsukhachev:
Fixed an error "m0addb2dump: Cannot initialise iterator: -71: No such file
or directory" during the parsing of rotated ADDB stob file.

There are two reasons of this error:
- incorrect detection that ADDB stob is rotated
- sometimes ADDB stob header is not correct. It is a copy of the header
  of the non-last frame while there are newer frames located at the stob.

This patch has improvements of detection that ADDB stob is rotated
and detection of last frame on the stob.

### DIFF
--- a/addb2/sit.c
+++ b/addb2/sit.c
@@ -205,7 +205,7 @@ static int it_init(struct m0_addb2_sit *it,
 	if (start != 0) {
 		result = header_read(it, h, start);
 	} else {
-		/* Search for the last frame on the stob. */
+		/* Search forward for the last frame on the stob. */
 		while (1) {
 			result = header_read(it, &header, h->he_offset + h->he_size);
 			if (result != 0 || header.he_seqno != h->he_seqno + 1)
@@ -214,6 +214,7 @@ static int it_init(struct m0_addb2_sit *it,
 		}
 
 		last_frame_end = h->he_offset + h->he_size;
+		/* Search backward */
 		while (1) {
 			result = header_read(it, &header, h->he_prev_offset);
 			if (result != 0 || header.he_seqno != h->he_seqno - 1)
--- a/addb2/sit.c
+++ b/addb2/sit.c
@@ -198,17 +198,30 @@ M0_INTERNAL int m0_addb2_storage_header(struct m0_stob *stob,
 static int it_init(struct m0_addb2_sit *it,
 		   struct m0_addb2_frame_header *h, m0_bindex_t start)
 {
-	struct m0_addb2_frame_header prev;
+	struct m0_addb2_frame_header header;
+	uint64_t                     last_frame_end;
 	int                          result;
 
 	if (start != 0) {
 		result = header_read(it, h, start);
 	} else {
+		/* Search for the last frame on the stob. */
 		while (1) {
-			result = header_read(it, &prev, h->he_prev_offset);
-			if (result != 0 || prev.he_seqno != h->he_seqno - 1)
+			result = header_read(it, &header, h->he_offset + h->he_size);
+			if (result != 0 || header.he_seqno != h->he_seqno + 1)
 				break;
-			*h = prev;
+			*h = header;
+		}
+
+		last_frame_end = h->he_offset + h->he_size;
+		while (1) {
+			result = header_read(it, &header, h->he_prev_offset);
+			if (result != 0 || header.he_seqno != h->he_seqno - 1)
+				break;
+			*h = header;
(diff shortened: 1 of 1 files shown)


addb2/sit.c
```diff
@@ -198,17 +198,30 @@ M0_INTERNAL int m0_addb2_storage_header(struct m0_stob *stob,
...
 {
-	struct m0_addb2_frame_header prev;
+	struct m0_addb2_frame_header header;
+	uint64_t                     last_frame_end;
 	int                          result;
 
 	if (start != 0) {
 		result = header_read(it, h, start);
 	} else {
+		/* Search for the last frame on the stob. */
```
huanghua78, line 208: s/Search for/Search forward for/

addb2/sit.c
```diff
@@ -198,17 +198,30 @@ M0_INTERNAL int m0_addb2_storage_header(struct m0_stob *stob,
...
+			result = header_read(it, &header, h->he_offset + h->he_size);
+			if (result != 0 || header.he_seqno != h->he_seqno + 1)
 				break;
-			*h = prev;
+			*h = header;
+		}
+
+		last_frame_end = h->he_offset + h->he_size;
+		while (1) {
+			result = header_read(it, &header, h->he_prev_offset);
```
huanghua78, line 218: Add comments:
```
/* Search backward */
```

addb2/sit.c
```diff
@@ -198,17 +198,30 @@ M0_INTERNAL int m0_addb2_storage_header(struct m0_stob *stob,
...
 	int                          result;
 
 	if (start != 0) {
 		result = header_read(it, h, start);
 	} else {
+		/* Search for the last frame on the stob. */
 		while (1) {
-			result = header_read(it, &prev, h->he_prev_offset);
-			if (result != 0 || prev.he_seqno != h->he_seqno - 1)
+			result = header_read(it, &header, h->he_offset + h->he_size);
```
huanghua78, line 210: This is search forward. How is the "rotation" handled correctly?

huanghua78:
>sometimes ADDB stob header is not correct. It is a copy of the header
>of the non-last frame while there are newer frames are located at the stob.

It is a copy of the header
of the non-last frame while there are newer frames located at the stob.

huanghua78:
Please remove the use of "err(3)" in our code, by defining a new function or macro.

As you said in the commit message:
>"m0addb2dump: Cannot initialise iterator: -71: No such file or directory"

Actually, "The  err(),  verr(),  warn(), and vwarn() functions append an error message obtained from strerror(3) based on the global variable errno"

But you see, in this case, we set the error as "-EPROTO". but the err(3) reports "No such file or directory", which is -ENOENT.
The "m0_addb2_sit_init" did not set the global errno. Actually, Motr itself does not set errno in cases of failures.
So, the following "err(3)" will print out wrong messages.


```
 341 static void file_dump(struct m0_stob_domain *dom, const char *fname)
 342 {       
...
...
 366         do {    
 367                 result = m0_addb2_sit_init(&sit, stob, offset);
 368                 if (delay > 0 && result == -EPROTO) {
 369                         printf("Sleeping for %i seconds (%lx).\n",
 370                                delay, offset);
 371                         sleep(delay); 
 372                         continue;
 373                 }       
 374                 if (result != 0)
 375                         err(EX_DATAERR, "Cannot initialise iterator: %d",
 376                             result);
...
...
386         m0_stob_destroy(stob, NULL);
387 } 

```

addb2/sit.c
alexsukhachev: search forward is not used to detect/handle "rotation". It's used for detection of last frame on the stob. Because in some cases we have in stob header copy of header of non-last frame. We must detect realy last frame on the stob because we use offset and size to detect the rotation below at the code.

addb2/sit.c
alexsukhachev: done

addb2/sit.c
alexsukhachev: done

madhavemuri:
http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/Motr/13/testReport/motr-single-node/
All tests passed except motr-single-node.47motr-tests-user-kernel.test, which is a known issue.

tejalsheth (approved):
Tested this patch on HW and it worked fine. Was able to capture the addb dumps with and without rotation  correctly.

huanghua78 (approved):

PR #38, opened 2020-09-24
EOS-13526 beck.c: filter bnodes based on generation identifier in beck tool

NikhilBirgade:
Following cases are covered:

1) If source segment header generation identifier is zero then process all the
bnodes.
2) If source segment header generation identifer is non-zero then process only
the bnodes with matching generation identifier.

### DIFF
--- a/be/tool/beck.c
+++ b/be/tool/beck.c
@@ -93,7 +93,7 @@ struct scanner {
 	 * exceed this value.
 	 */
 	m0_bcount_t	     s_max_reg_size;
-	/* Holds source metadata segment header generation identifier .*/
+	/* Holds source metadata segment header generation identifier. */
 	uint64_t	     s_gen;
 };
 
@@ -557,7 +557,7 @@ static int scanner_init(struct scanner *s)
 	int rc;
 
 	/** Initialising segment header generation identifier explicitly. */
-	s->s_seg = 0;
+	s->s_gen = 0;
 	rc = getat(s, 0, s->s_chunk, sizeof s->s_chunk);
 	if (rc != 0)
 		M0_LOG(M0_FATAL, "Can not read first chunk");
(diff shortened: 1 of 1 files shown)


be/tool/beck.c
```diff
@@ -1107,13 +1125,19 @@ static int seghdr_ver(struct scanner *s, struct rectype *r, char *buf)
...
 	int i;
+	if (gen_count >= ARRAY_SIZE(g))
+		return;
 
-	for (i = 0; i < ARRAY_SIZE(g); ++i) {
-		if (g[i].g_gen == gen || g[i].g_count == 0) {
+	for (i = 0; i < gen_count; ++i) {
+		if (g[i].g_gen == gen) {
 			g[i].g_count++;
 			break;
```
huanghua78, line 1134: Here, it can just be "return;"

be/tool/beck.c
```diff
@@ -1107,13 +1125,19 @@ static int seghdr_ver(struct scanner *s, struct rectype *r, char *buf)
...
 		}
 	}
+	if (i == gen_count) {
+		g[gen_count].g_gen = gen;
+		g[gen_count++].g_count = 1;
```
huanghua78, line 1139: It's better not to use the self increment operation here:
It is error prone. 

Let's use a separate clause:
```
g[gen_count].g_count = 1;
gen_count++;

```

be/tool/beck.c
NikhilBirgade: Done.

be/tool/beck.c
NikhilBirgade: Done.

RakeshVaghasiya (approved):

shashank-parulekar (approved):

madhavemuri (approved):

be/tool/beck.c
```diff
@@ -1108,12 +1131,18 @@ static void genadd(uint64_t gen)
...
+	for (i = 0; i < gen_count; ++i) {
+		if (g[i].g_gen == gen) {
 			g[i].g_count++;
-			break;
+			return;
 		}
 	}
+	g[gen_count].g_gen = gen;
+	g[gen_count].g_count = 1;
+	gen_count++;
```
nikitadanilov, line 1145: What was wrong with the original code?

PR #43, opened 2020-09-25
EOS-13639 be/tool/beck: handle beck tool dry run termination

mukundkanekar:
1. Added handling to gracefully terminate beck tool dry run using 'Ctrl+C'
2. Print dry run summary for scanned records.

Signed-off-by: Mukund Kanekar <mukund.kanekar@seagate.com>


be/tool/beck.c
```diff
@@ -427,6 +430,12 @@ static bool  dry_run = false;
 	M0_LOG(level, prefix " %li %s %hu:%hu:%u", s->s_off, recname(r), \
 	       (tag)->ot_version, (tag)->ot_type, (tag)->ot_size)
 
+static void  sig_handler(int num)
+{
+	printf("Cought Signal %d \n", num);
```
shashank-parulekar, line 435: Caught Signal

jamohite (approved):

shashank-parulekar (changes requested):

NikhilBirgade (approved):

be/tool/beck.c
mukundkanekar: Done.

mukundkanekar:
Jenkins job succeeded.
http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/Motr/7/

be/tool/beck.c
```diff
@@ -427,6 +430,12 @@ static bool  dry_run = false;
 	M0_LOG(level, prefix " %li %s %hu:%hu:%u", s->s_off, recname(r), \
 	       (tag)->ot_version, (tag)->ot_type, (tag)->ot_size)
 
+static void  sig_handler(int num)
```
huanghua78, line 433: remove extra space between "void" and "sig_handler".

be/tool/beck.c
```diff
@@ -515,6 +524,10 @@ int main(int argc, char **argv)
 	result = scanner_init(&s);
 	if (result != 0)
 		err(EX_CONFIG, "Cannot initialise scanner.");
+	if(dry_run) {
```
huanghua78, line 527: add a space after "if"

huanghua78: If not dry_run, then the SIGINT is not called. So CTRL+C will directly terminate this program.
Is this the expected behavior?

be/tool/beck.c
```diff
@@ -418,6 +420,7 @@ static struct builder b;
 static struct gen g[MAX_GEN] = {};
 
 static bool  dry_run = false;
+static bool  signaled = false;
 
 #define FLOG(level, s)						\
 	M0_LOG(level, "        at offset: %"PRId64" errno: %s (%i), eof: %i", \
```
huanghua78, line 426: I have raised this concern in another review. But I will raise it again here.

The `errno` (and also the strerror(errno)) will report incorrect error messages, because the `errno` is not set in any Motr functions.
For example here:

```
714 static int recdo(struct scanner *s, const struct m0_format_tag *tag,
 715                  struct rectype *r)
 716 {
 717         unsigned  size = tag->ot_size + sizeof(struct m0_format_footer);
 718         void     *buf;
 719         int       result;
 720 
 721         if (size > MAX_REC_SIZE)
 722                 return M0_RC(-EINVAL);
 723 
 724         buf = alloca(size);
 725         result = get(s, buf, size);
 726         if (result == 0) {
 727                 if (memcmp(tag, &r->r_tag, sizeof *tag) == 0) {
 728                         result = m0_format_footer_verify(buf, false);
 729                         if (result != 0) {
 730                                 RLOG(M0_DEBUG, "С", s, r, tag);
 731                                 FLOG(M0_DEBUG, s);
 732                                 r->r_stats.s_chksum++;
 733                                 if (!dry_run && r->r_ops != NULL &&
 734                                     r->r_ops->ro_check != NULL)
 735                                         result = r->r_ops->ro_check(s, r, buf);
 736                         } else {
 737                                 RLOG(M0_DEBUG, "R", s, r, tag);
 738                                 if (r->r_ops != NULL &&
 739                                     r->r_ops->ro_proc != NULL)
 740                                         result = r->r_ops->ro_proc(s, r, buf);
 741                         }
 742                 } else {
 743                         RLOG(M0_DEBUG, "V", s, r, tag);
 744                         FLOG(M0_DEBUG, s);
 745                         r->r_stats.s_version++;
 746                         if (r->r_ops != NULL && r->r_ops->ro_ver != NULL)
 747                                 result = r->r_ops->ro_ver(s, r, buf);
 748                 }
 749         }
 750         return M0_RC(result);
 751 }
```
The call of FLOG() on Line 731 and Line 744 will report incorrect error messages, because it will use `errno`, but `errno` is not set in previous calls of `m0_format_footer_verify` or `get`.
The real error number 'result' returned from previous functions is ignored.

be/tool/beck.c
mukundkanekar: In case of dry run now we are looking for printing summary of records scanned till 'Ctrl+C' is pressed.

huanghua78: Yes. I know in case of dry run, "CTRL+C" will set the flag "sig_caught" and then exit the loop, and then print out summary.

But in NON dry run (real run), the "CTRL+C" signal is not handled, and that will terminate the program run abruptly.

mukundkanekar: 1. In case of dry run the beck is invoked directly from command line, hence can receive signal. Also the dry run is read only mode and graceful termination is simpler.
2. In case of normal run the beck tool is invoked via recovery script and my not receive signal, also the normal run modifies data hence graceful shutdown will required additional logic. So the existing behavior is kept as is for normal run.

be/tool/beck.c
mukundkanekar: done.

be/tool/beck.c
mukundkanekar: > add a space after "if"

done.

mukundkanekar: As discussed in call.  We will be checking with Nikita and can be taken as PR.

huanghua78 (approved):

be/tool/beck.c
mukundkanekar: As discussed we will check with Nikita and can be taken as separate PR.

@nikitadanilov  can you please share your views on use of 'errno' in FLOG.

shashank-parulekar (approved):

PR #45, opened 2020-09-25
EOS-9668 Initial fix for invalid lgr_last_discarded corner case

tejalsheth:
Recreated the PR https://github.com/Seagate/cortx-motr/pull/100
Problem:
As per the current implementation, the discarded records are received via the AST-mechanism
in backwards order via the m0_be_group_format_discard() calls. If the lg_discarded position is moved by the records sizes , it is possible that lgr_last_discarded at m0_be_log_record_io_prepare() picks up some interim lg_discarded value between
the 1st and the last AST-calls of m0_be_group_format_discard() and write it down into the log. If it is the last log record just before the crash - the wrong discarded position will be taken during the recovery on m0d process restart resulting in a data corruption.
Solution:
Post a single AST for multiple log discard items during be_log_discard_sync_done_cb() and discard all the ldi items till ld->lds_sync_item from ld->lds_start_q in AST handler.
This will avoid the case where lgr_last_discarded can get updated with the interim lg_discarded value due to multiple ASTs being posted.

Initial fix to discard several log records at once in a single ast.
The implementation includes:

Move the for loop which posts multiple ASTs in be_log_discard_sync_done_cb() to be_log_discard_item_discard_ast()
Post just single AST in be_log_discard_sync_done_cb().
Get total discarded size in the for loop inside the ast handler and discard ldi. Call m0_be_log_record_discard() once outside the for loop to update the lg_discarded position only once.
P.S. -This is initial draft implementation for reviewing the approach and not ready for merging. UT changes to be done.

### DIFF
--- a/be/log_discard.c
+++ b/be/log_discard.c
@@ -49,7 +49,6 @@ enum be_log_discard_item_state {
 	LDI_INIT,
 	LDI_STARTING,
 	LDI_FINISHED,
-	LDI_SYNCED,
 	LDI_DISCARDED,
 };
 
@@ -59,7 +58,6 @@ struct m0_be_log_discard_item {
 	void                           *ldi_user_data;
 	m0_time_t                       ldi_time_start;
 	m0_time_t                       ldi_time_finish;
-	bool                            ldi_synced;
 	struct m0_ext                   ldi_ext;
 	struct m0_tlink                 ldi_start_link;
 	uint64_t                        ldi_magic;
@@ -114,7 +112,6 @@ M0_INTERNAL int m0_be_log_discard_init(struct m0_be_log_discard     *ld,
 		*ldi = (struct m0_be_log_discard_item){
 			.ldi_ld     = ld,
 			.ldi_state  = LDI_INIT,
-			.ldi_synced = false,
 		};
 		ld_be_pool_add(&ld->lds_item_pool, ldi);
 		ld_start_tlink_init(ldi);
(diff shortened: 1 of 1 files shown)


be/log_discard.c
```diff
@@ -230,72 +227,73 @@ static void be_log_discard_item_discard_ast(struct m0_sm_group *grp,
...
+	struct m0_ext *ext;
 
-	M0_ENTRY("ld=%p ldi=%p", ld, ldi);
-
-	ld->lds_cfg.ldsc_discard(ld, ldi);
-
+	/* if lds_sync_item is NULL all the ldi are already discarded
+	   and lds_start_q will not have any ldi with LDI_FINISHED state */
+	if (ld->lds_sync_item == NULL)
+		return;
```
max-seagate, line 236: [* defect *] It wouldn't be possible to catch log discard bugs if some asts would be silently discarded.

max-seagate (changes requested):

be/log_discard.c
tejalsheth: Working on solution to post an ast after fdatasync() (if there is no ast running already) and discard each log record sequentially in a loop in that ast.

be/log_discard.c
```diff
@@ -230,72 +227,73 @@ static void be_log_discard_item_discard_ast(struct m0_sm_group *grp,
...
 	be_log_discard_lock(ld);
-	ldi->ldi_state = LDI_DISCARDED;
-	m0_be_log_discard_item_put(ld, ldi);
+	M0_ENTRY("ld=%p ldi=%p", ld, ldi);
+	/* capture the extent and data here in temporary ldi */
+	tmp_ldi.ldi_ext = ldi->ldi_ext;
```
huanghua78, line 240: use `m0_be_log_discard_item_ext()`?

be/log_discard.c
```diff
@@ -230,72 +227,73 @@ static void be_log_discard_item_discard_ast(struct m0_sm_group *grp,
...
+		if (tmp_ldi.ldi_ext.e_end  < ext->e_end)
+			tmp_ldi.ldi_ext.e_end = ext->e_end;
+		if (ldi == ld->lds_sync_item)
+		{
+			ldi->ldi_state = LDI_DISCARDED;
+			m0_be_log_discard_item_put(ld, ldi);
+			break;
+		}
+		ldi->ldi_state = LDI_DISCARDED;
+		m0_be_log_discard_item_put(ld, ldi);
```
huanghua78, line 262: Can we have the following to avoid duplicated code?
```
                ldi->ldi_state = LDI_DISCARDED;
		m0_be_log_discard_item_put(ld, ldi);
                if (ldi == ld->lds_sync_item)
                        break;
```

be/log_discard.c
```diff
@@ -230,72 +227,73 @@ static void be_log_discard_item_discard_ast(struct m0_sm_group *grp,
...
+	} m0_tl_endfor;
+	ld->lds_sync_item = NULL;
+	ld->lds_sync_in_progress = false;
+	if (ld_start_tlist_is_empty(&ld->lds_start_q))
+		ld->lds_sync_deadline = M0_TIME_NEVER;
+	be_log_discard_check_sync(ld, false);
+	/* The log pointers (lg_discarded position) for all the discarded items will be updated only once.
```
huanghua78, line 269: Please limit the line to 80 chars.

be/log_discard.c
```diff
@@ -230,72 +227,73 @@ static void be_log_discard_item_discard_ast(struct m0_sm_group *grp,
...
+	   This will ensure that interim lg_discarded is not picked up in lgr_last_discarded
+	   during m0_be_log_record_io_prepare() */
+	ld->lds_cfg.ldsc_discard(ld, &tmp_ldi);
```
huanghua78, line 272: Please note: the `tmp_idl` only has its lid_ext and ldi_user_data valid. Others field members are not initialized. Be sure they are not used in this function.

huanghua78:
The fix is merged by #209 ?

tejalsheth:
> The fix is merged by #209 ?

Yes this fix is merged by #209

tejalsheth:
This fix is merged by #209

PR #51, opened 2020-09-25
EOS-8251: workaround fix for FOM HUNG

jadhavik1:
after reaching limit of hung message
hung_fom_notify() calls m0_fom_ready()
to move fom from waitq to runq so fom_exec
checks hung limit  and move fom phase to
M0_FOPH_FAILURE.

### DIFF
--- a/fop/fom.c
+++ b/fop/fom.c
@@ -52,6 +52,7 @@
 #include "rpc/rpc_opcodes.h"
 #include "fdmi/fol_fdmi_src.h"
 #include "motr/iem.h"
+#include "fop/fom_generic.h"
 
 /**
  * @addtogroup fom
@@ -138,6 +139,7 @@ enum {
 	HUNG_FOP_SEC_PERIOD   = 5,
 	HUNG_FOP_TIME_SEC_MAX = 2*60,
 	HUNG_FOP_TIME_SEC_IEM = 5*60,
+	HUNG_FOP_HIT_MAX_CNT  = 15,
 };
 
 /**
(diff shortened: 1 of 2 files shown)


fop/fom.c
```diff
@@ -361,6 +363,10 @@ static bool hung_fom_notify(const struct m0_fom *fom)
 		       fom->fo_fop == NULL ? 0 : m0_fop_opcode(fom->fo_fop),
```
huanghua78, line 363: s/&fom->fo_fop/fom->fo_fop/

I think the author of this function wanted to print out the pointer: fom->fo_fop
not the address of this pointer.
Otherwise, if we have `fom`, then the '&fom->fo_fop' is just the `fom+some offset`. 
Please fix it.

huanghua78: The same to line 384 in original file.

jadhavik1:
Incorrect approach so closing this.

fop/fom.c
```diff
@@ -813,25 +819,31 @@ static void fom_exec(struct m0_fom *fom)
...
+				 * call-back is not allowed to destroy a fom.
+				 */
+				M0_ASSERT(m0_fom_phase(fom) != M0_FOM_PHASE_FINISH);
+				if (fom_state(fom) != M0_FOS_WAITING)
+					break;
+			}
 		}
+
+		if (fom->fo_hung_cnt >= HUNG_FOP_HIT_MAX_CNT)
+			m0_fom_phase_move(fom, M0_ERR(-EAGAIN), M0_FOPH_FAILURE);
```
huanghua78, line 845: But I have a concern here:
Some of FOM state are not allowed to move to M0_FOPH_FAILURE.
This is configured in fom sm transition table.

For example, the following sate:
```
        [M0_FOPH_TXN_COMMIT_WAIT] = {
                .sd_name      = "tx_commit_wait",
                .sd_allowed   = M0_BITS(M0_FOPH_TXN_INIT, M0_FOPH_FINISH)
        },
```

and this one:
```
        [FOPH_RM_REQ_WAIT] = {
                .sd_name    = "RM_Request_Wait",
                .sd_allowed = M0_BITS(FOPH_RM_REQ_FINISH)
        },
```

And these two fom states were reported to be HUNG in the following ticket:
```
https://jts.seagate.com/browse/COSTOR-898?focusedCommentId=1455506&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-1455506
```

So we need to make sure this state transition is allowed.

jadhavik1: Agree with comment, this was incorrect approach.

PR #58, opened 2020-09-29
EOS-13177: bug fix in error_injection.py to create holes in metadata

gaurkrgaurav:
1. To keep search record if no of error induced is greater than available records
2. Added Seed for generating random number
3. loggin start-stop time
4. Skip if valid type not found in the metadata
5. Miscellaneous small fixes

Signed-off-by: Gaurav Gaur <gaurav.gaur@seagate.com>

### DIFF
--- a/scripts/beck/st/README
+++ b/scripts/beck/st/README
@@ -30,14 +30,15 @@ How to run the script:
    On srvnode-2 : /dev/vg_metadata_srvnode-2/lv_raw_metadata
 
 2. Whenever you need to induce holes in metadata, you need to first shutdown the cluster
-   On HW you can use "pcs cluster stop --all" and on VM "hctl shutdown"
+   On HW you can run "pcs cluster stop --all" and on VM "hctl shutdown"
    
 3. It will create a log named as hole_creation_YYYYDDMM_HHMMSS.log
 
 4. For injecting error on random record with user given number of error.
    example: error_injection.py -rn -e <number of error> -m <metadata path>
    
-   It will print seed as well for random number generation. User can use that seed for same hole injection
+   It will print seed as well for random number generation. User can take that seed for 
+   initialize the "random" library: to initialize the random generation 
    
 5. For injecting error on particular record type with user given number of error.
    example: error_injection.py -rt <record type> -e <number of error> -m <metadata path>
@@ -56,10 +57,10 @@ How to run the script:
 9. For injecting 512k bytes errors in metadata. The starting point in error injection will be random.
    example: error_injection.py -512k -m <metadata path>
    
-10. For injecting huge errors in metadata, Means script will randomly take a starting point and start corrupting till end of metadata.
-    example: error_injection.py -512k -m <metadata path>
+10. For injecting huge errors in metadata, script will randomly take a starting point and start corrupting till end of metadata.
+    example: error_injection.py -huge -m <metadata path>
 	
-11. Count all record numbers. This is use for verifying record numbers.
+11. Count all record numbers. This is useful for verifying record numbers.
     example: error_injection.py -v -m <metadata path>
 	
 12. Use -seed for recreation of any random scenario which is -rn, -gmd, -dmd, -a, -512k, -huge.
--- a/scripts/beck/st/README
+++ b/scripts/beck/st/README
@@ -1,0 +1,65 @@
+The script will create hole in metadata as per user choice. There are n number of ways to create hole using error_injection.py script.
+User can check the option using "error_injection.py -help".
+
+***** Script Started *****
+Date Time
+usage: error_injection.py [-h] [-rn] [-e NOOFERR] [-rt RECORD_TYPE] [-m FILE]
+                          [-v] [-a] [-gmd] [-dmd] [-512k] [-huge] [-seed SEED]
+
+Basic Arguments to run the script
+
+optional arguments:
+  -h, --help       show this help message and exit
+  -rn              For inducing error at Random place
+  -e NOOFERR       How Many number of error do you want to induce in Metadata
+  -rt RECORD_TYPE  Record Type For inducing error at perticular record like
+                   BE_BTREE, BE_EMAP, CAS_CTG etc
+  -m FILE          Metadata Path
+  -v               Read full Metadata and Print all the Records entry counts
+  -a               Induce Error in All Record at Random place
+  -gmd             Induce Error in All GMD type of Record at Random place
+  -dmd             Induce Error in All DMD type of Record at Random place
+  -512k            Induce 512K bytes error in Metadata
+  -huge            Induce Huge amount of corruption in Metadata
+  -seed SEED       Seed for generating Random errors corruption
+
+How to run the script:
+1. Find the location of your metadata and note it down.
+   Usually it looks like this:
+   On srvnode-1 : /dev/vg_metadata_srvnode-1/lv_raw_metadata
+   On srvnode-2 : /dev/vg_metadata_srvnode-2/lv_raw_metadata
+
+2. Whenever you need to induce holes in metadata, you need to first shutdown the cluster
+   On HW you can use "pcs cluster stop --all" and on VM "hctl shutdown"
+   
+3. It will create a log named as hole_creation_YYYYDDMM_HHMMSS.log
+
+4. For injecting error on random record with user given number of error.
+   example: error_injection.py -rn -e <number of error> -m <metadata path>
+   
+   It will print seed as well for random number generation. User can use that seed for same hole injection
+   
+5. For injecting error on particular record type with user given number of error.
+   example: error_injection.py -rt <record type> -e <number of error> -m <metadata path>
+   
+   Record types are BE_BTREE, BE_EMAP, CAS_CTG, BE_BNODE, BALLOC etc.
+   
+6. For injecting error on all global metadata type records with random number of error.
+   example: error_injection.py -gmd -m <metadata path>
+   
+7. For injecting error on all dynamic metadata type records with random number of error.
+   example: error_injection.py -dmd -m <metadata path>
+   
+8. For injecting error on all metadata type records with random number of error.
+   example: error_injection.py -a -m <metadata path>
+   
+9. For injecting 512k bytes errors in metadata. The starting point in error injection will be random.
+   example: error_injection.py -512k -m <metadata path>
+   
+10. For injecting huge errors in metadata, Means script will randomly take a starting point and start corrupting till end of metadata.
+    example: error_injection.py -512k -m <metadata path>
+	
+11. Count all record numbers. This is use for verifying record numbers.
+    example: error_injection.py -v -m <metadata path>
+	
+12. Use -seed for recreation of any random scenario which is -rn, -gmd, -dmd, -a, -512k, -huge.
--- a/scripts/beck/st/error_injection.py
+++ b/scripts/beck/st/error_injection.py
@@ -12,14 +12,14 @@ import logging
 timestr = time.strftime("%Y%m%d-%H%M%S")
 log_filename = "hole_creation_"+timestr+".log"
 
-logger=logging.getLogger() 
+logger=logging.getLogger()
 logger.setLevel(logging.DEBUG)
 
 fh=logging.FileHandler(log_filename)
-fh.setLevel(logging.DEBUG) 
+fh.setLevel(logging.DEBUG)
 
 ch=logging.StreamHandler()
-ch.setLevel(logging.DEBUG) 
+ch.setLevel(logging.DEBUG)
 fformatter = logging.Formatter('%(asctime)s - %(levelname)s - %(message)s')
 cformatter = logging.Formatter('%(levelname)s : %(message)s')
 fh.setFormatter(fformatter)
@@ -33,7 +33,7 @@ parser = argparse.ArgumentParser(description="Basic Arguments to run the script"
 parser.add_argument('-rn', action='store_true', default=False, dest='random', help='For inducing error at Random place')
 parser.add_argument('-e', action='store', default= 0, type=int, dest='noOfErr', help='How Many number of error do you want to induce in Metadata')
 parser.add_argument('-rt', action='store', dest='Record_Type', help='Record Type For inducing error at perticular record like BE_BTREE, BE_EMAP, CAS_CTG etc')
-parser.add_argument('-m', action='store', dest='file', help='Metadata Path')
+parser.add_argument('-m', action='store', dest='mfile', help='Metadata Path')
 parser.add_argument('-v', action='store_true', default=False, dest='verify', help='Read full Metadata and Print all the Records entry counts')
 parser.add_argument('-a', action='store_true', default=False, dest='allErr', help='Induce Error in All Record at Random place')
 parser.add_argument('-gmd', action='store_true', default=False, dest='allGMD', help='Induce Error in All GMD type of Record at Random place')
@@ -43,13 +43,12 @@ parser.add_argument('-huge', action='store_true', default=False, dest='hugeCorru
 parser.add_argument('-seed', action='store', default=0, type=float, dest='seed', help='Seed for generating Random errors corruption')
 
 args = parser.parse_args();
-#logger.info("Random {0} No of Errors {1} Record Types {2} File name {3} Verify {4}".format(args.random, args.noOfCorr, args.recordType, args.filename, args.verify ))
 
 results = parser.parse_args()
 logger.info('Induce Random Error        = {!r}'.format(args.random))
 logger.info('No of Error induce         = {!r}'.format(args.noOfErr))
 logger.info('Record Type                = {!r}'.format(args.Record_Type))
-logger.info('Metadata file path         = {!r}'.format(args.file))
+logger.info('Metadata file path         = {!r}'.format(args.mfile))
 logger.info('Verify Record entries      = {!r}'.format(args.verify))
 logger.info('Induce Error in All Record = {!r}'.format(args.allErr))
 logger.info('Induce Error in GMD Record = {!r}'.format(args.allGMD))
@@ -58,12 +57,12 @@ logger.info('Induce 512k errors         = {!r}'.format(args.err512k))
 logger.info('Induce huge corruption     = {!r}'.format(args.hugeCorruption))
 logger.info('Seed for random number     = {!r}'.format(args.seed))
 
-filename  = args.file
+filename  = args.mfile
 recordType= args.Record_Type
 noOfCorr  = args.noOfErr
 
 if args.seed != 0:
-	seed = args.seed()
+	seed = args.seed
 	logger.info("Seed used: {}".format(seed))
 else:
 	seed = time.time()
@@ -160,10 +159,10 @@ def recordOffset(record, i, size):
 
 def ReadTypeSize(byte):					#Ex: 0001(ver) 0009(type) 00003dd8(size) 
 	ver=byte[:4]						#.ot_version = src->hd_bits >> 48,
-	type=byte[6:8]						#.ot_type    = src->hd_bits >> 32 & 0x0000ffff,
+	rtype=byte[6:8]						#.ot_type    = src->hd_bits >> 32 & 0x0000ffff,
 	size =byte[8:16]					#.ot_size    = src->hd_bits & 0xffffffff
-	#logger.info("Version {}, Type {}, Size {}".format(ver, type, size))
-	return type,size
+	#logger.info("Version {}, Type {}, Size {}".format(ver, rtype, size))
+	return rtype,size
 
 def EditMetadata(offset):
 	with open(filename, 'r+b') as wbfr:
@@ -297,7 +296,7 @@ def InduceErrinDMDRecords():
 def induceHugeError():
 	count = 0
 	with open(filename, 'r+b') as wbfr:
-		logger.info("** Corrupting 8byte of Metadata at offset {} with b'1111222244443333' **".format(hex(offset)))
+		logger.info("** Corrupting 8byte of Metadata with b'1111222244443333' all place")
 		wbfr.seek(-1, os.SEEK_END)
 		endoffset=wbfr.tell()
 		offset = random.randint(1, endoffset)
@@ -357,7 +356,6 @@ def read_metadata_file():
 		#if os.path.getsize(filename):
 		i=0
 		while 1:
-			flag=0
 			byte=metadata.read(8)
 			i=i+8
 			if not byte:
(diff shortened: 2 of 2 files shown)


ketanarlulkar (approved):

ravindrachoudhari (approved):

scripts/beck/st/error_injection.py
```diff
@@ -1,66 +1,78 @@
-#!/usr/bin/env python3
-#
-# Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
```
huanghua78, line 3: Why is the copy right herder removed?

scripts/beck/st/README
```diff
@@ -0,0 +1,65 @@
...
+How to run the script:
+1. Find the location of your metadata and note it down.
+   Usually it looks like this:
+   On srvnode-1 : /dev/vg_metadata_srvnode-1/lv_raw_metadata
+   On srvnode-2 : /dev/vg_metadata_srvnode-2/lv_raw_metadata
+
+2. Whenever you need to induce holes in metadata, you need to first shutdown the cluster
+   On HW you can use "pcs cluster stop --all" and on VM "hctl shutdown"
+   
+3. It will create a log named as hole_creation_YYYYDDMM_HHMMSS.log
```
huanghua78, line 35: "It"?
Do you mean, this script will created a log when executing?

scripts/beck/st/README
```diff
@@ -0,0 +1,65 @@
...
+
+4. For injecting error on random record with user given number of error.
```
huanghua78, line 37: s/user given number of error/user provided number of error/

scripts/beck/st/README
```diff
@@ -0,0 +1,65 @@
...
+   example: error_injection.py -rn -e <number of error> -m <metadata path>
+   
+   It will print seed as well for random number generation. User can use that seed for same hole injection
+   
+5. For injecting error on particular record type with user given number of error.
+   example: error_injection.py -rt <record type> -e <number of error> -m <metadata path>
+   
+   Record types are BE_BTREE, BE_EMAP, CAS_CTG, BE_BNODE, BALLOC etc.
```
huanghua78, line 45: > Record types are BE_BTREE, BE_EMAP, CAS_CTG, BE_BNODE, BALLOC etc.
Can you explain which is gmd? and which is dmd?

scripts/beck/st/README
```diff
@@ -0,0 +1,65 @@
...
+7. For injecting error on all dynamic metadata type records with random number of error.
+   example: error_injection.py -dmd -m <metadata path>
+   
+8. For injecting error on all metadata type records with random number of error.
+   example: error_injection.py -a -m <metadata path>
+   
+9. For injecting 512k bytes errors in metadata. The starting point in error injection will be random.
+   example: error_injection.py -512k -m <metadata path>
+   
+10. For injecting huge errors in metadata, Means script will randomly take a starting point and start corrupting till end of metadata.
```
huanghua78, line 59: s/Means//

scripts/beck/st/README
```diff
@@ -0,0 +1,65 @@
...
+    example: error_injection.py -512k -m <metadata path>
```
huanghua78, line 60: typo:

s/-512k/-huge/

scripts/beck/st/README
```diff
@@ -0,0 +1,65 @@
...
+	
+11. Count all record numbers. This is use for verifying record numbers.
```
huanghua78, line 62: s/use/useful/

scripts/beck/st/README
```diff
@@ -0,0 +1,65 @@
...
+    example: error_injection.py -v -m <metadata path>
+	
+12. Use -seed for recreation of any random scenario which is -rn, -gmd, -dmd, -a, -512k, -huge.
```
huanghua78, line 65: What purpose is the seed used for?
Please explain more.

ravindrachoudhari (approved):

scripts/beck/st/README
```diff
@@ -0,0 +1,65 @@
+The script will create hole in metadata as per user choice. There are n number of ways to create hole using error_injection.py script.
```
huanghua78, line 1: Please explain, "create hole", "error injection".
What is the difference?
Why do you say "create hole?"
Do you mean to insert a segment with all zeroes?
"error injection"? What error is going to injected into the metadata file?
Just some array of random bytes with random size?
Or some array of all zeroes with random size?
At any offset in the metadata file?
or it will overwrite some records with random bytes? At the record boundary?

Please explain all these topics.

scripts/beck/st/error_injection.py
```diff
@@ -1,66 +1,78 @@
...
-#
+#!/usr/bin/python3
 import binascii
 import sys
 import os
 import random
 import argparse
+import time
+import logging
+#import pdb
```
huanghua78, line 27: Please remove unnecessary code.

ketanarlulkar (approved):

scripts/beck/st/README
gaurkrgaurav: Yes. at same directory

scripts/beck/st/error_injection.py
gaurkrgaurav: Those are debug purpoase

scripts/beck/st/error_injection.py
```diff
@@ -1,66 +1,78 @@
...
-print('Induce Error in All Record = {!r}'.format(args.allErr))
-print('Induce Error in GMD Record = {!r}'.format(args.allGMD))
-print('Induce Error in DMD Record = {!r}'.format(args.allDMD))
-print('Induce 512k errors         = {!r}'.format(args.err512k))
-print('Induce huge corruption     = {!r}'.format(args.hugeCorruption))
-
-filename  = args.file
-recordType= args.recordT
+logger.info('Induce Random Error        = {!r}'.format(args.random))
+logger.info('No of Error induce         = {!r}'.format(args.noOfErr))
```
huanghua78, line 67: s/No of Error induce/Number of error induce/

No need to abbreviation.

scripts/beck/st/error_injection.py
```diff
@@ -1,66 +1,78 @@
...
+logger.info('Record Type                = {!r}'.format(args.Record_Type))
```
huanghua78, line 68: Record_Type?
The name of this field is using different naming rules.
recordType?

Also, "Record_Type" has a different color here in the PR. Why? Is it a keyword? a reserved word? A special or already used word?

scripts/beck/st/error_injection.py
```diff
@@ -1,66 +1,78 @@
...
+logger.info('Induce Error in All Record = {!r}'.format(args.allErr))
+logger.info('Induce Error in GMD Record = {!r}'.format(args.allGMD))
+logger.info('Induce Error in DMD Record = {!r}'.format(args.allDMD))
+logger.info('Induce 512k errors         = {!r}'.format(args.err512k))
+logger.info('Induce huge corruption     = {!r}'.format(args.hugeCorruption))
+logger.info('Seed for random number     = {!r}'.format(args.seed))
+
+filename  = args.mfile
+recordType= args.Record_Type
 noOfCorr  = args.noOfErr
```
huanghua78, line 80: Please use the same name:
corruption?
Error?

scripts/beck/st/error_injection.py
```diff
@@ -1,66 +1,78 @@
...
 
+if args.seed != 0:
+	seed = args.seed
+	logger.info("Seed used: {}".format(seed))
+else:
+	seed = time.time()
+	logger.info("Seed used: {}".format(seed))
+
+random.seed(seed)
```
huanghua78, line 89: From here, I know the 'seed' is only used to initialize the "random" library: to initialize the random generation.

Please explain this in help.

scripts/beck/st/error_injection.py
```diff
@@ -147,14 +159,14 @@ def recordOffset(record, i, size):
 
 def ReadTypeSize(byte):					#Ex: 0001(ver) 0009(type) 00003dd8(size) 
 	ver=byte[:4]						#.ot_version = src->hd_bits >> 48,
```
huanghua78, line 179: s/byte[:4]/byte[0:4]/


No need to omit the starting index.

scripts/beck/st/README
gaurkrgaurav: As per Motr it is error injection in Metadata means corrupt some entries with over-writing some known pattern data on a particular location
As per S3 they call it a Hole. So kept both for better understanding 

Do you mean to insert a segment with all zeroes?
No, Not segment. Segment is different
"error injection"? What error is going to inject into the metadata file?
Write some unexpected data is Error in metadata
Just some array of random bytes with random size?
No, not that straight forward. Fix pattern "111122223334444" pattern
Or some array of all zeroes with random size?
At any offset in the metadata file?
Depend on what you are testing.
or it will overwrite some records with random bytes? At the record boundary?
Now it will overwrite some record depending on what functionality you are choosing at a particular offset of a record.

scripts/beck/st/error_injection.py
```diff
@@ -147,14 +159,14 @@ def recordOffset(record, i, size):
...
-	type=byte[6:8]						#.ot_type    = src->hd_bits >> 32 & 0x0000ffff,
+	rtype=byte[6:8]						#.ot_type    = src->hd_bits >> 32 & 0x0000ffff,
```
huanghua78, line 180: Should this be:
```
rtype=byte[4:8]
```
Why is it [6:8]?

scripts/beck/st/README
gaurkrgaurav: If someone wants to reproduce the scenario.

scripts/beck/st/README
gaurkrgaurav: GMD is the one that you can not recover and DMD records are recoverable.

scripts/beck/st/error_injection.py
huanghua78: I can not add comments to existing code. But I have comments too:

1. 
typeDict:
can we put one element on a separate line?
and other two dicts.

2.def recordOffset(record, i, size):
This function is ugly.

Can we do this:
```
def recordOffset(record, i, size):
        recordDict[record].append(i)
        sizeDict[record].append(size)
```

Maybe we need to assert the key exists in these two dicts.

scripts/beck/st/error_injection.py
```diff
@@ -147,14 +159,14 @@ def recordOffset(record, i, size):
...
+	#logger.info("Version {}, Type {}, Size {}".format(ver, rtype, size))
+	return rtype,size
 
 def EditMetadata(offset):
 	with open(filename, 'r+b') as wbfr:
-		print("** Corrupting 8byte of Metadata at offset {} with b'1111222244443333' **".format(hex(offset)))
+		logger.info("** Corrupting 8byte of Metadata at offset {} with b'1111222244443333' **".format(hex(offset)))
 		wbfr.seek(offset)
 		wbfr.write(b'\x33\x33\x44\x44\x22\x22\x11\x11')
 		wbfr.seek(offset)
```
huanghua78, line 190: "wbfr" is closed before return?

scripts/beck/st/error_injection.py
```diff
@@ -164,128 +176,156 @@ def ReadMetaData(offset):
 	with open(filename, "rb") as mdata:
 		mdata.seek(offset)
 		data=binascii.hexlify((mdata.read(8))[::-1])
-		print("** Metadata at offset {} is {}".format(hex(offset), data))
+		logger.info("** Metadata at offset {} is {}".format(hex(offset), data))
```
huanghua78, line 197: file mdata is not closed?

scripts/beck/st/error_injection.py
```diff
@@ -164,128 +176,156 @@ def ReadMetaData(offset):
...
 
 def induceCorruption(recordType, noOfCorr):
-	print(recordType)
-	print("No of Error want to induce: {}".format(noOfCorr))
+	count = 0
+	read_metadata_file()
+	logger.info(recordType)
+	logger.info("No of Error want to induce: {}".format(noOfCorr))
```
huanghua78, line 203: s/No of Error/Number of Error/

scripts/beck/st/error_injection.py
```diff
@@ -164,128 +176,156 @@ def ReadMetaData(offset):
...
 	lookupList = recordDict[recordType]
 	if (len(lookupList) & noOfCorr) == 0:
```
huanghua78, line 205: This is bit AND operation.
I don't understand this condition.

scripts/beck/st/error_injection.py
```diff
@@ -164,128 +176,156 @@ def ReadMetaData(offset):
...
-		print("Record List is empty. Please choose another Record")
+		logger.error("Record List is empty. Please choose another Record")
+		count = 0
+		return count
 	elif len(lookupList) < noOfCorr:
-		print(" Record List contains Less no of entry than Input. Please reduce the number of Error Injection")
+		logger.error(" Record List contains Less no of entry than Input. Please reduce the number of Error Injection")
```
huanghua78, line 210: Please replace all "no" with "number"

huanghua78: s/entry/entries/

huanghua78: s/Input/input/

scripts/beck/st/error_injection.py
```diff
@@ -164,128 +176,156 @@ def ReadMetaData(offset):
...
-		print(lookupList)
-		print("**** Inducing {} Error in Record: {} ****".format(noOfCorr,recordType))
+		logger.info(lookupList)
+		logger.info("**** Inducing {} Error in Record: {} ****".format(noOfCorr,recordType))
 		for i in range(noOfCorr):
-			offset = lookupList[i]
+			offset = lookupList[i]  ## Please add offset here for starting from middle of offset list
 			ReadMetaData(offset+8)
-			EditMetadata(offset+8)			
+			EditMetadata(offset+8)
```
huanghua78, line 219: I can not comment on existing code of "EditMetadata()".
But I prefer to remove the calling of "ReadMetaData" at the end of "EditMetadata()"
Otherwise this is very confusing.

Then we can do:
```
ReadMetaData(offset+8)   # Read Oringal
EditMetadata(offset+8)     # Modify
ReadMetaData(offset+8)   # Verify
```

scripts/beck/st/error_injection.py
```diff
@@ -164,128 +176,156 @@ def ReadMetaData(offset):
...
+		else:
+			lookupList = random.sample(lookupList,noOfCorr)
+			logger.info(lookupList)
+			for i in range(noOfCorr):
+				offset = lookupList[i]
+				logger.info("**** Inducing RANDOM Error in Record at offsets {}****".format(hex(offset+8)))
+				ReadMetaData(offset+8)
+				EditMetadata(offset+8)
+				ReadMetaData(offset+8)
+				count = count + 1
```
huanghua78, line 243: Can we call induceCorruption()?
Otherwise, code are duplicated.

scripts/beck/st/error_injection.py
```diff
@@ -164,128 +176,156 @@ def ReadMetaData(offset):
...
-			print(lookupList)
+			logger.info("Inducing {} Error at these offsets".format(len(lookupList)))
+			logger.info(lookupList)
 			for offset in lookupList:
-				print("**** Inducing Error in Record at offsets {}****".format(hex(offset+8)))
+				logger.info("**** Inducing Error in Record at offsets {}****".format(hex(offset+8)))
 				ReadMetaData(offset+8)
 				EditMetadata(offset+8)
 				ReadMetaData(offset+8)
+				count = count + 1
```
huanghua78, line 266: Can we call induceCorruption()?

scripts/beck/st/error_injection.py
```diff
@@ -164,128 +176,156 @@ def ReadMetaData(offset):
...
+	return count
 
 #Corrupt Metadata file from random location till end of metadata file
 def induceHugeError():
+	count = 0
 	with open(filename, 'r+b') as wbfr:
-		print("** Corrupting 8byte of Metadata at offset {} with b'1111222244443333' **".format(hex(offset)))
+		logger.info("** Corrupting 8byte of Metadata with b'1111222244443333' all place")
 		wbfr.seek(-1, os.SEEK_END)
```
huanghua78, line 318: This is the last but one byte in a file.
Is this right?

scripts/beck/st/error_injection.py
```diff
@@ -164,128 +176,156 @@ def ReadMetaData(offset):
...
 
 #Corrupt 512k Metadata in Metadata file from random location
 def induce512kbError():
+	count = 0
 	i = 0
 	j = 0
 	with open(filename, 'r+b') as wbfr:
-		#print("** Corrupting 8byte of Metadata at offset {} with b'1111222244443333' **".format(hex(offset)))
+		#logger.info("** Corrupting 8byte of Metadata at offset {} with b'1111222244443333' **".format(hex(offset)))
 		wbfr.seek(-524400, os.SEEK_END)
```
huanghua78, line 341: 512K == 512 << 10 == 524288

Please explain or add comment to the number "524400".

scripts/beck/st/error_injection.py
```diff
@@ -164,128 +176,156 @@ def ReadMetaData(offset):
...
+				#wbfr.seek(offset)
+				count = count + 1
+	return count
 
 #Corrupt 512k Metadata in Metadata file from random location
 def induce512kbError():
+	count = 0
 	i = 0
 	j = 0
 	with open(filename, 'r+b') as wbfr:
```
huanghua78, line 339: wbfr is closed?

scripts/beck/st/error_injection.py
```diff
@@ -295,62 +335,77 @@ def induce512kbError():
...
+					#wbfr.write(b'\x33\x33\x44\x44\x22\x22\x11\x11')
+					count = count + 1
+	return count
 
 		
 def verifyLengthOfRecord(recordDict):
-	print("***********Record list will be print here************")
+	count = 0
+	read_metadata_file()
+	logger.info("***********Record list will be logger.info here************")
```
huanghua78, line 366: logger.info is not a "verb"

scripts/beck/st/error_injection.py
```diff
@@ -295,62 +335,77 @@ def induce512kbError():
...
-
-	
-if args.err512k:
-	induce512kbError()
-	
-if args.hugeCorruption:
-	induceHugeError()
+		logger.info(" {} :  {}".format(record, len(items)))
+		count = count + 1
+	return count
```
huanghua78, line 370: Where is the "verify"?
It only output the records into logs.
It does not "verify".

scripts/beck/st/error_injection.py
```diff
@@ -295,62 +335,77 @@ def induce512kbError():
...
 	
-
-with open(filename, "rb") as metadata:
-	if os.path.getsize(filename):
+def read_metadata_file():
+	with open(filename, "rb") as metadata:
+		#if os.path.getsize(filename):
```
huanghua78, line 374: Why is it commented out?
If it is not used, please remove it.

scripts/beck/st/error_injection.py
```diff
@@ -295,62 +335,77 @@ def induce512kbError():
...
-			flag=0
 			byte=metadata.read(8)
 			i=i+8
 			if not byte:
 				break
 			byte=binascii.hexlify(byte[::-1])
 			if byte == header:
-				byte=binascii.hexlify((metadata.read(8))[::-1])					#Read the Type Size Version
+				byte=binascii.hexlify((metadata.read(8))[::-1])			#Read the Type Size Version
 				i=i+8
```
huanghua78, line 384: This and everywhere else:
i = i + 8
Please follow the coding style.
one space between variable, =, +

scripts/beck/st/README
huanghua78: I see.

scripts/beck/st/error_injection.py
gaurkrgaurav: Yes, that true. It will print out all record numbers in the log So user can verify those number with a dry run of beck tool.

scripts/beck/st/error_injection.py
gaurkrgaurav: its typical s/print/loggger.info/g mistake. Thanks I will correct this

scripts/beck/st/error_injection.py
gaurkrgaurav: Took a bigger number than 512k so a few records may left without corruption for smooth run.

scripts/beck/st/error_injection.py
gaurkrgaurav: Did not understand? No I am using wbfr in this function and It will close automatically once the function end

scripts/beck/st/error_injection.py
gaurkrgaurav: b'1111222244443333' I choose this pattern to overwrite on all existing metadata bytes

scripts/beck/st/error_injection.py
gaurkrgaurav: No. this is for taking a log like Read, Corrupt, Read at a particular offset and log it. So log will be user friendly for reading.

scripts/beck/st/error_injection.py
gaurkrgaurav: Induce Corruption functionality is different. Which is induce corruption at a particular record type. 
No. this is for taking a log like Read, Corrupt, Read at a particular offset and log it. So log will be user friendly for reading.

scripts/beck/st/error_injection.py
gaurkrgaurav: Thats correct 👍 I will add a comment in my script

scripts/beck/st/error_injection.py
gaurkrgaurav: Yes if the length is zero and the number of errors user input is zero then only fulfill this condition

scripts/beck/st/error_injection.py
gaurkrgaurav: If a file handler open using with then no need to close it. It will automatically be closed once its definition/function gets over.

scripts/beck/st/error_injection.py
gaurkrgaurav: It is particular for function only

scripts/beck/st/error_injection.py
gaurkrgaurav: No. Dropped a few bit for a specific reason. which no need to mention

scripts/beck/st/error_injection.py
gaurkrgaurav: Let me test that 👍

scripts/beck/st/error_injection.py
gaurkrgaurav: I don't know why is it so. This is not a keyword, reserved or special word.

scripts/beck/st/error_injection.py
gaurkrgaurav: Added again. Mistakenly copied local file

ravindrachoudhari (approved):

ketanarlulkar (approved):

scripts/beck/st/error_injection.py
huanghua78: "No need to mention"?

Please add comments here to mention this.
If the specific reason is not needed to explain, at least to mention:
```
A few bit is dropped for a specific reason.
```
Otherwise, reviewer and later code maintainer will not understand this.

huanghua78 (approved):

PR #61, opened 2020-09-29
EOS-10102 utils: script to delete dump and trace files

vishwas-bhat:
This helper bash script will check for existence of m0crate process and
will delete core dump files in /var/crash and trace files in the
specified ioservice directory in /var/motr every 1000 seconds.

Signed-off-by: Vishwas Bhat <vishwas.bhat@seagate.com>

### DIFF
--- a/utils/m0_clear_dump.sh
+++ b/utils/m0_clear_dump.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+count=1
+if [[ ! $# -eq 2 ]]
+then
+	echo "Invalid usage: m0_clear_dump -i <ioservice name>"
+	exit 2
+fi
+while getopts i: flag
+do
+	case "${flag}" in
+		i) ioservice=${OPTARG};;
+	esac
+done
+ios_dir="/var/motr/m0d-$ioservice"
+if [ ! -d "$ios_dir" ] 
+then
+	echo "Error: directory for $ioservice doest not exist" 
+	exit -20 # ENOTDIR
+fi
+m0crate_pid=($(pgrep m0crate))
+if [ -z "${m0crate_pid[0]}" ]
+then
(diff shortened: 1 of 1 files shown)


utils/m0_clear_dump.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
```
huanghua78, line 1: Personally, I think this can be a utility script to delete dump files/trace files/core files in a test. This is useful when doing a specific test case.

But to add it into the source code, I am not sure about the necessity.

utils/m0_clear_dump.sh
```diff
@@ -0,0 +1,38 @@
...
+
+while (( ${#m0crate_pid[@]} )); do
+
+	echo '-----------'
+	echo 'Count :' $count
+	echo '-----------'
+	((count++))
+	sleep 1000
+	rm -rf "$ios_dir"/m0trace.*
+	rm -rf /var/crash/core.0.*
```
huanghua78, line 36: Are the crash files moved to "/var/log/crash"?

vishwas-bhat: No they aren't - they're deleted, do you want me to delete the crash files there as well?

huanghua78: I see.

The core files are generated by m0crate.
They are not the same as "m0d" crash cores.

madhavemuri:
I think this change is not required and is specific to VM, as log rotate for m0traces and crashes is already available in production environments.

huanghua78:
@vishwas-bhat  Is this PR still needed?

vishwas-bhat:
> @vishwas-bhat Is this PR still needed?

@huanghua78 , not sure. I can delete it if no longer applicable.

madhavemuri:
This PR is not required.

PR #66, opened 2020-09-30
m0_filesystem_stats: use minion id instead of hostname in consul query

yeshpal-jain-cs:
hostname was used to get the current host endpoints from consul,
in the recent builds, consul sends minion_id instead of hostname
so server and client endpoints were not recieved.
Solution: use minion_id instead of hostname.

### DIFF
--- a/utils/spiel/m0_filesystem_stats
+++ b/utils/spiel/m0_filesystem_stats
@@ -150,12 +150,25 @@ if [ $rc -ne 0 ] ; then
 fi
 }
 
+get_hostname()
+{
+        host=" "
+        if [[ -f /etc/salt/minion_id ]] ; then
+            host=$(cat /etc/salt/minion_id)
+        fi
+
+        if [ -z $host ] ; then
+           host="localhost"
+        fi
+        echo $host
+}
+
 # Get endpoints of local node's processes that run service(s) of given type.
 get_endpoints_of() {
 	local svc_type=$1
 
 	$consul_query_cmd |
-        awk -F/ -v host=$(hostname --fqdn) -v svc_type=$svc_type '
+        awk -F/ -v host=$(get_hostname) -v svc_type=$svc_type '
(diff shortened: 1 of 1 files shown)


utils/spiel/m0_filesystem_stats
```diff
@@ -150,12 +150,25 @@ if [ $rc -ne 0 ] ; then
...
 
+get_hostname()
+{
+        host=" "
+        if [[ -f /etc/salt/minion_id ]] ; then
+            host=$(cat /etc/salt/minion_id)
+        fi
+
+        if [ -z $host ] ; then
+           host="localhost"
```
yatin-mahajan, line 161: [Suggestion]
Keep your exiting logic as it is [i.e. host=$(hostname --fqdn) ] in absent of minion_id file. 
localhost would work for singlenode only. 
In case of multinode and deployed with only (motr + hare) we need to use actual fqdn.

yatin-mahajan (approved):
Looks good to me. One minor suggestion.

utils/spiel/m0_filesystem_stats
```diff
@@ -150,12 +150,25 @@ if [ $rc -ne 0 ] ; then
...
 }
 
+get_hostname()
+{
+        host=" "
+        if [[ -f /etc/salt/minion_id ]] ; then
+            host=$(cat /etc/salt/minion_id)
+        fi
+
+        if [ -z $host ] ; then
```
huanghua78, line 160: [defect]. Please write as this:
```
if [ -z "$host" ] ; then
```

String must be quoted, otherwise this condition will always be true.

yeshpal-jain-cs: This should not affect for zero check, it is recommended to use "" for string comparision.
have did this changes.

huanghua78 (approved):

madhavemuri (approved):

PR #68, opened 2020-09-30
EOS-8613:Increase MOTR_M0D_BELOG_SIZE to 4GB.

tejalsheth:
Problem: Performance is degraded to more than 60% after failover
	 and never comes back.

Analysis:
	Observed log starvation when TX FOMs transit from open to grouping state
	When Tx FOMs transit from open to grouping in 	[be_engine_got_tx_open]
	Tx Engine tries log reservation for each IO FOMs that participate in grouping.
	If there is insufficient log space, Tx engine has to wait till availability of log space for grouping.
	If behavior persist, it will overall increase the	Tx completion time.
	Finally, it will reflect in completion of IO FOMs leading	to performance drop.
Solution:
	Increased Log size 128MB[default] to 4GB,to overcome the	log starvation.


nihar1313 (approved):

madhavemuri (approved):

just-now (approved):
I'm OK with this change.

PR #72, opened 2020-10-02
EOS-13707: not posting ast to cancel DIX_NEXT operation

jadhavik1:
m0_op_wait after m0_op_cancel() wait continiusly for state
M0_OS_STABLE or M0_OS_FAILED for DIX_NEXT index operation.
DIX_NEXT operation has launched but not completed/failed
it is ongoing, It is expected that posted ast to cancel
operation should cancel ongoing DIX_NEXT operation
as seen op->dg_completed_nr < rop->dg_cas_reqs_nr.

sm group oi->oi_sm_grp could not call
any posted ast callback, it observed sm group
waiting at m0_chan_wait for corresponding
loc_handler_thread.

This issue observed with st "26motr-user-kernel-tests"
if run in iterations and seen only for DIX_NEXT operation
so not posting ast to cancel DIX_NEXT operation.

Signed-off-by: Vishwas Bhat <vishwas.bhat@seagate.com>

### DIFF
--- a/motr/idx_dix.c
+++ b/motr/idx_dix.c
@@ -1074,7 +1074,8 @@ M0_INTERNAL int m0__idx_cancel(struct m0_op_idx *oi)
 	 * Do no post cancel ast when oi completion/fail
 	 * already called and for not handled types.
 	 */
-	if (!M0_IN(dreq->dr_type, (DIX_CREATE,
+	if (!M0_IN(dreq->dr_type, (DIX_NEXT,
+				   DIX_CREATE,
 				   DIX_DELETE,
 				   DIX_CCTGS_LOOKUP)) &&
 	    !oi->oi_in_completion)


jadhavik1:
Ignore , need better fix as DIX_NEXT cancellation required

jadhavik1:
you can review patch, 
Better solution can be added, tried multiple solution but did not work.
so request to consider this for GA as it is required.

huanghua78:
> you can review patch,
> Better solution can be added, tried multiple solution but did not work.
> so request to consider this for GA as it is required.

Do you want to add comments to say that this is a temporary workaround and fix to this problem is still needed?

jadhavik1:
Yes, this temporary workaround for GA and fix to this problem is still needed, overall cancel support is not stable as we not handle server side.

huanghua78 (approved):

huanghua78:
What is the next step?

jadhavik1:
as per @madhavemuri this issue won't come on cluster as seen part of kernel st, he suggested to test list operation from s3, this can be closed if not seen if listing indexes, need help of s3 test team here.

madhavemuri:
Let's wait for QA testing cycle to complete, if this issue is seen theh this PR can be landed.

jadhavik1:
This is not seen during QA cycle and as discussed with team this won't occur on cluster.

PR #85, opened 2020-10-05
EOS-13890 m0reportbug: Removed halon related things and fixed warn/error logs

yatin-mahajan:
* In last version update there was version mismatch, corrected that.

* Updated correct name of libgalois. libgalois  library name itself
  contains version name (libgalois-0.1.so) updated actual name in
  path_rpm() and path_src()

* Redirected error logs (m0tracedump, journalctl) into REPORT file so
  that it won't reflect on terminal but can be seen in report summary.

* Removed logs/binary collection related to halon since we are not using
  it, also bumped up version by 2 (since it was missed in last commit)

### DIFF
--- a/utils/m0reportbug
+++ b/utils/m0reportbug
@@ -26,7 +26,7 @@ export PS4='+ [${FUNCNAME[0]:+${FUNCNAME[0]}:}${LINENO}] '
 ### When modifying this script, please increment VERSION and add new
 ### entry to the 'History' section below.
 ### ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-VERSION=0.50
+VERSION=0.52
 
 OUT_DATA_DIR=m0reportbug-data
 OUT_CORES_DIR=m0reportbug-cores
@@ -179,7 +179,6 @@ probes() {
         'uptime'
         'm0version'
         'm0d -v'
-        'halond -v'
         'rpm -qi cortx-motr'
         'rpm -qi cortx-hare'
         'rpm -qR cortx-motr'
@@ -215,7 +214,6 @@ probes() {
         'ls /dev/disk/by-*'
         'dmesg'
         'systemctl status -l'
-        'systemctl status -l halond'
         'systemctl status -l motr-kernel'
         'lctl dk'
         'hctl status'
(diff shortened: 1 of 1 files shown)


huanghua78 (approved):

utils/m0reportbug
```diff
@@ -709,7 +681,7 @@ m0traces() {
         ## Kernel buffer.
         if [ -x "$m0trace" ] && lsmod | grep -q '^m0ctl\b'; then
             mkdir $outdir
-            $m0trace kernel-buf -Y | _xz >$outdir/kernel-buf.yaml.xz
+            $m0trace kernel-buf -Y 2>>$REPORT | _xz >$outdir/kernel-buf.yaml.xz
```
yeshpal-jain-cs, line 684: If by any chance m0trace fails because of some other issue, then we would not be able to understand why it is failing, 
Can we collect the error in another variable, output the to terminal if it is failed.

yatin-mahajan: I think we can remove this part now. 
I see this error every time now since we have removed debugfs for opensource. 
and there is no point to collects kernel-buf because we won't be getting hence onwards.

madhavemuri: mount | grep m0t1fs
Get this info only if above command show the m0fs mount.

yatin-mahajan: @madhavemuri  Updated the patch as you mention, but I am quite unsure, will we have kernel buffer in case of m0t1fs client. 

Because to test this I tried to collect kernel-buffer (m0trace kernel-buf -Y | less) while m0t1fs is mounted during m0t1fs_dgmode_io.sh test run, but still I observed similar error.
"m0tracedump: Failed to open input file '/sys/kernel/debug/motr/trace/buffer': No such file or director"

huanghua78: @yatin-mahajan 
Can you please ignore the following error in script:
```
"m0tracedump: Failed to open input file '/sys/kernel/debug/motr/trace/buffer': No such file or director"
```

The '/sys/kernel/debug/motr/trace/buffer' is not available at this moment, because we have lack of debugfs support due to NON-GPL kernel module license.

Please skip this '/sys/kernel/debug/motr/trace/buffer' related code or ignore the error in code.

Please refer to "https://jts.seagate.com/browse/EOS-10625"

yatin-mahajan: After discussion with Madhav, I have removed it completely.

madhavemuri (approved):

yeshpal-jain-cs (approved):

PR #107, opened 2020-10-09
EOS-14131: replace core to motr in setup.yaml

yeshpal-jain-cs:


scripts/install/opt/seagate/cortx/motr/conf/setup.yaml
```diff
@@ -17,7 +17,7 @@
 # please email opensource@seagate.com or cortx-questions@seagate.com.
 #
 
-core:
+motr:
```
yatin-mahajan, line 20: I believe this file is also used by CSM, confirm once does they also need to change anything. 
ignore if you have already checked.

yatin-mahajan (approved):
One minor comment/doubt.

huanghua78 (approved):

PR #111, opened 2020-10-12
EOS-13842: unique RPC id across cluster

AbhishekSahaSeagate:
Now RPC ids are generated by hashing the unique fid of reqh service
with the timestamp seed to make the RPC id unique across the cluster.

This change is required to mitigate the case when RPC ids collide
due to them generated at the same local time across different systems
with unsynchronized clocks.

Signed-off-by: Abhishek Saha <abhishek.saha@seagate.com>

### DIFF
--- a/rpc/session_utils.c
+++ b/rpc/session_utils.c
@@ -33,6 +33,7 @@
 #include "lib/bitstring.h"
 #include "lib/finject.h"       /* M0_FI_ENABLED */
 #include "lib/uuid.h"
+#include "lib/hash.h"          /* m0_hash() */
 #include "fop/fop.h"
 #include "reqh/reqh.h"
 #include "rpc/rpc_internal.h"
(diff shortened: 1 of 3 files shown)


huanghua78 (approved):
Looks fine.

rpc/session_utils.c
```diff
@@ -104,6 +105,8 @@ M0_INTERNAL uint64_t m0_rpc_id_generate(void)
 		id = (millisec << 10) | (m0_atomic64_get(&cnt) & 0x3FF);
 	} while (id == 0 || id == UINT64_MAX);
 
+	id = m0_hash(m0_fid_hash(uniq_fid) + id);
```
huanghua78, line 108: I think we need to move this into the above "do {} while()" loop.
In rare case, 'id' count be zero or UINT64_MAX, and we need to re-generate this id.

huanghua78 (changes requested):
Changes need.

rpc/session_utils.c
AbhishekSahaSeagate: Done

huanghua78 (approved):

PR #115, opened 2020-10-12
[EOS-14191] To reduce failover/failback time kill m0d processes

VinothVijayakumar-dev:
- added ExecStop=/bin/kill -9 {PID} in m0d@service

Signed-off-by: Vinoth <Vinoth.V@.seagate.com>

### DIFF
--- a/scripts/install/usr/lib/systemd/system/m0d@.service
+++ b/scripts/install/usr/lib/systemd/system/m0d@.service
@@ -32,5 +32,5 @@ Environment=SHELL=/bin/bash
 EnvironmentFile=/etc/sysconfig/m0d-%i
 Group=motr
 ExecStart=/usr/libexec/cortx-motr/motr-server m0d-%i
-ExecStop=/bin/bash -c "/usr/bin/echo 'killing PID : ${MAINPID}'"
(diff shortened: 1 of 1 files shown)


scripts/install/usr/lib/systemd/system/m0d@.service
```diff
@@ -32,3 +32,5 @@ Environment=SHELL=/bin/bash
 EnvironmentFile=/etc/sysconfig/m0d-%i
 Group=motr
 ExecStart=/usr/libexec/cortx-motr/motr-server m0d-%i
+ExecStop=/bin/bash -c "/usr/bin/echo 'killing PID : ${MAINPID}'"
```
huanghua78, line 35: Why don't directly exec echo:
```
ExecStop=/usr/bin/echo 'killing PID : ${MAINPID}'
```
?
What is the reason to run it in a bash shell?

madhavemuri:
I think this needs to be moved to cortx-ha build ees HA script, which makes it specific to LDR R1.

VinothVijayakumar-dev:
Closing this PR, as the change has been moved to  cortx-ha
https://github.com/Seagate/cortx-ha/pull/214

as this change is not required for JBOD.

PR #116, opened 2020-10-13
m0reportbug: Collects journal logs even though there is nothing in li…

yatin-mahajan:
…st boot

We used to collect journalctl logs for last N boots listed into
`journalctl --lit-boots` but on some node it has been observed that
list-boots doesn't list anything. Started collection of journal logs
with boot id `0` (which is current boot) in such cases.


yatin-mahajan:
looks like this is caused by bug in systemd, as I had seen this core of journal log there. 
after re-imaged it is listing boot id's properly. 
may be this patch is not needed, hence discarding and closing this PR.

PR #123, opened 2020-10-15
EOS-14161: m0_fop_put() was missing from some UTs

yatin-mahajan:
Problem:
    1) arg_fop was getting allocated on stack in remote_invocation() that
       way same fop was getting re-used on futher remote_invocation call,
       that was causing to stuck in m0_tlist_invarient().

    2) m0_fop_put() was missingfom in some of sub ut's of "isc-service-ut"
       which was casuing to panic in rpc_session_fini() since m0_fop_fini()
       is not called where deletion of xid happen from session->xid_list.
    ```
      #0 in raise () from /lib64/libc.so.6
      #1 in abort () from /lib64/libc.so.6
      #2 in m0_arch_panic () at lib/user_space/uassert.c:131
      #3 in m0_panic () at lib/assert.c:52
      #4 in m0_list_fini () at lib/list.c:38
      #5 in m0_tlist_fini () at lib/tlist.c:56
      #6 in xidl_tlist_fini () at rpc/item.c:85
      #7 in m0_rpc_item_xid_list_fini () at rpc/item.c:1682
      #8 in m0_rpc_session_fini_locked () at rpc/session.c:321
      #9 in m0_rpc_session_fini () at rpc/session.c:304
      #1 in m0_rpc_session_destroy ()at rpc/session.c:567
    ```

    Solution:
    1) Used heap allocation for fop_arg instead of stack allocation.
    2) Added m0_fop_put0_locked() at missing places.
    3) Fixing up "remote-comp-signature" ut reveiled other two ut failures
       due to similar reason, fixed those as well.

### DIFF
--- a/iscservice/ut/service_ut.c
+++ b/iscservice/ut/service_ut.c
@@ -593,6 +593,58 @@ static uint32_t remote_invocation_async(int exp_rc, uint32_t buf_type,
 	return FPP_INVALID;
 }
 
+static void req_fop_prepare2(struct m0_fop *req_fop,
+			     uint32_t       buf_type,
+			     struct m0_fid *fid,
+			     uint32_t       f_type)
+{
+	struct m0_fop_isc     *fop_isc;
+	struct m0_buf          data = M0_BUF_INIT0;
+	struct m0_rpc_machine *rmach;
+	char                  *in_buf;
+	uint32_t               size;
+	int                    rc;
+
+	fop_isc = m0_fop_data(req_fop);
+	size = 0;
+	rmach = &isc_ut_cctx.rcx_rpc_machine;
+	fop_isc->fi_comp_id = *fid;
+	m0_rpc_at_init(&fop_isc->fi_args);
+	switch (f_type) {
+	case FT_NEITHER_IO:
+	case FT_NO_INPUT:
+		break;
+	case FT_NO_OUTPUT:
+		in_buf = m0_strdup(fixed_str);
+		M0_UT_ASSERT(in_buf != NULL);
+		m0_buf_init(&data, (void *)in_buf, strlen(fixed_str));
+		break;
+	case FT_BOTH_IO:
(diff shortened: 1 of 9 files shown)


iscservice/ut/service_ut.c
```diff
@@ -563,7 +569,11 @@ static uint32_t remote_invocation_async(int exp_rc, uint32_t buf_type,
 		arg_fop[0] = m0_alloc(sizeof arg_fop[0][0]);
 		M0_UT_ASSERT(arg_fop[0] != NULL);
 		m0_fop_init(arg_fop[0], &m0_fop_isc_fopt,
-			    &remote_call_info.riv_req, m0_fop_release);
+			    NULL, m0_fop_release);
+		rc = m0_fop_data_alloc(*arg_fop);
```
huanghua78, line 573: I think we can use `arg_fop[0]` here, instead of `*arg_fop`,
because it is used elsewhere.

iscservice/ut/service_ut.c
```diff
@@ -885,7 +900,8 @@ static void test_remote_waiting(void)
 
 	/* Wait for reply fop. */
 	m0_chan_wait(&waiter);
-	phase = remote_invocation_async(0, BT_INLINE, phase, &arg_fop);
+	phase = remote_invocation_async(&fid, 0, BT_INLINE, phase, &arg_fop);
+	m0_fop_put0_lock(arg_fop);
```
huanghua78, line 904: Is this already called inside the remote_invocation_async()?

yatin-mahajan: m0_fop_put0_lock() is there inside remote_invocation_async(), 
but it does not call m0_fop_fini() in it, because ref->ref_cnt is not 0 at that point.
Hence called again.

iscservice/ut/service_ut.c
yatin-mahajan: Addressed in new PR raised against main branch.

PR #128, opened 2020-10-16
Uploading PDF files and RST Files conversion

VenkyOS:
This PR is raised to upload the PDF documents into GitHub. There are also few RST files that were converted from Sharepoint.

Regards,
Venky


huanghua78:
Some of the pdf files, have special letters in their name: this include SPACE, '(', ')', and others.
Please use "under score" to connect the words in names.

huanghua78:
Those newly uploaded pdf files and rst files haven't been added to "doc/motr-design-doc-list.rst", have they?

There are two new columns still empty.

VenkyOS:
> Those newly uploaded pdf files and rst files haven't been added to "doc/motr-design-doc-list.rst", have they?
> 
> There are two new columns still empty.

Yes. I want to upload the files first. Then, add links and raise a  separate PR for that.

VenkyOS:
> Some of the pdf files, have special letters in their name: this include SPACE, '(', ')', and others.
> Please use "under score" to connect the words in names.

I have updated that. Can you please look into it now?

huanghua78 (approved):

doc/HLD_Resource_Management_Interface.rst
```diff
@@ -0,0 +1,251 @@
+==================================
+HLD Resource Management Interface
+==================================
+
+This document presents a high level design (HLD) of scalable resource management interfaces for Mero. The main purposes of this document are: (i) to be inspected by M0 architects and peer designers to ascertain that high level design is aligned with M0 architecture and other designs, and contains no defects, (ii) to be a source of material for Active Reviews of Intermediate Design (ARID) and detailed level design (DLD) of the same component, (iii) to serve as a design reference document.
```
madhavemuri, line 5: s/Mero/motr

VenkyOS: Incorporated.

andriytk:
Some of the pdf files are really very outdated. For example, `doc/PDF/Motr_ Architecture_Documentation.pdf` file is kind of a guideline for Colibri architecture documentation. It even uses old C2 prefix, not mentioning that all the links point to old Xyratex google-docs there of which no one is working, of course.

Another example is `doc/PDF/Motr_FAQ.pdf` which mentions db5 as a backend KV-store, which is not used in Motr for many years already (instead, we have BE).

And these are only two documents I opened. I'm sure all of them need to be revised. Otherwise, we risk only to confuse the community instead of helping with understanding of Motr architecture.

cc @johnbent , @nikitadanilov

huanghua78:
> Some of the pdf files are really very outdated. For example, `doc/PDF/Motr_ Architecture_Documentation.pdf` file is kind of a guideline for Colibri architecture documentation. It even uses old C2 prefix, not mentioning that all the links in it point to old Xyratex google-docs (of which no one is working, of course).
> 
> Another example is `doc/PDF/Motr_FAQ.pdf` which mentions db5 as a backend KV-store, which is not used in Motr for many years already (instead, we have BE).
> 
> And these are only two documents I opened. I'm sure all of them need to be revised. Otherwise, we risk only to confuse the community instead of helping with understanding the Motr architecture.
> 
> cc @johnbent , @nikitadanilov

Hello @andriytk , you are right. These documents are directly generated from the existing and old design documents.
As discussed, it's better to revise these documents according to our current architecture and implementation.
But this will be a future task.

The google links are just for reference, because they were in the code. We have some of them migrated to Sharepoints.
Some of them are still not found in sharepoints. We will delete the google links and sharepoints links when all documents have been converted to rst format.

andriytk:
@huanghua78 thanks for the update.

PR #102, opened 2020-10-08
EOS-13886 rpc/item: adjust session xid if remote item is not going to come

max-seagate:

PR #130, opened 2020-10-17
EOS-14165 del rpc_item from session->xidl if fini session before fop

huanghua78:
In Spiel, the fop and rpc session are initialized and finalized separately.
Sometime, the rpc session is finalized before fop.
So, we need to delete rpc_item from session->xidl if fini session before fop in that case.


huanghua78:
Please note, only the last commit "https://github.com/Seagate/cortx-motr/pull/130/commits/a7b7743ab276e04c5fdfe82805a0e5d4592bda3a" is for this ticket.
Other commits are from #102.

So, when you do review for this PR, please focus on the last commit: a7b7743

max-seagate (changes requested):
> So, we need to delete rpc_item from session->xidl if fini session before fop in that case.

The situation when xidl is not empty when rpc session is finalised is a bug, and the patch silences it. It has to be fixed, not silenced.

RPC item lifetime should be less than RPC item's session lifetime. Having something in xidl during session indicates that some rpc items from the session are not finalised.

huanghua78:
> > So, we need to delete rpc_item from session->xidl if fini session before fop in that case.
> 
> The situation when xidl is not empty when rpc session is finalised is a bug, and the patch silences it. It has to be fixed, not silenced.
> 
> RPC item lifetime should be less than RPC item's session lifetime. Having something in xidl during session indicates that some rpc items from the session are not finalised.


Then, spiel code needs to be re-organized to init rpc session before (creating fops & sending fops & destroy fops) and finalize rpc session after that.

huanghua78:
More work is needed.

PR #139, opened 2020-10-20
EOS-14345: Use setup.yaml to update BE LOG size in LDR R1

imvenkip:
Problem: We need to update BE log size during update cluster in
maintenance mode if MOTR_M0D_BELOG_SIZE change in /etc/sysconfig/motr

Solution: "/opt/seagate/cortx/motr/conf/setup.yaml" will trigger during
update cluster in maintenance mode. Added "be_log_resize" in setup.yaml
to trigger "m0provision be_log_resize". It will execute be_resize.sh to
modify BE log size if any change in /etc/sysconfig/motr.

Signed-off--by: Venkateswarlu Payidimarry <venkateswarlu.payidimarry@seagate.com>

### DIFF
--- a/scripts/install/opt/seagate/cortx/motr/config/be_resize.sh
+++ b/scripts/install/opt/seagate/cortx/motr/config/be_resize.sh
@@ -1,0 +1,194 @@
+#!/usr/bin/env bash
+# set -x
+
+current_time=$(date +"%Y_%m_%d__%H_%M_%S")
+LOG_DIR="/var/log/seagate/motr"
+mkdir -p $LOG_DIR
+LOG_FILE="$LOG_DIR/be_log_resize_$current_time.log"
+touch $LOG_FILE
+MOTR_CONF_FILE="/etc/sysconfig/motr"
+M0BETOOL="m0betool"
+MD_DIR="/var/motr"
+BE_REPLY="be_recovery_run"
+BE_RESIZE="be_log_resize"
+
+local_md_volume=""
+remote_md_volume=""
+new_be_log_size=""
+current_be_log_size=""
+local_node=""
+remote_node=""
+loca_ios_fid=""
+remote_ios_fid=""
+
+be_resize_log() {
+    echo "[`date`]: $@" 2>&1 | tee -a ${LOG_FILE}
+}
+
+be_resize_soft_exit() { be_resize_log "$@" ; exit 0; }
+
+be_resize_hard_exit() { be_resize_log "$@" ; exit 1; }
+
+# Verifying cluster status
+check_cluster_status() {
+	if  [[ $((hctl status) 2>&1 | grep "Cluster is not running") ]]; then
+		be_resize_log "cluster is not running"
+	else
+		be_resize_hard_exit "cluster should be stop for BE log resize"
+	fi
+}
+
+# Extract current BE Log file size
+get_be_log_size() {
+	new_be_log_size=`grep -w MOTR_M0D_BELOG_SIZE $MOTR_CONF_FILE | grep -v '#' | cut -d"=" -f2`
+	if [[ -z $new_be_log_size ]]; then
+		be_resize_soft_exit "BELOG Size is not configured"
+	fi
+	be_resize_log "Given BE log size: $new_be_log_size"
+}
+
+# The function will execute command on the local node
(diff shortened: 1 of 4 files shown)


scripts/install/opt/seagate/cortx/motr/config/be_resize.sh
```diff
@@ -0,0 +1,194 @@
...
+BE_REPLY="be_recovery_run"
+BE_RESIZE="be_log_resize"
+
+local_md_volume=""
+remote_md_volume=""
+new_be_log_size=""
+current_be_log_size=""
+local_node=""
+remote_node=""
+loca_ios_fid=""
```
imvenkip, line 21: s/loca_ios_fid/local_ios_fid

scripts/install/opt/seagate/cortx/motr/config/be_resize.sh
```diff
@@ -0,0 +1,194 @@
...
+	fi
+	be_resize_log "Completed unmount_md_device"
+}
+
+check_cluster_status
+get_be_log_size
+mount_md_device
+get_ios_fid
+verify_be_log_size
+be_log_reply
```
madhavemuri, line 191: be_log_replay would be the right name for it.

scripts/install/opt/seagate/cortx/motr/config/be_resize.sh
```diff
@@ -0,0 +1,194 @@
...
+
+	elif [[ $(cat /var/lib/hare/node-name) == "srvnode-2" ]]; then
+		local_node="srvnode-2"
+		remote_node="srvnode-1"
+		local_md_volume=$(grep right-volume /opt/seagate/cortx/ha/conf/build-ees-ha-args.yaml |\
+				    awk '{ print $2 }')
+		remote_md_volume=$(grep left-volume /opt/seagate/cortx/ha/conf/build-ees-ha-args.yaml |\
+				    awk '{ print $2 }')
+	else
+		be_resize_hard_exit "ERROR: node-name is not avaialble"
```
madhavemuri, line 85: This is a valid case for LDR R2, needs to be handled, add a todo here.

scripts/install/opt/seagate/cortx/motr/config/be_resize.sh
```diff
@@ -0,0 +1,194 @@
...
+	fi
+
+	if ! run_cmd_on_local_node "mount $local_md_volume $MD_DIR"; then
```
madhavemuri, line 88: There are two cases here, if log resize is run through HA module, need to run on
both the nodes and when provisioner invokes update to will run on each node independently. So need to differentiate these two cases.

scripts/install/opt/seagate/cortx/motr/config/be_resize.sh
```diff
@@ -0,0 +1,194 @@
...
+    echo "[`date`]: $@" 2>&1 | tee -a ${LOG_FILE}
+}
+
+be_resize_soft_exit() { be_resize_log "$@" ; exit 0; }
+
+be_resize_hard_exit() { be_resize_log "$@" ; exit 1; }
+
+# Verifying cluster status
+check_cluster_status() {
+	if  [[ $((hctl status) 2>&1 | grep "Cluster is not running") ]]; then
```
madhavemuri, line 34: Check for m0d's too, which should not be running.

madhavemuri: Also for LDR R1, if mount point exists, return from here.

imvenkip: Handled two conditions.

scripts/install/opt/seagate/cortx/motr/config/be_resize.sh
imvenkip: Done.

scripts/install/opt/seagate/cortx/motr/config/be_resize.sh
imvenkip: be_resize.sh will execute in local node.  Modified code to support VM, LDR1 and LDR2 .

imvenkip:
BE log resize design has changed.  Created new ticket EOS-14600.

PR #142, opened 2020-10-21
EOS-14338 scripts: Changes to gen-src-skel to modify licence header

ydb242:
The file scripts/gen-src-skel was using license from top cortx-motr directory to generate header.
This needed to be changed as the licence in header was modifed for newer files.

Added a new file scripts/licence_header which has the required header text.
scripts/gen-src-skel now uses this file to add header text.


huanghua78 (approved):

rajkumarpatel2602 (approved):
looks good!

madhavemuri (approved):

PR #147, opened 2020-10-22
EOS-14408: Combine HW and VM motr master data recovery script

kanchan-chaudhari:
Problem:
Current motr master data recovery script doesn't support recovery on VM.

Solution:
Added support for hctl commands which will be being used to do cluster
operations on the VM.
Added code to save the m0trace files to /var/motr/datarecovery location
which will be generated while getting generation id of seg0/seg1 using m0beck.
Added code to get the local and remote motr device path from the `mount`
if script is executed on the VM.
To get the ioservice fids /etc/motr/confd.xc is used.

Signed-off-by: Kanchan Chaudhari <kanchan.chaudhari@seagate.com>


huanghua78:
Hello @kanchan-chaudhari ,
Please check if this PR is still valid and needed.
If not, please close it.
If yes, please update the status.

kanchan-chaudhari:
Closing this PR as this task is not current priority, we can reopen this PR whenever needed.

PR #150, opened 2020-10-23
scripts: support short client id in timelines scripts

sergey-shilov:
Problem:
As a result of the renaming we have got an inconsistency
between scripts that builds database and timelines scripts
for ioo and dix operations: prior ones use short 'id' as a
client id while latter ones use long 'client_id'.

Solution:
Proper solution is to use long client id, but currently for
several reasons we need to support short id in timelines
scripts.

THIS IS A TEMPORARY HACK NOT FOR LANDING!!!


madhavemuri:
Not needed

PR #158, opened 2020-10-26
EOS-14450: Added changes to resize addb stobs without mkfs

tejalsheth:
1. Generate new addb-stobs directory during every m0d restart instead of during mkfs only.
2. The newly generated addb-stob directory is appended with the server pid. i.e addb-stobs-$PID
3. Limit number of addb-stob directories in m0d directory to 2.

### DIFF
--- a/scripts/install/opt/seagate/cortx/motr/libexec/m0addb_logrotate.sh
+++ b/scripts/install/opt/seagate/cortx/motr/libexec/m0addb_logrotate.sh
@@ -1,0 +1,150 @@
+#!/bin/sh
+#
+# Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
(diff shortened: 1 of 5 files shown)


scripts/install/opt/seagate/cortx/motr/libexec/m0addb_logrotate.sh
```diff
@@ -0,0 +1,150 @@
...
+# argument1: <number of latest addb stob dir to retain>
+# Default number of latest addb stob dir is 2
+# ./m0addb_logrotate.sh 2
+
+usage() { 
+        echo "Usage: bash $(basename $0) [--help|-h]
+                     [-n LogDirCount]
+Retain recent modified directories of given count and remove older motr addb stob directories.
+where:
+-n            number of latest addb stob directories to retain
```
madhavemuri, line 31: Let's keep two addb directories for both VM and HW, no need to check for get_platform logic.

tejalsheth: Done

scripts/install/opt/seagate/cortx/motr/libexec/m0addb_logrotate.sh
```diff
@@ -0,0 +1,124 @@
...
+# Default number of latest addb stob dir is 2
+# ./m0addb_logrotate.sh 2
+
+usage() { 
+        echo "Usage: bash $(basename $0) [--help|-h]
+                     [-n LogDirCount]
+Retain recent modified directories of given count and remove older motr addb stob directories.
+where:
+-n            number of latest addb stob directories to retain 
+              Physical : Default count of addb stob directories are 2+2 first directories
```
madhavemuri, line 32: Keep a single comment for physical and virtual setups with 2 as an option.

tejalsheth: Done.

PR #160, opened 2020-10-27
EOS-14234 beck: Add scanner thread and partition support

NikhilBirgade:
1) Divide scanner into two parts. The first part will find the bnodes and
put these bnodes into a queue. The second part will process the bnodes 
in a thread.
2) Add partitions in scanner. There is one partition for COB, CTG and AD domain
each. 
3) Use IONBF as option in setvbuf for performance improvement.

### DIFF
--- a/be/tool/beck.c
+++ b/be/tool/beck.c
@@ -73,10 +73,29 @@
 M0_TL_DESCR_DECLARE(ad_domains, M0_EXTERN);
 M0_TL_DECLARE(ad_domains, M0_EXTERN, struct ad_domain_map);
 
-struct queue;
+struct queue {
+	pthread_mutex_t q_lock;
+	pthread_cond_t  q_cond;
+	struct action  *q_head;
+	struct action  *q_tail;
+	uint64_t        q_nr;
+	uint64_t        q_max;
+};
 
 struct scanner {
 	FILE		    *s_file;
+	/**
+	 * It holds the BE segment offset of the bnode which is then used by
+	 * scanner thread to read bnode.
+	 */
+	off_t		     s_start_off;
+	/**
+	 * Queue to store bnode offset in the BE segment.
+	 */
+	struct queue	     s_bnode_q;
+	/** Scanner thread which processes the bnodes. */
+	struct m0_thread     s_thread;
+	struct m0_mutex      s_lock;
 	off_t		     s_off;
 	off_t		     s_pos;
 	bool		     s_byte;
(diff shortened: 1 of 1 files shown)


be/tool/beck.c
```diff
@@ -400,14 +416,15 @@ static struct btype bt[] = {
 
 enum {
 	MAX_GEN    	 =     256,
+	MAX_SCAN_QUEUED	 = 1000000,
```
RakeshVaghasiya, line 419: Can we make it 10000000? This way we can ensure scanner can not be throttled by builder.
As per current speed of builder can process max 10000 actions and in node average action should be >= 128. builder will took 35.5 hours to complete this much node.While 1000000 can be processed by builder in 3.55 hours

be/tool/beck.c
```diff
@@ -73,10 +73,29 @@
 M0_TL_DESCR_DECLARE(ad_domains, M0_EXTERN);
 M0_TL_DECLARE(ad_domains, M0_EXTERN, struct ad_domain_map);
 
-struct queue;
+struct queue {
+	pthread_mutex_t q_lock;
```
huanghua78, line 77: Why don't we use m0_mutex?

be/tool/beck.c
```diff
@@ -73,10 +73,29 @@
...
+	pthread_cond_t  q_cond;
```
huanghua78, line 78: Why don't we use m0_cond?

be/tool/beck.c
```diff
@@ -162,6 +183,11 @@ struct action {
 	struct action           *a_prev;
```
huanghua78, line 183: Why don't we use m0_list_head and m0_list_link?

be/tool/beck.c
shashank-parulekar: @huanghua78 : I think you are right we should have done these and other changes you mentioned below and we will do these maybe as a part of beck enhancement since we will need to test these a lot. Since this code needs to land today for the code-freeze we will file a new Jira for these changes in the meanwhile you can approve these changes for now.

shashank-parulekar (approved):

huanghua78 (approved):

RakeshVaghasiya (approved):

madhavemuri:
@NikhilBirgade : Can you resolve the conflicts ?

NikhilBirgade:
> @NikhilBirgade : Can you resolve the conflicts ?

Done.

PR #173, opened 2020-10-30
 EOS-14234 beck: integrate tx bulk framework with builder

RakeshVaghasiya:
Following points are addressed
 1.  Integrate tx-bulk framework with beck
 2.  Fix race condition on s_chunk between deref() and get()
 3.  Tune be-engine parameters to get optimized performance.
 4.  Added  init and fini for mutex.


be/tool/beck.c
```diff
@@ -1554,16 +1624,37 @@ static void builder_fini(struct builder *b)
...
 	m0_reqh_fini(&b->b_reqh);
 	m0_free(b->b_backend.but_stob_domain_location);
 
-	printf("builder: actions: %9"PRId64" txs: %9"PRId64"\n",
-	       b->b_act, b->b_tx);
+	printf("builder: actions: %9"PRId64"\n", b->b_act);
+}
+
+static void  be_cfg_default_init(struct m0_be_domain_cfg  *dom_cfg,
+				 struct m0_be_tx_bulk_cfg *tb_cfg)
```
shashank-parulekar, line 1638: Mention some comment indicating these values provided maximum performance after experimentation. Or else Code reader will not know why these values were used instead of the ones which are filled by default for be_cfg.

mukundkanekar:
This PR is not against main branch, is that intended ?

be/tool/beck.c
```diff
@@ -282,7 +290,8 @@ static int  builder_init   (struct builder *b);
 static void builder_fini   (struct builder *b);
 static void ad_dom_fini    (struct builder *b);
 static void builder_thread (struct builder *b);
-static void builder_process(struct builder *b);
+static void  be_cfg_default_init(struct m0_be_domain_cfg  *dom_cfg,
```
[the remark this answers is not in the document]
NikhilBirgade: Done.

be/tool/beck.c
NikhilBirgade: Done.

NikhilBirgade:
> This PR is not against main branch, is that intended ?

The changes had dependency on tx bulk changes. The changes were created on top of `tx-bulk-integrate-be-queue` branch. To show only the diff of this PR, remote branch is selected as `tx-bulk-integrate-be-queue`. Once `tx-bulk-integrate-be-queue` is merged , destination will automatically get changed to `main`

huanghua78:
Which is the exact change commit in this PR?

RakeshVaghasiya:
> Which is the exact change commit in this PR?

Main change  commit :- 12f331d67d74442db553f8aa49b8ed5d3b3fa3f1 (EOS-14234 beck: integrate tx bulk framework with builder )
Addressed review commit  on top of main change : 989c12beae99afbfd3636432c81f572a857aa7ef (EOS-14234: address review comments)

be/tool/beck.c
```diff
@@ -1836,35 +1927,30 @@ static struct cache_slot *ctg_getslot_insertcred(struct ctg_action *ca,
...
-		ca->cta_cid.ci_fid = ca->cta_fid;
-		m0_fid_tchange(&ca->cta_cid.ci_fid, 'T');
-		ca->cta_cid.ci_layout.dl_type = DIX_LTYPE_DESCR;
-		m0_dix_ldesc_init(&ca->cta_cid.ci_layout.u.dl_desc,
-				  &(struct m0_ext) { .e_start = 0,
-				  .e_end = IMASK_INF },
-				  1, HASH_FNC_CITY, &b->b_pver_fid);
-		m0_ctg_ctidx_insert_credits(&ca->cta_cid, accum);
-	}
+	slot = cache_insert(&b->b_cache, cas_ctg_fid);
```
max-seagate: [* defect *] slot is being reinitialised before it's used. What is the use case?

RakeshVaghasiya: Already fixed, Please review with all commits

be/tool/beck.c
```diff
@@ -1218,6 +1247,7 @@ static int emap_prep(struct action *act, struct m0_be_tx_credit *credit)
 		return M0_RC(-EINVAL);
 	}
```
huanghua78, line 1243: Please assert:
0 <= id < AO_NR - AO_EMAP_FIRST;

be/tool/beck.c
```diff
@@ -459,6 +464,15 @@ static bool  dry_run = false;
 static bool  disable_directio = false;
 static bool  signaled = false;
 
+static struct m0_be_tx_bulk_cfg default_tb_cfg = (struct m0_be_tx_bulk_cfg){
+		.tbc_q_cfg = {
+			.bqc_q_size_max       = 1000,
+			.bqc_producers_nr_max = 1,
+		},
+			.tbc_workers_nr       = 0x40,
```
huanghua78, line 472: why are these constants used?
Do we have some explanation why they are?

be/tool/beck.c
```diff
@@ -459,6 +464,15 @@ static bool  dry_run = false;
 static bool  disable_directio = false;
 static bool  signaled = false;
 
+static struct m0_be_tx_bulk_cfg default_tb_cfg = (struct m0_be_tx_bulk_cfg){
```
huanghua78, line 467: I can not add comment for LINE 454, 455 in original code.

Can we move "struct scanner s;" and "struct builder b;" into main()?

If possible, don't define them as global variables. We can pass them from function to functions.  Using global variables is not a good idea.

be/tool/beck.c
```diff
@@ -1256,15 +1281,20 @@ static void emap_act(struct action *act, struct m0_be_tx *tx)
...
 			bo_reserve_extent(adom->sad_ballroom,
 					  tx, &ext,
 					  M0_BALLOC_NORMAL_ZONE);
 		if (rc != 0) {
+			m0_mutex_unlock(&b.b_emaplock[id]);
 			M0_LOG(M0_ERROR, "Failed to reserve extent rc=%d", rc);
 			return;
 		}
 
+		b.b_data += ((ext.e_end - ext.e_start) << adom->sad_babshift)
```
huanghua78, line 1295: For example, here, you are accessing and modifying member field of a global variable.
This is a multi-threaded program. Yes, this is under the "&b.b_emaplock[id])".
But this is still confusing.


And also when accessing the "&b.b_emaplock[id]".

be/tool/beck.c
```diff
@@ -252,7 +252,7 @@ struct builder {
 	const char                *b_be_config_file; /** BE configuration */
 
 	uint64_t                   b_act;
-	uint64_t                   b_tx;
+	uint64_t                   b_data;
```
huanghua78, line 255: Please add comments for these field members.

From the code, my understanding is that, this field is only used for debug log messages.

huanghua78 (approved):

be/tool/beck.c
RakeshVaghasiya: This is just for displaying data throughput.
Its okay if this value is approximate.

be/tool/beck.c
RakeshVaghasiya: No we can't.
Scanner and builder are very large in size so we can stack overflow issue. 
i.e. scanner structure  s_chunk and and in builder b_cache.

NikhilBirgade (approved):

be/tool/beck.c
huanghua78: I see. But we still can use malloc() to allocate them from heap.

Anyway, this is just a recommendation.

be/tool/beck.c
RakeshVaghasiya: Added explanation

be/tool/beck.c
RakeshVaghasiya: done

be/tool/beck.c
RakeshVaghasiya: Noted, Will take care in future

nikitadanilov (approved):

PR #181, opened 2020-11-04
EOS-14915: change m0trace log location to /var/log/seagate/motr

yeshpal-jain-cs:
- This is done as a precautionary measure, as trace buffers data is seen
  on meta-data segment.


huanghua78 (approved):

huanghua78:
This PR can be closed? 
Since #206 is created.

PR #191, opened 2020-11-09
motr dep: libedit and libedit-devel added in build dependency

bansisoni:
Signed-off-by: Bansidhar Soni <bansidhar.soni@seagate.com>


nikitadanilov (changes requested):
Please explain why is this needed and how it worked without it before.

Pull requests without description are not very useful.

bansisoni:
> Please explain why is this needed and how it worked without it before.
> 
> Pull requests without description are not very useful.

It is for m0hsm tool. But I can see this changes in https://github.com/Seagate/cortx-motr/pull/92/ pull request. So, closing this.

PR #202, opened 2020-11-11
EOS-14813: discard multiple log record sequentially in single ast

tejalsheth:
Fix for EOS-14813 and EOS-9668 
Problem:
As per the current implementation, the discarded records are received via the AST-mechanism
in backwards order via the m0_be_group_format_discard() calls. If the lg_discarded position is moved by the records sizes , it is possible that lgr_last_discarded at m0_be_log_record_io_prepare() picks up some interim lg_discarded value between
the 1st and the last AST-calls of m0_be_group_format_discard() and write it down into the log. If it is the last log record just before the crash - the wrong discarded position will be taken during the recovery on m0d process restart resulting in a data corruption.
Solution:
Post a single ast after fdatasync() (if there is no ast running already) and discard each log record sequentially in a loop in that ast.
This will avoid the case where lgr_last_discarded can get updated with the interim lg_discarded value due to multiple ASTs being posted.

### DIFF
--- a/be/log_discard.c
+++ b/be/log_discard.c
@@ -69,7 +69,6 @@ struct m0_be_log_discard_item {
 	 * To run ldsc_discard callback outside the m0_be_log_discard lock.
 	 * XXX Temporary solution.
 	 */
-	struct m0_sm_ast                ldi_discard_ast;
 };
 
 
@@ -122,7 +121,6 @@ M0_INTERNAL int m0_be_log_discard_init(struct m0_be_log_discard     *ld,
 	ld->lds_need_sync = false;
 	ld->lds_sync_in_progress = false;
 	ld->lds_sync_deadline = M0_TIME_NEVER;
-	ld->lds_discard_left = 0;
 	ld->lds_discard_waiting = false;
 	m0_be_op_init(&ld->lds_sync_op);
 	m0_be_op_callback_set(&ld->lds_sync_op, &be_log_discard_sync_done_cb,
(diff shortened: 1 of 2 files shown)


be/log_discard.c
```diff
@@ -69,7 +69,6 @@ struct m0_be_log_discard_item {
 	 * To run ldsc_discard callback outside the m0_be_log_discard lock.
```
just-now, line 69: Remove this comment

be/log_discard.c
```diff
@@ -273,32 +248,69 @@ static void be_log_discard_item_trydiscard(struct m0_be_log_discard      *ld,
...
-	} m0_tl_endfor;
+	be_log_discard_unlock(ld);
+
+	do {
+               be_log_discard_lock(ld);
+               sync_item = ld->lds_sync_item;
+               ldi = ld_start_tlist_pop(&ld->lds_start_q);
+               be_log_discard_unlock(ld);
+
+               M0_ASSERT(ldi != NULL); // it's not an endless loop
```
just-now, line 270: comment has to be cleaner, or just remove it. also, avoid C++ style.

be/log_discard.h
```diff
@@ -115,8 +115,14 @@ struct m0_be_log_discard {
 	 * XXX Temporary solution.
 	 */
 	struct m0_semaphore                  lds_discard_wait_sem;
-	int                                  lds_discard_left;
 	bool                                 lds_discard_waiting;
+
+
+	/**
+	 * Add single ast to discard multiple ldi
```
just-now, line 122: Please write a better comment, try to reuse previous:

`To run ldsc_discard callback outside the m0_be_log_discard lock.`

just-now (commented):
looks ok.

just-now:
rebranded as pull/209

PR #203, opened 2020-11-11
Adding function shipping document

johnbent:
It's out of date and needs to be replaced with a newer one (preferrably in RST or MD and not PDF) but at least it is something useful for community to see.


huanghua78 (approved):

doc/reading-list.md
```diff
@@ -79,6 +79,12 @@ RPC
 
 * AR of rpc layer
 
+Function Shipping
+---
+
+* Outdated document about how function shipping works in motr: [PDF](PDF/motr_function_shipping.pdf).
```
andriytk, line 85: I wouldn't say it's an outdated document. It's still quite up to date. Nothing was changed (in terms of Function Shipping) since the time it was written.

huanghua78: @andriytk Do you like the idea to remove the "Outdated document about"?

andriytk: I'd put it as "Function Shipping usage" or something.

huanghua78:
@andriytk Please take a look and approve if no more comments.

andriytk (approved):

PR #215, opened 2020-11-16
Fix dependency error at example1.c

daniarherikurniawan:
There are two updates:

1. `export LD_LIBRARY_PATH` is needed so that GCC can get the linked libraries. Otherwise, we will get this error:
`./example1: error while loading shared libraries: libmotr.so.1: cannot open shared object file: No such file or directory`

2. `#include "motr/idx.h"` is needed because this example uses `M0_IDX_DIX` which is located at `idx.h`. Otherwise, we will get this error:

```
example1.c: In function ‘main’:
example1.c:411:2: error: invalid use of undefined type ‘struct m0_idx_dix_config’
  motr_dix_conf.kc_create_meta = false;
  ^
example1.c:421:39: error: ‘M0_IDX_DIX’ undeclared (first use in this function)
  motr_conf.mc_idx_service_id        = M0_IDX_DIX;
                                       ^
example1.c:421:39: note: each undeclared identifier is reported only once for each function it appears in
```

Signed-off-by: Daniar Kurniawan <ddhhkk2@gmail.com>

### DIFF
--- a/motr/examples/example1.c
+++ b/motr/examples/example1.c
@@ -36,7 +36,8 @@
  * Please change the configuration according to you development environment.
  *
  * How to run:
- * LD_LIBRARY_PATH=/work/cortx-motr/motr/.libs/                              \
+ * LD_LIBRARY_PATH=/work/cortx-motr/motr/.libs/                              
+ * export LD_LIBRARY_PATH
  * ./example1 172.16.154.179@tcp:12345:34:1 172.16.154.179@tcp:12345:33:1000 \
  *         "<0x7000000000000001:0>" "<0x7200000000000001:64>" 12345670
  */
@@ -45,6 +46,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <errno.h>
+#include "motr/idx.h"
 
 static struct m0_client         *m0_instance = NULL;
 static struct m0_container       motr_container;


motr/examples/example1.c
```diff
@@ -36,7 +36,8 @@
  * Please change the configuration according to you development environment.
  *
  * How to run:
- * LD_LIBRARY_PATH=/work/cortx-motr/motr/.libs/                              \
+ * LD_LIBRARY_PATH=/work/cortx-motr/motr/.libs/                              
+ * export LD_LIBRARY_PATH
  * ./example1 172.16.154.179@tcp:12345:34:1 172.16.154.179@tcp:12345:33:1000 \
  *         "<0x7000000000000001:0>" "<0x7200000000000001:64>" 12345670
```
huanghua78, line 42: Original, my intention was to put these in the same line. 
Please refer to the back slash `\` at the end of LINE 39.

It is better to export it, so no need to add " LD_LIBRARY_PATH=/work/cortx-motr/motr/.libs/" at every run.

motr/examples/example1.c
```diff
@@ -45,6 +46,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <errno.h>
+#include "motr/idx.h"
```
huanghua78, line 49: Can you please show me how you compiled this file?

I did not see such error you mentioned in the commit message.
"motr/idx.h" in included in "motr/client.h". So, "motr/client.h" is the only header file a Motr app needs.

Can you confirm?

daniarherikurniawan: I add this example1.c manually and compile it while using the "dev" branch, perhaps that's why I got an error. The current main branch is broken, I hope this [PR213](https://github.com/Seagate/cortx-motr/pull/213) will be merged soon.

motr/examples/example1.c
daniarherikurniawan: I see, so you had `export LD_LIBRARY_PATH` in your system by default. However, when we install CORTX on the clean server (Emulab or Chameleon testbed cluster), we need to install all the dependency manually. Perhaps, adding a note saying that the `LD_LIBRARY_PATH` must be exported would be great. What do you think?

huanghua78: Developers are encouraged to do this by his own if he'd like to export globally LD_LIBRARY_PATH.
Explanation is already in the code itself.

motr/examples/example1.c
huanghua78: Please don't use "dev" branch any more. It is obsolete. It is not maintained and will not be updated.

Along with this example, 
>"motr/idx.h" in included in "motr/client.h". So, "motr/client.h" is the only header file a Motr app needs.
This is to minimize the requirements of header files.

If you work on the "dev" branch, it's OK to include the "motr/idx.h" in source file.
But this is not needed for main branch.

daniarherikurniawan:
Alright! I will close this PR since it is not applicable to the main branch. 
Thanks

PR #226, opened 2020-11-23
added binary files (m0hsm, example1) into .gitignore

huanghua78:


andriytk:
Why?

huanghua78:
> Why?

```
apr-u520428l001:cortx-motr huanghua$ git status
On branch main
Your branch is up to date with 'origin/main'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	hsm/m0hsm
	utils/example1
	utils/m0hsm
```

These are the results of build. If they are not included in ignore file, git status will give the above output.
So, we usually add them in a list of .gitignore.
The same to all build generate files, test generated files.

andriytk (approved):
Right, agree. I just was not sure whether `git clean -dfx` would clean it up then. It will - just checked.

Thanks!

nihar1313 (approved):

PR #242, opened 2020-11-25
doc/Quick-Start-Guide.rst: drop sudo at m0 make

andriytk:
sudo is not needed for m0 make/rebuild commands.


huanghua78 (approved):

VenkyOS (approved):
Approved.

PR #247, opened 2020-11-27
scripts/m0: fix sudo requirement for make/rebuild

andriytk:
sudo should not be required for make/rebuild commands.


huanghua78 (approved):

nihar1313 (approved):
Looks good.

PR #250, opened 2020-11-29
EOS-15338 motr-free-space-monitor: changed IEM alert message format 

yatin-mahajan:
Current libmotr api "m0_iem()" is generalize for all IEM alerts from motr.
Which is not allowing to change format of free space monitor alert message as
required, hence added print log (instead of api) to provide alert in required format.
Keeping exiting logic as well (commented) as it may required in LDR-R2

Signed-off-by: Yatin Mahajan <yatin.mahajan@seagate.com>

### DIFF
--- a/scripts/install/usr/libexec/cortx-motr/motr-free-space-monitor
+++ b/scripts/install/usr/libexec/cortx-motr/motr-free-space-monitor
@@ -197,14 +197,16 @@ if __name__ == '__main__':
                 cluster_usage_percentage >= iem_alert_high_thresold):
                 send_iem = True
 
+#           update idx as per current usage percent to get current next low from
+#           from threshold list.
+            idx = bisect.bisect_right(param_list, cluster_usage_percentage)
+            if idx >= list_len:
+                idx = list_len - 1
+            iem_alert_low_new = param_list[idx]
+
             if send_iem:
-                idx = bisect.bisect_right(param_list, int(iem_alert_low_thresold))
-                if idx >= list_len:
-                    idx = list_len - 1
-                iem_alert_low_new = param_list[idx]
-                if iem_alert_high_thresold > iem_alert_low_thresold:
-                    iem_alert_low_thresold = iem_alert_low_new
-                    print_iem(cluster_usage_percentage)
+                print_iem(cluster_usage_percentage)
+            iem_alert_low_thresold = iem_alert_low_new
             cluster_usage_prev = cluster_usage_percentage
         else:
             if cluster_usage_prev > iem_min_threshold:
--- a/scripts/install/usr/libexec/cortx-motr/motr-free-space-monitor
+++ b/scripts/install/usr/libexec/cortx-motr/motr-free-space-monitor
@@ -123,9 +123,16 @@ def print_iem(usage_percentage):
     evt_id = iem_event_id
     msg = f"Cluster size {usage_percentage}"
     msg_b = msg.encode('utf-8')
-    libmotr.m0_time_init()
-    libmotr.m0_iem(ctypes.c_char_p(file_parm_b), function_parm, line_num,
-                        sev_id, mod_id, evt_id, ctypes.c_char_p(msg_b))
+
+#   Current libmotr api "m0_iem()" is generalize for all IEM alerts from motr.
+#   Which is not allowing to provide free space monitor alert message as
+#   required format, hence added print log to provide alert in required format.
+#   Keeping exiting logic as well (commented) as it may required in LDR-R2
+
+    log.error(f"IEC: AS0020020002: Usable storage is {usage_percentage} % full");
+#    libmotr.m0_time_init()
+#    libmotr.m0_iem(ctypes.c_char_p(file_parm_b), function_parm, line_num,
+#                        sev_id, mod_id, evt_id, ctypes.c_char_p(msg_b))
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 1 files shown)


scripts/install/usr/libexec/cortx-motr/motr-free-space-monitor
```diff
@@ -123,9 +123,16 @@ def print_iem(usage_percentage):
...
-    libmotr.m0_time_init()
-    libmotr.m0_iem(ctypes.c_char_p(file_parm_b), function_parm, line_num,
-                        sev_id, mod_id, evt_id, ctypes.c_char_p(msg_b))
+
+#   Current libmotr api "m0_iem()" is generalize for all IEM alerts from motr.
+#   Which is not allowing to provide free space monitor alert message as
+#   required format, hence added print log to provide alert in required format.
+#   Keeping exiting logic as well (commented) as it may required in LDR-R2
+
+    log.error(f"IEC: AS0020020002: Usable storage is {usage_percentage} % full");
```
huanghua78, line 132: Do you mean, the log will be parsed by IEM and then be shown on GUI?

yatin-mahajan: Yes. It will parse though journal logs for IEM logs and will be shown to GUI.

scripts/install/usr/libexec/cortx-motr/motr-free-space-monitor
```diff
@@ -182,11 +190,11 @@ if __name__ == '__main__':
 
         cluster_usage_percentage = cluster_size[USAGE_PERCENTAGE]
         if cluster_usage_percentage > iem_min_threshold:
-            if (cluster_usage_prev < iem_alert_low_thresold and
-                cluster_usage_percentage > iem_alert_low_thresold):
+            if (cluster_usage_prev <= iem_alert_low_thresold and
```
huanghua78, line 193: The percentage of usage usually changes very slowly.
If the cluster_usage_prev == iem_alert_low_thresold and cluster_usage_percentage == iem_alert_low_thresold (both approximately), 
this will cause the alert to be sent multiple times?

yatin-mahajan: You, mean to say the where "cluster_usage_prev == cluster_usage_percentage == iem_alert_low_thresold", but whenever we send IEM alert we do updating low_threshold to new low, which updates "iem_alert_low_thresold" and that should failed if check at line 194 and won't send IEM alert again.

huanghua78: OK

huanghua78 (approved):

scripts/install/usr/libexec/cortx-motr/motr-free-space-monitor
```diff
@@ -182,11 +190,11 @@ if __name__ == '__main__':
...
+                cluster_usage_percentage >= iem_alert_low_thresold):
                 send_iem = True
-            if (cluster_usage_prev < iem_alert_high_thresold and
-                cluster_usage_percentage > iem_alert_high_thresold):
+            if (cluster_usage_prev <= iem_alert_high_thresold and
```
shashank-parulekar, line 196: The right way to do this is to first check for the maximum threshold and send IEMs if this threshold has reached and if not then look for the next lower threshold and if that has been reached then send the alert for this 2nd threshold till we reach the lowest threshold.
The reason for this is that suppose we have the threshold's defined as the current example of 75 80 85 90 and currently the storage has reached 93% having sent the alerts to the CSM at 75, 80, 85 and 90 things have been behaving properly.
Now at this point if the cluster is restarted then the code will again start to send the alerts again for 75, 80, 85 and 90 instead of just sending the alert to have crossed 90% mark !!!!
Since this is not a frequent occurrence we might make an exception for this one for now but we should work on this one for next version.

scripts/install/usr/libexec/cortx-motr/motr-free-space-monitor
```diff
@@ -182,11 +190,11 @@ if __name__ == '__main__':
 
         cluster_usage_percentage = cluster_size[USAGE_PERCENTAGE]
         if cluster_usage_percentage > iem_min_threshold:
```
shashank-parulekar, line 192: Considering the threshold example of : 70 80 85 90.
I see that if the storage is getting filled and exceeds 70% then we will send one alert and another when we cross over 80%. But at this point if the storage space is freed by doing deletes and the storage reaches 77% and again if the storage starts to fill (because of Object writes) and exceeds 80% then we will not be raising any alerts. Is that by design?
I  see that if the storage occupancy drops below 70% and then starts to increase again in which case we will be having an alert when we exceed the minimum threshold. Just confirm what I mention is by design and then that is fine. Or maybe you might want to ask Pradeep or someone to know the times when the alert needs to be triggered.

yatin-mahajan: I am not sure weather it is by design or not need to be confirmed with someone.

scripts/install/usr/libexec/cortx-motr/motr-free-space-monitor
yatin-mahajan: If the storage is 93% full and and in that case we restarted cluster or motr-free-space-monitor service alone, it sends only one alerts (i.e. Usable storage is 93 % full) and not all defined in threshold list. 
This is tested by restarting motr-free-space-monitor service alone.  Only thing is it sell send alert with whatever current usage. (i.e. 93% in this scenario)

shashank-parulekar: If it sends with current usage 93% then that is fine.

scripts/install/usr/libexec/cortx-motr/motr-free-space-monitor
yeshpal-jain-cs: if the storage size drops less than minimum threshold then only the alert would be sent, if they delete few things then CSM dashboard has a dial pad which shows the current usage.

shashank-parulekar (approved):

yeshpal-jain-cs (approved):

shashank-parulekar (approved):

huanghua78:
This PR is still needed to be merged to release branch.

PR #265, opened 2020-12-07
Vagrantfile: fix reading M0_VM_SHARE_TYPE env var

andriytk:
Currently, M0_VM_SHARE_TYPE env variable is not honoured due to the bug in the code.

Solution: fix the bug.


andriytk:
Included at #269.

PR #267, opened 2020-12-07
WIP Problem: notifying systemd most m0_cs_start() delays motr process startup

mssawant:
Solution:
Notify systemd after motr env setup.


huanghua78:
Hello @mssawant ,

Please check if this PR is still valid and needed.
If not, please close it.
If yes, please update the status.

huanghua78:
I don't think we still need it.

PR #269, opened 2020-12-07
Vagrantfile: add VMware support on Windows PC hosts

andriytk:
The series of patches here add support for VMware into Vagrantfile under Windows PC hosts.

Here is the list of changes:

* Vagrantfile: fix reading M0_VM_SHARE_TYPE env var
* Vagrantfile: add SMB-type shared dir support
* Vagrantfile: add vmware_desktop provider
* xcode/proto: fix directories traversing on smbfs

With all these patches I was able to provision and startup the cmu node in m0vg under VMware Workstation 14 on Windows 10 (1909) PC and build Motr from the source code directory shared via SMB from the Windows host.

-----------

More info about the patch for xcode/proto (which is not only about the Vagrantfile):

Currently, when Motr sources are mounted via SMB share xcode/proto does not produce any result and the build fails like this:

```
Checking motr public API... OK
Checking xcode protocol... FAIL
*ERROR* `xcode/proto` output differs from `xcode/m0protocol`
--- xcode/protocol-be.txt       2020-12-07 19:55:26.510082700 +0000
+++ protocol-be.txt     2020-12-07 20:01:59.338894900 +0000
@@ -1,4 +1,468 @@
 Motr binary protocol
 Domains: be
-Total structures: 0
+Total structures: 77
```

The problem is because Perl's File::Find module does not descend into subdirs on smbfs (refer to https://github.com/perl/perl5/issues/16759 issue).

Solution: set $File::Find::dont_use_nlink = 1, as suggested by the above link.


chumakd (approved):

scripts/provisioning/Vagrantfile
```diff
@@ -598,6 +598,43 @@ def setup(node, cpu_nr, mem_mb, disk_nr = 0, disk_size = 1)
     end
   end
 
+  # VMware Desktop --------------------------
```
huanghua78, line 601: What is the change for?
I don't see the relevance of this change and the title of this PR.

andriytk: @huanghua78 , I run Linux in VMware VMs on Windows PC under Vagrant. On Windows, the shared folder (where all the sources are stored) is mounted by SMB/Samba/CIFS protocol. (NFS is not supported by Vagrant.) This patch fixes Motr build when the sources are mounted via CIFS.

Hope this helps? If not, don't hesitate to ask more. For more information about provisioning VMs for Motr with Vagrant refer to https://github.com/Seagate/cortx-motr/tree/main/scripts/provisioning.

andriytk: I have updated description here also so that it would be more clear what this patch is for. I agree - it was not quite clear before.

andriytk: Ah, sorry - looks like you are asking about another patch. Yes, it had his own PR - https://github.com/Seagate/cortx-motr/pull/268. But I closed it as it seemed more convenient to have all these patches in one PR as they all are related.

huanghua78: OK.

I used Vagrant in the past, but not now. I can not verify this patch myself.

andriytk: No worries - I have verified it. :)

And @chumakd (the author of our Vagrantfile) has approved it also. We continue to work with @chumakd on further improvements, so there might be another patch(es) in future...

chumakd:
looks good to me

chumakd:
looks good

andriytk:
@huanghua78 (or whoever will do it) - please don't squash the commits of this PR whenever you will land it. Let them go separately. Thank you.

PR #272, opened 2020-12-09
EOS-15132: reset final extent if not part of free extent list

jadhavik1:
Assert occured in allocate_blocks() while checking final extent is
part of extent free list. It has found that best and final extent is
not part of extent list element. It has seen that one of extent in list
has enough space/freeblocks available but not considered.
balloc_wild_scan_group() gets final extent from best extent but
it does not guarantee that scanned best extent is part of free list of extent.
Reset final extent if balloc_wild_scan_group() unable to get correct
final extent which is not part of extent free list.

Co-authored-by: Venkateswarlu Payidimarry <venkateswarlu.payidimarry@seaga
Signed-off-by: Vikram Jadhav <vikram.jadhav@seagate.com>

### DIFF
--- a/balloc/balloc.c
+++ b/balloc/balloc.c
@@ -1554,6 +1554,56 @@ static bool is_extent_free(struct m0_balloc_group_info *grp,
 	return frags < zp->bzp_fragments;
 }
 
+static bool is_extent_partof_best_extent(struct m0_balloc_group_info *grp,
+					 struct m0_ext *best, uint64_t alloc_type,
+					 struct m0_ext **current)
+{
+	struct m0_lext              *le;
+	struct m0_balloc_zone_param *zp;
+	m0_bcount_t                  frags = 0;
+
+	M0_ENTRY();
+
+	zp = is_spare(alloc_type) ? &grp->bgi_spare : &grp->bgi_normal;
+
+	m0_list_for_each_entry(&zp->bzp_extents, le, struct m0_lext, le_link) {
+		*current = &le->le_ext;
+		if (m0_ext_is_partof(best, *current))
+			break;
+		++frags;
+	}
+	return frags < zp->bzp_fragments;
+}
+
+/**
+ * Reset final extent if current extent is part of best extent
+ *
+ * @param[in]  bac	  balloc allocation context
+ * @param[in]  cur	  current extent in extent list.
+ * @param[in]  alloc_type allocation type.
+ * @return true if reset final and best extent.
+ *	   false if current extent is not part of best extent
+ */
+static bool reset_final_extent(struct balloc_allocation_context *bac,
+			       struct m0_ext *cur, uint64_t alloc_type)
+{
+	m0_bindex_t    start = bac->bac_best.e_start;
+	m0_bcount_t    len   = m0_ext_length(&bac->bac_goal);
+
+	M0_ENTRY();
+	if (m0_ext_is_partof(&bac->bac_best, cur)) {
+		while (start < cur->e_start)
+			start += len;
+		if (start < cur->e_end && len <= cur->e_end - start) {
+			bac->bac_best = *cur;
+			bac->bac_final.e_start = start;
+			bac->bac_final.e_end = bac->bac_final.e_start + len;
+			return true;
+		}
+	}
+	return false;
+}
+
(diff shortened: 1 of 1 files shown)


balloc/balloc.c
```diff
@@ -1554,6 +1554,56 @@ static bool is_extent_free(struct m0_balloc_group_info *grp,
...
+	zp = is_spare(alloc_type) ? &grp->bgi_spare : &grp->bgi_normal;
+
+	m0_list_for_each_entry(&zp->bzp_extents, le, struct m0_lext, le_link) {
+		*current = &le->le_ext;
+		if (m0_ext_is_partof(best, *current))
+			break;
+		++frags;
+	}
+	return frags < zp->bzp_fragments;
+}
```
huanghua78, line 1576: Matching M0_LEAVE() is needed.

balloc/balloc.c
```diff
@@ -1554,6 +1554,56 @@ static bool is_extent_free(struct m0_balloc_group_info *grp,
...
+	m0_bcount_t    len   = m0_ext_length(&bac->bac_goal);
+
+	M0_ENTRY();
+	if (m0_ext_is_partof(&bac->bac_best, cur)) {
+		while (start < cur->e_start)
+			start += len;
+		if (start < cur->e_end && len <= cur->e_end - start) {
+			bac->bac_best = *cur;
+			bac->bac_final.e_start = start;
+			bac->bac_final.e_end = bac->bac_final.e_start + len;
```
huanghua78, line 1600: But the `best` is also modified (or reset). So, should we rename the function to `reset_final_best_extent()`?

balloc/balloc.c
```diff
@@ -1554,6 +1554,56 @@ static bool is_extent_free(struct m0_balloc_group_info *grp,
...
+			return true;
+		}
+	}
+	return false;
+}
```
huanghua78, line 1605: Matching M0_LEAVE() is neede.

balloc/balloc.c
```diff
@@ -2532,8 +2582,12 @@ static int allocate_blocks(int cr, struct balloc_allocation_context *bac,
 	if (rc == 0 && bac->bac_status == M0_BALLOC_AC_FOUND) {
 		if (len < m0_ext_length(&bac->bac_best))
 			balloc_new_preallocation(bac);
-		M0_ASSERT(is_extent_free(grp, &bac->bac_final, alloc_type,
-					 &cur));
+		if (!is_extent_free(grp, &bac->bac_final, alloc_type, &cur)) {
+			M0_ASSERT(is_extent_partof_best_extent(grp,
+				 &bac->bac_best, alloc_type, &cur));
+			M0_ASSERT(reset_final_extent(bac, cur, alloc_type));
+		}
```
huanghua78, line 2589: According to `doc/coding-style`:
>  * use assertions freely to verify state invariants. An asserted
       expression should have no side-effects;

The original code and the patch are both wrong.
These functions are modifying `cur`, and `the reset_final_extent()` function is modifying the 'final' and 'best'.

huanghua78: The `Root Cause` of this problem is not fixed.
The "&bac->bac_final" should be in the free list. We have tried to find a suitable extent from free space. But the result is not in the free list.
This means, something is wrong in the "balloc_simple_scan_group()" or balloc_wild_scan_group()".

The "&bac->bac_best" should be the best candidate from the free space list.
The "&bac->bac_best" is used in the search. It is not used as a result.
So, no need to modify it here.

jadhavik1: can we assert by taking return code, do you want me to remove this assert ? previously M0_ASSERT was there for is_extent_free(&cur), so added further assert if is_extent_free() return false.

jadhavik1: balloc_simple_scan_group() looks ok , problem seems with balloc_wild_scan_group() because somehow it gives incorrect best
extent, best extent should be lies in free list of extents.

huanghua78: Yes. The correct way is to take the return code, and assert the return code.

jadhavik1:
https://github.com/Seagate/cortx-motr/pull/281 another PR opened with proper root cause and fix

PR #282, opened 2020-12-14
EOS-14943 : Implement endpoint create using libfabirc APIs

upendrapatwardhan:


net/net.h
```diff
@@ -403,6 +403,9 @@ struct m0_net_domain {
          */
 	m0_bcount_t         nd_get_max_buffer_desc_size;
 
+	/** The network domain name which is to be initialiazed */
+	const char         *nd_name;
+
```
huanghua78, line 408: I don't think we need this.
We can use xprt->nx_name

net/libfab/libfab.c
```diff
@@ -61,54 +61,85 @@
...
+
+	M0_ENTRY();
+
+	if (name == NULL)
+		return M0_RC(0);
+
+	while ((fab != NULL) && (rc != 0)) {
+		rc = strcmp(name, fab->domain_attr->name);
+		if (rc == 0)
+			fi_list = fab;
```
huanghua78, line 91: Why do you update the variable `fi_list`?

This assignment has no effect outside this function.

net/net.h
upendrapatwardhan: The xprt->name  would be lnet/sock/libfabric which does not provide any info about the interface that we want to use.
Hence this was added. But after discussion with Nikita yesterday, there is some design change required for this. Hence will cancel this PR as of now

net/libfab/libfab.c
```diff
@@ -61,54 +61,85 @@
...
+static int libfab_dom_find(const char *name, struct fi_info *fi_list);
+static int libfab_ep_res_init(struct m0_fab__dom_param *fab,
+			      struct m0_fab__ep_param *ep);
+static struct m0_fab__ep_param *libfab_ep_net(struct m0_net_end_point *net);
+static int libfab_ep_find(const char *name, struct m0_net_transfer_mc *tm, 
+			  struct m0_net_end_point **epp);
+static void libfab_ep_param_free(struct m0_fab__ep_param *ep);
+static void libfab_ep_res_free(struct m0_fab__ep_res *ep_res);
+
+/* Find and select the domain from the list */
```
huanghua78, line 77: Add comments about the parameters and return value.

upendrapatwardhan:
After discussion with Nikita yesterday, there is some design change required for this. Hence cancelling this PR as of now

net/libfab/libfab.c
```diff
@@ -212,11 +247,129 @@ static int libfab_ma_stop(struct m0_net_transfer_mc *net, bool cancel)
...
+	M0_LEAVE();
+}
+
+/** Converts generic end-point to its libfabric structure. */
+static struct m0_fab__ep_param *libfab_ep_net(struct m0_net_end_point *net)
+{
+	return container_of(net, struct m0_fab__ep_param, fep_nep);
+}
+
+static int libfab_ep_find(const char *name, struct m0_net_transfer_mc *tm,
```
huanghua78, line 277: Please comment the parameter and return value.

If possible, it seems "bool" is enough for its return value.

net/libfab/libfab_internal.h
```diff
@@ -39,16 +40,41 @@ extern struct m0_net_xprt m0_net_libfab_xprt;
...
+};
+
+struct m0_fab__ep_res {
+	struct fi_cq_attr   *fer_cq_attr;   /* Completion Queue attributes */
+	struct fi_av_attr   *fer_av_attr;   /* Address Vector attributes */
+	struct fi_eq_attr   *fer_eq_attr;   /* Event Queue attributes */
+	struct fi_cntr_attr *fer_cntr_attr; /* Counter attributes */
+};
+
+struct m0_fab__ep_param {
```
huanghua78, line 64: This is the internal representative of a EP?
Then we can rename this to `struct m0_net_fab_ep`?

huanghua78: or `struct m0_fab_ep`?

PR #286, opened 2020-12-15
Eos 15552 ep decode

VinothVijayakumar-dev:
[EOS-15552] Implement API for endpoint address decode
        - API to decode node and port number from given ep_name


net/libfab/libfab.c
```diff
@@ -77,16 +79,49 @@ struct transfer_ma {
...
+
+	memcpy(node, ep_name, n);
+	node[n] = 0;
+
+	++cp;
+	n=strlen(cp);
+	memcpy(port, cp, n);
+	port[n] = 0;
+
+        return M0_RC(rc);
```
VinothVijayakumar-dev, line 114: alignment

VinothVijayakumar-dev:
discard due to in correct parent branch

PR #309, opened 2020-12-22
Vagrantfile: fix lustre-client provisioning

andriytk:
The regression was introduced at commit b0d129e77 when
we started checking for vmware when configuring lnet.conf.
The problem is that we check for the 'facter_virtual' but
facter is not installed by the time when the lustre-client
ansible task if performed. facter utility is installed
later during motr-runtime task.

Solution: check for 'ansible_virtualization_type' variable
instead of 'facter_virtual'. 'ansible_virtualization_type'
is internal to ansible and is independent of facter.


chumakd (approved):

chumakd:
looks good

huanghua78 (approved):

PR #328, opened 2021-01-14
process balloc request in FIFO order for a group

jadhavik1:
group try lock failed if already acquired by other thread
in this case wait until lock get released by adding request bac
at the end of  group specific bac list.
This changes will make sure request are processed in order of request.
This change made on top of RAID alignment changes.

### DIFF
--- a/balloc/balloc.c
+++ b/balloc/balloc.c
@@ -70,6 +70,8 @@ struct balloc_allocation_context {
 	uint32_t		       bac_scanned; /* groups scanned */
 	uint32_t		       bac_found;   /* count of found */
 	uint32_t		       bac_status;  /* allocation status */
+	struct m0_list_link            bac_link;  /* to add in group bac list */
+	struct m0_clink                bac_clink;
 };
 
 /** XXX @todo rewrite using M0_BE_OP_SYNC() */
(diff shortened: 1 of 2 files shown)


balloc/balloc.h
```diff
@@ -161,6 +161,14 @@ struct m0_balloc_group_info {
 	struct m0_lext              *bgi_extents;
 	/** per-group lock */
 	struct m0_be_mutex           bgi_mutex;
+
+	/** balloc allocation context list head */
+	struct m0_list               bgi_bac_head;
```
huanghua78, line 166: If possible, m0_tlist should be used instead of m0_list.
The same for "struct m0_list_link            bac_link; "

m0_tlist will enforce some type checking.

balloc/balloc.c
```diff
@@ -403,6 +405,10 @@ static int balloc_group_info_init(struct m0_balloc_group_info *gi,
 static void balloc_group_info_fini(struct m0_balloc_group_info *gi)
 {
 	m0_mutex_fini(bgi_mutex(gi));
+	m0_mutex_fini(&gi->bgi_bac_mutex);
+	m0_list_fini(&gi->bgi_bac_head);
+	m0_chan_fini_lock(&gi->bgi_chan);
+	m0_mutex_fini(&gi->bgi_chan_mutex);
 	m0_list_fini(&gi->bgi_normal.bzp_extents);
```
huanghua78, line 412: These lists should be replaced by m0_tlist.

This can be done in another ticket.

huanghua78:
How do we evaluate the performance impact of this change?

balloc/balloc.h
jadhavik1: ok

jadhavik1:
IO requests got very slow with this changes, so closing this experiment PR.

PR #332, opened 2021-01-14
go/mcp: add check for already existing object

andriytk:
Report error and exit if the object already exists in another pool, i.e. not the one specified by user via the `-pool` option.

Example:
```
$ mcp -prof 0x7000000000000001:0x4e9 -hax 172.18.1.22@o2ib:12345:1:1 -ep 172.18.1.22@o2ib:12345:4:1 -proc 0x7200000000000001:0x13a -pool 0x6f00000000000001:0x467 /dev/zero 12345:123458539
2021/01/19 14:15:21 mcp.go:107: the object already exists in another pool: 0x6f00000000000001:0x47f
```


huanghua78 (approved):

andriytk:
cc @chumakd

bindings/go/mcp/mcp.go
```diff
@@ -92,12 +92,21 @@ func main() {
...
+                log.Fatalf("invalid pool specified: %v", *pool)
+            }
+        }
         if err = mioW.Open(dst); err != nil {
             if err = mioW.Create(dst, objSize, *pool); err != nil {
                 log.Fatalf("failed to create object %v: %v", dst, err)
             }
         }
         defer mioW.Close()
+        if *pool != "" && !mioW.InPool(*pool) {
```
chumakd, line 106: Not clear why inversion (`!`) is needed for the `InPool()` return value. I would expect this condition to read like this:
```
if pool ID is not empty and object is present in that pool, then report an error "object already exist"
```

andriytk: @chumakd, no. You should read it: "if pool is not empty (i.e. was specified by user), but the object is *not* in that pool - report error".

PR #345, opened 2021-01-20
go/mio: fix getting the composite objs layout id

andriytk:
The 64bit layout id consists of 2 parts: the layout type and the id itself. The layout type is not zero for the composite objects, so M0_OBJ_LAYOUT_ID() macro must be used to get the id part.


andriytk:
cc @chumakd

huanghua78 (approved):

chumakd (approved):

PR #352, opened 2021-01-21
EOS-16421: [DTM0] Add client SM DLD.

ivan-alekhin:
The patch introduces a document that describes the changes
that are needed to be done on the client side
(Motr client, DIX, CAS) in order to enable DTM there.

### DIFF
--- a/dtm0/doc/dtx.txt
+++ b/dtm0/doc/dtx.txt
@@ -1,0 +1,515 @@
+--------------------------------------------------------------------------------
+# Overview
+
+This document defines a DLD for the client-side
+state machines touched by DTM0. It has the following sections:
+
+* Component diagram - defines the scope.
+* With/Without DTM enabled - defines states and high-level view of 
+  the transitions.
+* Low-level details on events - a human-readable description of the
+  previous section.
+* DTX connections in psudo-code - describes what needs to be changed
+  in the existing code and how the changes are connected to DTM.
+* Q&A - a list of questions to clarify design decisions.
+
+--------------------------------------------------------------------------------
+# Component diagram
+
+      * Client-side SMs
+     \|/
++------------------+                       +-------------+
+| M0Client/DIX/CAS |                       | DTM Service |
++------------------+                       +-------------+
+        |                                       |
+        +---+-----------------------------------+ <-* "Interconnections" 
+            |
+           \|/
+        +----------+
+        | DTM0 log |
+        +----------+
+
+The scope of this document is limited by "Interconnections" and 
+"Client-side SMs".
+
+--------------------------------------------------------------------------------
+# Terms and definitions
+
+M0Client - Motr client;
+m0_op - Motr client operation;
+DIX-req - DIX request;
+CAS-req - CAS request;
+pa - participant;
+DTX - distributed transaction;
+DTX-pa - state of a participant inside a DTX.
+
+--------------------------------------------------------------------------------
+# Without DTM enabled
+
+This section is an overview of how Client-side SMs are defined
+right now.
+
+SM states:
+
+M0client-op: 
+	Init
+	Launched
+	Executed
+	Stable
+	Failed
+
+DIX-req:
+	Init 
+	 ... (discovery) ... 
+	InProgress(N)
+	Final
+	Failure
+
+CAS-req: Init, Sent, Final, Failure
+
+SM Events:
+
+launch(m0c):
+	m0c: Init -> Launched
+	dix: Init
+	cas: Init
+	ast: async call dix_send
+
+dix_send(dix):
+	dix: Init -> InProgress
+	cas: Init -> Sent
+	rpc: call cas_repl on reply
+
+cas_repl(cas[0]):
+	cas: Sent -> Final
+	dix: InProgress, nr_received++
+
+cas_repl(cas[N-1]):
+	cas: Sent -> Final
+	dix: InProgress -> Final
+	m0c: Launched -> Executed -> Stable
+
+--------------------------------------------------------------------------------
+# With DTM enabled
+
+This sections describes the proposed changes in the SMs.
+
+SM states:
+
+M0client-op: 
+	Init
+	Launched
+	Executed
+	Stable
+	Failed
+
+DIX-req:
+	Init 
+	 ... (discovery) ... 
+	InProgress(N)
+	Final
+	Failure
+
+CAS-req: Init, Sent, Final, Failure
+
+# State of a participant. A transaction has an array of states. Each element
+# of such array describes the state of a participant of the transaction.
+# For example, "dtx: [InProgress]" means that
+# the transaction has not been executed yet but the corresponding
+# requests have been prepared (and they may have been already sent).
+# "dtx: [InProgress, Executed, InProgress]" means that the inferred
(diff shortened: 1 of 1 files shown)


dtm0/doc/dtx.txt
```diff
@@ -0,0 +1,409 @@
...
+	post ast dtm_service.sm_group process_persistent_notice(fop);
+}
+
+
+--------------------------------------------------------------------------------
+# NOTES
+
+CS should check the previous timestamp, and it should wait until the value
+changes. It could be a busy loop.
+
```
sergey-shilov, line 393: I propose to add one more note:

We need to preserve existing logic without DTM by adding of conditions like
        if (op->op_dtx)
                ......
by the following reasons:
 - we have a read requests that go without DTX but use common code;
 - we have a meta requests that possibly will not use DTX;
 - we need to support backward compatibility for easier testing and other purposes.

dtm0/doc/dtx.txt
```diff
@@ -0,0 +1,409 @@
...
+struct dtm_service {
+	struct dtm_log   ds_log;
+	struct sm_group *ds_sm_group;
+};
+
+ast_cb client_dtx_stable(op,...) {
+	struct m0_op *op;
+	
+	op->op_stable();
+	op->state = Stable;
```
sergey-shilov, line 327: set_state()?

dtm0/doc/dtx.txt
```diff
@@ -0,0 +1,409 @@
...
+	bool pa_executed; // got reply
+	bool pa_persistent; // got PERSISTENT notice
+};
+
+struct txr {
+	struct pa t_pa[];
+	/// ... version, tx_id and so on
+};
+
+strcut record {
```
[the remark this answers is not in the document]
ivan-alekhin: Addressed

dtm0/doc/dtx.txt
```diff
@@ -0,0 +1,409 @@
...
+# NOTES
+
+CS should check the previous timestamp, and it should wait until the value
+changes. It could be a busy loop.
+
+--------------------------------------------------------------------------------
+# Q&A
+
+Q1: Is M0Client "clinked" with dixreq? or is it clinked with dtx.sm?
+A2: it should be linked with dtx.
```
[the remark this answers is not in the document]
ivan-alekhin: Fixed

dtm0/doc/dtx.txt
ivan-alekhin: Added & updated

dtm0/doc/dtx.txt
ivan-alekhin: Yes, fixed

dtm0/doc/dtx.txt
```diff
@@ -0,0 +1,409 @@
...
+struct m0_op {
+	struct dix_req  op_dix_req;
+	// ...
+	struct dtx     *op_dtx;
+};
+
+m0op_invariant(op):
+	// Assume: when DTM0 flag is disabled,
+	//         op_dtx must be zeroed.
+	ergo(!DTM0, op_dtx == NULL)
```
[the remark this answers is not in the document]
ivan-alekhin: Fixed

sergey-shilov (approved):

dtm0/doc/dtx.txt
```diff
@@ -0,0 +1,441 @@
...
+	cas: Sent -> Final
+	dix: InProgress, nr_received++
+
+cas_repl(cas[N-1]):
+	cas: Sent -> Final
+	dix: InProgress -> Final
+	m0c: Launched -> Executed -> Stable
+
+--------------------------------------------------------------------------------
+# With DTM enabled
```
just-now, line 72: General notice:

Please add FAILED states also.

dtm0/doc/dtx.txt
```diff
@@ -0,0 +1,441 @@
...
+
+SM states:
+
+M0client-op: Init -- Launched -- Executed -- Stable
+
+DIX-req: Init --- ... (discovery) ... -- InProgress(N) -- Final
+
+CAS-req: Init -- Sent -- Final
+
+DTX-pa: Init -- Prepared -- InProgress -- Executed -- Persistent -- Stable
```
just-now, line 84: what is `-pa'?

dtm0/doc/dtx.txt
```diff
@@ -0,0 +1,441 @@
...
+	struct dtx     *dr_dtx; // borrowed from op->op_dtx
+};
+
+
+dix_req_create(...) {
+	if (op->dtx == NULL) {
+		/* Preserve non-DTM logic */
+		clink_init(..., dix_req_clink_cb);
+	} else {
+		clink_init(..., dtx_clink_cb);
```
just-now, line 246: Probably we'd need to be subscribed on both dix_req_clink_cb and dtx_clink_cb. JFYI.

dtm0/doc/dtx.txt
```diff
@@ -0,0 +1,441 @@
...
+
+* Having the ability to switch between these two modes (with/without DTM) helps
+  with testing and debugging. Also, it makes easier to propagate the changes
+  into the upstream.
+
+--------------------------------------------------------------------------------
+# Q&A
+
+Q1: Is M0Client "clinked" with dixreq? or is it clinked with dtx.sm?
+A1: it should be linked with dtx.
```
just-now, line 430: even both callbacks my trigger it

dtm0/doc/dtx.txt
```diff
@@ -0,0 +1,441 @@
...
+Q1.1: Should we extend the set of states of dixreq instead of adding dtx.sm?
+A1.1: We keep using dtx; no need to change dixreq states.
```
just-now, line 432: correct.

dtm0/doc/dtx.txt
```diff
@@ -0,0 +1,441 @@
...
+Q2: Is Executed == got-one-reply or Executed == got-all-replies?
+A2: We assume a dtx is Executed when we got at least one reply.
```
just-now, line 434: Generally, the result of such execution is known and returned to the dtx user... 
`at least one reply' is correct for current implementation but with spares or overwrites it may change. So it's better to use the first one.

dtm0/doc/dtx.txt
```diff
@@ -0,0 +1,441 @@
...
+Q3: How to protect DTM0 log from concurrent modifications by DTM service and
+    by the client-side?
+A3: DTM0 service will take care of it. It will provide an API to properly
```
just-now, line 437: It'll be around locks and ASTs.

just-now (commented):
Looks fine.
What's missing:
 - define triggers exposed from DTX SM which will trigger Clovis SM explicitly.
 - define how to trigger Clovis SM with both dix_req and dtx.

dtm0/doc/dtx.txt
ivan-alekhin: Added (partially) into some of the machines

dtm0/doc/dtx.txt
ivan-alekhin: PA == participant. I added "terms and definitions" section to cover this.

PR #354, opened 2021-01-22
motr.conf: fix performance of a single-threaded client

andriytk:
Currently, the value of MOTR_M0D_BETXGR_FREEZE_TIMEOUT_MIN parameter
is set to 300ms which makes the performance of a single-threaded
client about 12 MB/s on 2+1 pool of 4 devices.

MOTR_M0D_BETXGR_FREEZE_TIMEOUT_MIN parameter specifies how long to
wait before closing the BE transaction group when there are no more
transactions waiting for the group to be closed. So if this parameter
is set too high, the server will wait for more write fops to come
before closing the transaction group, thus affecting the performance.

Solution: decrease MOTR_M0D_BETXGR_FREEZE_TIMEOUT_MIN value to 1ms.
This will boost the performance on low loads from one side and,
won't degrade the performance on high loads, which is regulated by the MOTR_M0D_BETXGR_FREEZE_TIMEOUT_MAX parameter,
from another side. So we are in win-win position here.

Note: 1ms used to be the default value of this parameter for years
in /etc/sysconfig/motr and it's not clear now why and by whom it was
changed to 300ms at cortx/motr/conf/motr.conf - the file used by
provisioner to configure parameters at /etc/sysconfig/motr.

Tested on SAGE-cluster.


huanghua78:
Fantastic!
Have you done some tests to show performance boost?

andriytk:
@huanghua78 Yes. The performance was increased from 12 MB/s to 120 MB/s.

huanghua78:
```
apr-u520428l001:cortx-motr huanghua$ git grep MOTR_M0D_BETXGR_FREEZE_TIMEOUT_MIN
scripts/install/etc/sysconfig/motr:#MOTR_M0D_BETXGR_FREEZE_TIMEOUT_MIN=1
scripts/install/opt/seagate/cortx/motr/conf/motr.conf:MOTR_M0D_BETXGR_FREEZE_TIMEOUT_MIN=300
scripts/install/usr/libexec/cortx-motr/motr-service.functions:    if [[ -n $MOTR_M0D_BETXGR_FREEZE_TIMEOUT_MIN ]]; then
scripts/install/usr/libexec/cortx-motr/motr-service.functions:        optional_params+=" -y $MOTR_M0D_BETXGR_FREEZE_TIMEOUT_MIN"
apr-u520428l001:cortx-motr huanghua$ 
apr-u520428l001:cortx-motr huanghua$ git grep MOTR_M0D_BETXGR_FREEZE_TIMEOUT_MAX
scripts/install/etc/sysconfig/motr:#MOTR_M0D_BETXGR_FREEZE_TIMEOUT_MAX=50
scripts/install/opt/seagate/cortx/motr/conf/motr.conf:MOTR_M0D_BETXGR_FREEZE_TIMEOUT_MAX=300
scripts/install/usr/libexec/cortx-motr/motr-service.functions:    if [[ -n $MOTR_M0D_BETXGR_FREEZE_TIMEOUT_MAX ]]; then
scripts/install/usr/libexec/cortx-motr/motr-service.functions:        optional_params+=" -Y $MOTR_M0D_BETXGR_FREEZE_TIMEOUT_MAX"

```
@andriytk would you like also to modify the values in "scripts/install/etc/sysconfig/motr"?
@madhavemuri Is this file "scripts/install/etc/sysconfig/motr" still used somewhere?

huanghua78:
> @huanghua78 Yes. The performance was increased from 12 MB/s to 120 MB/s.

We also need to update ST scripts to have this parameter set properly.

andriytk:
> would you like also to modify the values in "scripts/install/etc/sysconfig/motr"?
@huanghua78 it is already set there to 1ms (default value), as far as I can see.

madhavemuri:
I think motr.conf is sufficient, @just-now may have better idea why this value is changed.
All BE params are changed based on performance testing done for R1.

Now with the default value we have to check the performance again on R1 setup.

andriytk:
I propose to merge it into main (R2) for now and keep an eye on it whenever it will be tested for the performance.

As of now, I've tested it on the SAGE-cluster. It does not seem to impact/degrade the high-load performance. But it considerably improves the low-load performance.

madhavemuri (approved):

PR #358, opened 2021-01-22
mcp: add -off offset option

andriytk:
Add `-off` option to start i/o on the Motr object from/at the specified offset.

### DIFF
--- a/bindings/go/mcp/mcp.go
+++ b/bindings/go/mcp/mcp.go
@@ -40,16 +40,18 @@ func usage() {
     flag.PrintDefaults()
 }
 
+var objOff  int64
 var objSize uint64
 var bufSize int
-var pool    *string
+var pool   *string
 
 func init() {
     log.SetFlags(log.Ldate | log.Ltime | log.Lshortfile)
 
     flag.Usage = usage
-    flag.IntVar(&bufSize, "bsz", 32, "I/O buffer `size` (in Mbytes)")
-    flag.Uint64Var(&objSize, "osz", 0, "object `size` (in Kbytes)")
+    flag.IntVar(&bufSize, "bsz", 32, "i/o buffer `size` (in MiB)")
+    flag.Int64Var(&objOff, "off", 0, "start object i/o at `offset` (in KiB)")
+    flag.Uint64Var(&objSize, "osz", 0, "object `size` (in KiB)")
     pool = flag.String("pool", "", "pool `fid` to create object at")
 }
 
@@ -59,6 +61,7 @@ func main() {
         usage()
         os.Exit(1)
     }
+    objOff  *= 1024
     objSize *= 1024
 
     src, dst := flag.Arg(0), flag.Arg(1)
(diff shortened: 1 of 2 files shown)


andriytk:
cc @chumakd

bindings/go/mcp/mcp.go
```diff
@@ -40,16 +40,18 @@ func usage() {
...
 
 func init() {
     log.SetFlags(log.Ldate | log.Ltime | log.Lshortfile)
 
     flag.Usage = usage
-    flag.IntVar(&bufSize, "bsz", 32, "I/O buffer `size` (in Mbytes)")
-    flag.Uint64Var(&objSize, "osz", 0, "object `size` (in Kbytes)")
+    flag.IntVar(&bufSize, "bsz", 32, "i/o buffer `size` (in MiB)")
+    flag.Int64Var(&objOff, "off", 0, "start object i/o at `offset` (in KiB)")
+    flag.Uint64Var(&objSize, "osz", 0, "object `size` (in KiB)")
```
huanghua78, line 54: Why are the "objOff" and "objSize" in KiB, but the bufSize in "MiB"?

huanghua78 (approved):

bindings/go/mcp/mcp.go
huanghua78: For utility "dd(1)", it has:
```
DD(1)                                                               User Commands                                                               DD(1)

NAME
       dd - convert and copy a file

SYNOPSIS
       dd [OPERAND]...
       dd OPTION


       seek=N skip N obs-sized blocks at start of output

       skip=N skip N ibs-sized blocks at start of input
```

How would you differentiate these two cases?

andriytk: In `dd` the offset is specified for the source/input (`skip`) and destination/output (`seek`), while in our case `-off` option is only for the Motr object (whether it is source or destination).

I was thinking about developing `mdd` utility later. There we might have skip and seek. :)

About objOff/Size in KiB and bufSize in MiB - it's all about convenience. Objects cannot be less than 4K (page size) in Motr, right? So 1K granularity for objects sizes/offsets is driven by this. From another side, the default buffer size in `mcp` (when you don't specify it) is 32 MiB already, so there is no point to do it in KBs.

Convenience above all!

huanghua78: But the mcp can have both the "src" and "dst" as "Motr object".
In this case, the offset will be applied to both objects, right?
I just want to confirm this use case.

andriytk: Yes. As help suggests: "start object i/o at `offset` (in KiB)". It does not matter whether it is src or dst object.

huanghua78: If the src is a Motr object, and at the same time, dst is also a Motr object, the "off" will apply to both src and dst?

andriytk: Yes, as I said in my prev. comment.

huanghua78: OK.

andriytk: The "object" notion here implies only Motr object.

siningwuseagate (approved):

PR #362, opened 2021-01-26
lib: introduce m0_buf_new_aligned()

andriytk:
Currently, to create a new aligned-allocated buffer one has to init some local buf 1st and then use m0_buf_copy_aligned() from it. For example:

```c
       struct m0_buf     buf_local;

       m0_buf_init(&buf_local, &data, sizeof data);
       *rc = m0_buf_copy_aligned(out, &buf_local, M0_0VEC_SHIFT);
```

Having m0_buf_new_aligned() allows to do this directly with one function call without additional local variable:

```c
       *rc = m0_buf_new_aligned(out, &data, sizeof data, M0_0VEC_SHIFT);
```

See an example at https://github.com/Seagate/cortx-motr-apps/pull/20.


lib/buf.c
```diff
@@ -109,13 +121,7 @@ M0_INTERNAL int m0_buf_copy_aligned(struct m0_buf *dst,
...
 {
 	M0_PRE(dst->b_nob == 0 && dst->b_addr == NULL);
-
-	M0_ALLOC_ARR_ALIGNED(dst->b_addr, src->b_nob, shift);
-	if (dst->b_addr == NULL)
-		return M0_ERR(-ENOMEM);
-	dst->b_nob = src->b_nob;
-	memcpy(dst->b_addr, src->b_addr, src->b_nob);
-	return 0;
+	return m0_buf_new_aligned(dst, src->b_addr, src->b_nob, shift);
```
huanghua78, line 124: We can define the src as const:
```
const struct m0_buf *src,
```

This is similar to "m0_buf_copy()".

And do you want to simplify the "m0_buf_copy()" to share common code?

andriytk: Good spot about `const`, will do. About `m0_buf_copy()` - not sure...

huanghua78 (approved):

nikitadanilov (approved):

siningwuseagate (approved):

PR #365, opened 2021-01-28
EOS-16866: [DTM0] Introduce tx_desc.

ivan-alekhin:
Transaction descriptor is an on-wire/on-disk
structure used to propagate DTM0-related
information accoss the network (rpc) and to preserve
it across reboots (be).
The tx_desc module defines the corresponding
data types and associated functions.


Anilsindhi (approved):
LGTM, +1.

PR #371, opened 2021-01-28
Update Quick-Start-Guide.rst

johnbent:
Signed-off-by: John Bent <john.bent@seagate.com>

Adding link to @andriytk new release.  Thanks Andriy!

### DIFF
--- a/doc/Quick-Start-Guide.rst
+++ b/doc/Quick-Start-Guide.rst
@@ -1,7 +1,9 @@
 =================
 Quick Start Guide
 =================
-This guide provides information to get Motr component ready.
+This guide provides information to get Motr component ready.  For your convenience to build and develop motr, 
+we offer a [CentOS-7.9-based Development VM](https://github.com/Seagate/cortx-motr/releases/tag/ova-centos79) 
+with all dependencies installed ready to build and start single-node Motr cluster.
 
 ************
 Prerequisite


huanghua78 (approved):

huanghua78 (approved):

doc/Quick-Start-Guide.rst
```diff
@@ -1,7 +1,9 @@
 =================
 Quick Start Guide
 =================
-This guide provides information to get Motr component ready.
+This guide provides information to get Motr component ready.  For your convenience to build and develop motr, 
+we offer a [CentOS-7.9-based Development VM](https://github.com/Seagate/cortx-motr/releases/tag/ova-centos79) 
+with all dependencies installed ready to build and start single-node Motr cluster.
```
andriytk, line 6: Looks like this section conflicts a bit with the following in prerequisites in terms of the kernel version and CentOS. I'd suggest moving it there and edit the existing prerequisites.

andriytk (changes requested):

doc/Quick-Start-Guide.rst
huanghua78: Since we officially support Cent OS 7.8, I'd suggest to deferring this.

andriytk: I thought, we officially support all CentOS 7.x versions. Is there an official page where it would be declared what we support?

Note: on the SAGE cluster we run the latest CentOS 7.9 versions on all the nodes.

huanghua78: The Lab machines and VMs are CentOS 7.8, I think. But as you said, the CentOS 7.x are all supported.
But to make things simple, CentOS 7.8 is the default supported OS version.
That's my understanding.

Let's open this PR again when we switch to Cent OS 7.9.

PR #373, opened 2021-01-29
build-prep-1node.sh: improve to use by non-root users

andriytk:
It's still required that the user have sudo privileges to run the script though.

### DIFF
--- a/scripts/provisioning/build-prep-1node.sh
+++ b/scripts/provisioning/build-prep-1node.sh
@@ -1,24 +1,64 @@
 #!/bin/sh
 
 # Clone, build and prepare to start a single-node cluster.
+#
+# Usage: ./build-prep-1node.sh [-dev]
+#
+#   -dev  development mode, don't build and install rpms
+#
+set -e -o pipefail
 
-set -eu -o pipefail
+dev_mode=
 
-[[ -d cortx-motr ]] ||
+[[ $1 == "-dev" ]] && dev_mode='yes'
+
+[[ -d cortx-motr ]] || {
     git clone --recurse https://github.com/Seagate/cortx-motr.git &&
         ln -s cortx-motr motr
+}
 cd motr
-echo 'Building and installing Motr...'
-./autogen.sh && ./configure --disable-expensive-checks && make -j4 &&
+echo 'Configure Motr...'
+[[ -f configure ]] && git clean -dfx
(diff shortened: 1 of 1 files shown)


nezzzu:
Also, if we can have 3 pools and if the hctl status output resembles that of Sage cluster then same motraddr.sh script in m0client-sample-apps can be used. @andriytk ?

nezzzu:
A further improvement would be to completely skip the build step if the required rpms are already installed on the system (or just an option to skip the build step) and go directly to single node setup steps. This will probably require the singlenode.yaml contents incorporated in the script if they are not rpm installed by default. @andriytk ?

andriytk:
@nezzzu this is devvm - for development purposes, not just install and run.

m0client-sample-apps can be used even now without multiple pools, one pool is enough. The rc-files for the utilities can be created manually.

Not familiar with motraddr.sh script, cannot advise on that.

huanghua78 (commented):
13  ./scripts/install-motr-service

21 make && make devinstall

also need `sudo`, I think.

andriytk:
Yes, thanks @huanghua78 .

huanghua78 (approved):

andriytk:
@nezzzu as you suggested - it builds and installs rpms by default now. The development mode is still available via the `-dev` cmdline option.

Give it a try.

scripts/provisioning/build-prep-1node.sh
```diff
@@ -1,31 +1,72 @@
...
 # Clone, build and prepare to start a single-node cluster.
+#
+# Usage: ./build-prep-1node.sh [-dev]
+#
+#   -dev  development mode, don't build and install rpms
+#
+set -e -o pipefail
 
-set -eu -o pipefail
+dev_mode=
```
huanghua78, line 11: Add comments like:
# set this to "YES" to build source code in local dir and then install only motr services
# and leave it empty to build RPMs and install them.

huanghua78 (approved):

scripts/provisioning/build-prep-1node.sh
andriytk: Hmm, isn't it enough what's written in the "Usage" above - "development mode, don't build and install rpms"?

madhavemuri (approved):

PR #376, opened 2021-02-01
EOS-16179: [DTM0] DTM0 Service improvements

mehjoshi:
This PR contains code changes for addressing EOS-16179.


dtm0/dtm0.h
```diff
@@ -0,0 +1,116 @@
...
+ * - DT_REDO: for a dtm0 redo-request operation
+ * Each dtm0_op also carries a dtm_msg type which tells what kind of message it
+ * is carrying.  A dtms0_op containing a request must specify a message type of
+ * either DMT_EXECUTE_DTX,indicating that this request contains an operation to
+ * be executed, a m0_dtms0_op containing a reply operation should contain either
+ * a DMT_EXECUTED or a DMT_PERSISTENT message.
+ */
+
+/**
+ * @{
```
ivan-alekhin, line 96: We need to add an API that allows us to initialize a DTM0 service instance (DTM0 service + DTM0 log + clock source). It will be used inside motr/client_init.c on the originator side. At this moment it is possible to initialize DTM0 service only on the server side (motr/init.c).

ivan-alekhin (commented):
Thanks for the patch. I posted a few suggestions.

dtm0/fop.h
```diff
@@ -37,12 +50,52 @@ extern const struct m0_rpc_item_ops dtm0_req_fop_rpc_item_ops;
...
+ */
+enum m0_dtm0s_op_flags {
+	/**
+	 * Define any operation flags
+	 */
+	DOF_NONE = 0,
+	/**
+	 * Delay reply until transaction is persisted.
+	 */
+	DOF_SYNC_WAIT = 1 << 0,
```
just-now, line 64: what's the usecase for this flag?

dtm0/fop.h
```diff
@@ -37,12 +50,52 @@ extern const struct m0_rpc_item_ops dtm0_req_fop_rpc_item_ops;
...
+	DT_REDO,
+	DT_NR,
+} M0_XCA_ENUM;
+
+enum m0_dtm0s_msg {
+	DMT_EXECUTE_DTX,
+	DMT_EXECUTED,
+	DMT_PERSISTENT,
+	DMT_NR,
+} M0_XCA_ENUM;
```
just-now, line 82: what about to merge all these into a single enum?
DMT_EXECUTE,
DMT_EXECUTED,
DTM_PERSISTENT,
DMT_REDO

just-now (approved):
Looks good to me.

dtm0/dtm0.h
ivan-alekhin: Just to clarify: it was just a general comment, we don't have to address it in the scope of this patch. It looks like it will be covered (at least partially) by this patch (https://github.com/Seagate/cortx-motr/pull/382). There are `m0_dtm__client_service_start` and `m0_reqh_service_lookup`.

cas/cas.h
```diff
@@ -293,7 +293,9 @@ struct m0_cas_op {
 	 *
 	 * Array should be non-empty.
 	 */
-	struct m0_cas_recv cg_rec;
+	struct m0_cas_recv  cg_rec;
+	/* dtm0 txr payload */
+	struct m0_dtm0_txr *txr;
```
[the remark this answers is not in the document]
mehjoshi: Done.

dtm0/dtm0.h
mehjoshi: OK. Will review m0_dtm__client_service_start routine from the above patch and then discuss what is missing and what needs to be added.

dtm0/fop.h
mehjoshi: This change is leftover from my earlier changes. It is not needed anymore. Will remove it. Thanks!

dtm0/fop.h
mehjoshi: Done.

PR #379, opened 2021-02-01
Fix iem alert formatting

shashank-parulekar:
Code changes are done to make sure the IEM messages, generated in case of BE data structure corruption, are meaningful to the customer so that he can take appropriate actiions.


format/format.c
```diff
@@ -137,15 +133,16 @@ M0_INTERNAL int m0_format_footer_verify_generic(
 	M0_PRE(footer != NULL);
 
 	if (footer->ft_magic != M0_FORMAT_FOOTER_MAGIC) {
-		format_iem_post("format footer magic mismatch,",
-				(uint64_t)M0_FORMAT_FOOTER_MAGIC,
-				footer->ft_magic, iem);
+		format_iem_post("Internal data appears to be inconsistent and "
```
huanghua78, line 136: In this case, do we need to log something into m0trace log?
For example, the expected "M0_FORMAT_FOOTER_MAGIC" and the "footer->ft_magic",
and other information about the buffer.

So, when Seagate support analyzes or inspects this issue, more information can be used to locate the data.

huanghua78 (approved):

format/format.c
huanghua78: OK

mukundkanekar (approved):

yeshpal-jain-cs (approved):

huanghua78 (approved):

yeshpal-jain-cs (approved):

PR #383, opened 2021-02-01
EOS-17070 ISA-L: Integrate ISAL encode APIs and check performance.

sanjognaik:
EOS-17070 ISA-L: Integrate ISAL encode APIs and check performance.

- Added option 'run-ub' to m0 script to run unit benchmarking.
- Used Intel ISA arithmetic APIs in m0_parity_mul() and m0_parity_div() functions.
- Intel ISA APIs are accessed in user space only.
- Galois APIs are called only in kernel space for now.
- Integrated ISA encoding APIs
- Implemented isal_encode() and isal_recover() functions.
- Updated ub tests for parity_math.
- Increased number of iterations for parity_math ub tests to 100.
- Added function for differential parity calculation using Intel ISA API.

### DIFF
--- a/sns/parity_math.c
+++ b/sns/parity_math.c
@@ -311,18 +311,14 @@ M0_INTERNAL int m0_parity_math_init(struct m0_parity_math *math,
 
 	M0_SET0(math);
 
-	math->pmi_data_count	      = data_count;
-	math->pmi_parity_count	      = parity_count;
+	math->pmi_data_count	= data_count;
+	math->pmi_parity_count	= parity_count;
 
-        if (parity_count == 1) {
+	if (parity_count == 1) {
 		math->pmi_parity_algo = M0_PARITY_CAL_ALGO_XOR;
 		return 0;
 	} else {
 		math->pmi_parity_algo = M0_PARITY_CAL_ALGO_REED_SOLOMON;
-		/*
-                 * init galois, only first call makes initialisation,
-		 * no de-initialisation needed.
-		 */
 
 		ret = vandmat_init(&math->pmi_vandmat, data_count,
 				   parity_count);
--- a/sns/parity_math.c
+++ b/sns/parity_math.c
@@ -18,6 +18,11 @@
  *
  */
 
+#ifndef __KERNEL__
+#include <isa-l.h>
+#endif /* __KERNEL__ */
+
+#define DEBUG 0
 
 #include "lib/arith.h"
 #include "lib/assert.h"
@@ -32,6 +37,12 @@
 #define M0_TRACE_SUBSYSTEM M0_TRACE_SUBSYS_SNS
 #include "lib/trace.h"
 
+#if DEBUG
+#ifndef __KERNEL__
+#include <stdio.h>
+#endif /* __KERNEL__ */
+#endif /* DEBUG */
+
 #define ir_invalid_col_t UINT8_MAX
 
 /* Forward declarations */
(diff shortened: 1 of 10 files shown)


sns/parity_math.c
```diff
@@ -670,6 +858,168 @@ static void xor_recover(struct m0_parity_math *math,
...
+	uint8_t *err_list = NULL;
+	uint8_t s;
+	int ret;
+
+	fail = (uint8_t*) fails->b_addr;
+	fail_count = fails_count(fail, unit_count);
+
+#if DEBUG
+	printf("\n fail_count: %d", fail_count);
+#endif
```
huanghua78, line 887: No need to use printf();

M0_LOG(M0_ALWAYS, ...) is good enough.
Or
m0_console_printf() is also OK.

sns/parity_math.c
```diff
@@ -670,6 +858,168 @@ static void xor_recover(struct m0_parity_math *math,
...
+	}
+
+	for (ui = 1; ui < math->pmi_data_count; ++ui)
+		M0_ASSERT(block_size == data[ui].b_nob);
+
+	for (ui = 0; ui < math->pmi_parity_count; ++ui)
+		M0_ASSERT(block_size == parity[ui].b_nob);
+
+	M0_ALLOC_ARR(temp_matrix, (unit_count * math->pmi_data_count));
+	M0_ASSERT(temp_matrix != NULL);
```
huanghua78, line 934: All of these error (Out-of-memory) should be handled properly.
No M0_ASSERT() here.

sanjognaik: Created another ticket to return error codes for these functions.

sns/parity_math.c

sns/parity_math.c
```diff
@@ -297,12 +346,16 @@ M0_INTERNAL void m0_parity_math_fini(struct m0_parity_math *math)
...
+#endif /* __KERNEL__ */
 }
 
 M0_INTERNAL int m0_parity_math_init(struct m0_parity_math *math,
 				    uint32_t data_count, uint32_t parity_count)
 {
 	int ret;
+#ifndef __KERNEL__
+	uint32_t total_count;
+#endif /* __KERNEL__ */
```
huanghua78, line 358: We can define this variable in following section where it is used.

Usually we define a variable at the beginning of a function. But for simplicity, I think it is OK to do in the other way.

sns/parity_math.c
```diff
@@ -500,6 +651,35 @@ static void reed_solomon_encode(struct m0_parity_math *math,
...
+{
+	uint32_t   pi; /* parity unit index. */
+	uint32_t   di; /* data unit index. */
+	uint32_t   block_size = data[0].b_nob;
+
+	math->data_frags[0] = (uint8_t *)data[0].b_addr;
+	for (di = 1; di < math->pmi_data_count; ++di) {
+		M0_ASSERT(block_size == data[di].b_nob);
+		math->data_frags[di] = (uint8_t *)data[di].b_addr;
+	}
```
huanghua78, line 669: Any reason to handle the "math->data_frags[0]" with exception?
Why not in the "for" loop and the "di" from 0?

sns/parity_math.c
```diff
@@ -670,6 +854,168 @@ static void xor_recover(struct m0_parity_math *math,
...
+	m0_console_printf("\n err_list: ");
+#endif
+	for (i = 0, j = 0, r = 0; i < unit_count; i++) {
+		if (fail[i]) {
+			err_list[j] = i;
+
+			if (i < math->pmi_data_count)
+				data_out[j] = math->data_frags[i];
+			else
+				data_out[j] = math->parity_frags[i - math->pmi_data_count];
```
huanghua78, line 907: Here, the "math->data_frags[i]" and "math->parity_frags[xxx]" are not initialized to point to any buffer.


They are assigned in isa_encode(). But isa_encode and isa_recover() are independent from each other.

sanjognaik: Done.

sns/parity_math.c
sanjognaik: In loop we are comparing block_size, which is data[0].b_nob, with b_nob of every data. If we start with di = 0, we are again comparing block_size with data[0].b_nob. Hence we started with di = 1.

sns/parity_math.c
sanjognaik: Done.

sanjognaik:
We have gathered performance data for encoding and decoding using Intel ISA APIs. This PR was just to get comments and get POC code reviewed.

PR #402, opened 2021-02-03
go/mio: fix getOptimalUnitSz to take pool into account

andriytk:
Currently, getOptimalUnitSz does not take pool into account
when calculating the optimal unit size - which is wrong. It
was missed somehow by accident.

Solution: fix getOptimalUnitSz to involve the pool's EC
configuration in calculations.


andriytk:
cc @chumakd

siningwuseagate (approved):

huanghua78 (approved):

PR #407, opened 2021-02-04
Dtm0 eos 16179

mehjoshi:
checkin for eos-16179 including code changes and DLD.


cas/cas.h
```diff
@@ -293,7 +294,9 @@ struct m0_cas_op {
 	 *
 	 * Array should be non-empty.
 	 */
-	struct m0_cas_recv cg_rec;
+	struct m0_cas_recv  cg_rec;
+	/* dtm0 txr descriptor */
+	struct m0_dtm0_tx_desc *cg_txd;
```
ivan-alekhin, line 299: We cannot put bare pointers into ~XDR~ XC structures. It will be treated as uint64_t. It may work fine within a single address space but won't work between different processes. It would be better to introduce `m0_dtm0_tx_desc_is_empty` and `m0_dtm0_tx_desc_init_empty` if we want this field to be optional.

See this code for details: https://github.com/Seagate/cortx-motr/blob/3323e0a8fabbe632de04947078de0c9436439205/xcode/m0gccxml2xcode#L309

dtm0/fop.c
```diff
@@ -122,6 +124,32 @@ int m0_dtm0_fop_init(void)
...
+{
+	struct dtm0_req_fop  *op;
+	struct m0_fop	     *fop;
+	int		      rc = 0;
+
+	*out = NULL;
+
+	fop = m0_fop_alloc_at(session, &dtm0_req_fop_fopt);
+	if (fop == NULL)
+		rc = -ENOMEM;
```
ivan-alekhin, line 144: Probably, it has to be `return M0_ERR(-ENOMEM);`

ivan-alekhin (commented):
I posted a couple of suggestions.

cas/cas.h
mehjoshi: Done.

dtm0/fop.c
mehjoshi: Done

dtm0/tx_desc.c
```diff
@@ -27,13 +27,33 @@
...
 
+M0_INTERNAL void m0_dtm0_tx_desc_init_none(struct m0_dtm0_tx_desc *td)
+{
+	td->dtd_id.dti_fid = M0_FID0;
+	td->dtd_id.dti_ts = M0_DTM0_TS_INIT;
+	td->dtd_pg.dtpg_nr = 0;
+	td->dtd_pg.dtpg_pa = NULL;
+}
+
+M0_INTERNAL bool m0_dtm0_tx_desc_is_none(const struct m0_dtm0_tx_desc *td)
```
ivan-alekhin, line 46: Let's use the simplest possible model: zeroed tx_desc == empty tx_desc:

```c
void init_none(td) {
  M0_SET0(td);
}
bool is_none(td) {
  return M0_IS0(td);
}
```

cas/service.c
```diff
@@ -1326,6 +1330,24 @@ static int cas_fom_tick(struct m0_fom *fom0)
 			m0_ctg_op_init(&fom->cf_ctg_op, fom0,
 				       cas_op(fom0)->cg_flags);
 		}
+
+		/* If cas request has a txr payload, this a dtm0 operation.
+		 * We need to calculate credits for creating a dtm0 log record.
+		 *
+		 */
+		if (!m0_dtm0_tx_desc_is_none(&cas_op(fom0)->cg_txd)) {
+			struct m0_be_tx_credit dtm0logrec_cred;
```
ivan-alekhin, line 1339: I wonder if credit calculation would work here - the structure was not initialized and it may have some random bytes in it. Was this code path covered in a UT?

mehjoshi: Thanks for catching this. This part of the code is in cas_fom_tick() and it is not covered in our UT today. 
I will fix this.

dtm0/tx_desc.c
mehjoshi: Done.

huanghua78:
Hello @mehjoshi , is this PR still needed? Can it be closed?

huanghua78:
This work is closed. https://jts.seagate.com/browse/EOS-16179

PR #411, opened 2021-02-05
layout: add Access Plan API

andriytk:
Designed by Nikita in 2016, but never implemented. Maybe it's a good time to revive the idea as it might be useful for the In-Storage Compute (ISC, aka Function Shipping).


layout/plan.h
```diff
@@ -0,0 +1,428 @@
...
+m0_composite_layout_at(const struct m0_composite_layout *layout, uint32_t idx);
+void m0_composite_layout_add(struct m0_composite_layout *layout, uint32_t idx);
+void m0_composite_layout_del(struct m0_composite_layout *layout, uint32_t idx);
+
+void m0_composite_layout_write(struct m0_composite_layout *layout,
+			       struct m0_dtx *dtx,
+			       struct m0_op **op);
+
+void m0_composite_layer_index(struct m0_composite_layout_layer *layer,
+			      bool write_mask, struct m0_idx *idx);
```
andriytk, line 51: @nikitadanilov what are these structs and functions in the beginning and how they are related to the rest of the API?

siningwuseagate: @andriytk  here is something I remember when implementing composite layout (for Sage1 project): a composite object operation usually needs to issue a few ops,  such as querying extents from index and then sending IO ops, currently the composite object code does these ops in sequential steps, meaning we have to wait for (or block on) one step until we can move to next steps. Another is for the error handling in composite layout, when one op fails, how to deal with the whole IO.  Back then when implementing composite layout, layout plan was one of the option, but implementing it means the whole clovis IO code had to be re-written. So Sage1 didn't do it.

If we're going to implement the layout plan and use it for isc only, then motr may have 2 set of code doing similar thing (motr client IO code and layout plan), I am not sure if it is a good idea. But if motr is going to refactor client side code to use layout plan, that is going to be a huge job.

johnbent:
Plus @TechWriter-Mayur who I believe has already added this same plan.h somewhere else in cortx-motr repo.  Mayur, I trust Andriy to know the best logical place to add this file.  Can you please watch this PR and after it gets merged, can you remove your plan.h and change the links that you made to point to this new one?

layout/plan.h
```diff
@@ -0,0 +1,428 @@
...
+ */
+
+#pragma once
+
+#ifndef __MOTR_LAYOUT_PLAN_H__
+#define __MOTR_LAYOUT_PLAN_H__
+
+struct m0_layout {
+	struct m0_entity la_ent;
+};
```
huanghua78, line 31: m0_layout is already defined in "layout/layout.h", and please note, the two definitions are different.

Can we just have a forward declaration here?

huanghua78: I know this file has not been included in the Makefile. But removing these obvious bug seems necessary.

huanghua78 (approved):

layout/plan.h
andriytk: @huanghua78, I asked @nikitadanilov above about these structs, let's hear from him. Of course, we shouldn't land it as it is.

layout/plan.h
```diff
@@ -0,0 +1,461 @@
...
+ * is called and returns M0_LAT_READ plop. In this plop user gets cob fid
+ * (at m0_layout_plop::pl_ent) and m0_layout_io_plop::iop_ext describing where
+ * the unit to be read is located at the cob. Now, having this information the
+ * user is ready to send the fop the ioservice.
+ *
+ * After data is read, the user puts it at m0_layout_io_plop::iop_data and
+ * calls m0_layout_plop_done(). One iteration of the graph traversing loop is
+ * done now. If the object consists only of a single unit, the next call to
+ * m0_layout_plan_get() would return M0_LAT_DONE indicating that the i/o m0_op
+ * is done and the plan is finished. But most of the times objects consist
```
andriytk, line 145: As appeared, there will be always `M0_LAT_OUT_READ` after `M0_LAT_READ` to make the data available for user.

layout/plan.h
andriytk: According to @nikitadanilov these two structures and functions were added by mistake to this file, and they just can be removed.

layout/plan.h
```diff
@@ -0,0 +1,434 @@
...
+ * As an optimisation, implementation deterministically colours plops, in such a
+ * way that a user can improve locality of reference by assigning colours to
+ * processor cores. Colouring is optional and does not affect correctness.
+ *
+ * Let's consider, for example, the i/o read operation. The user calls
+ * m0_layout_plan_build() passing m0_op as argument describing the operation.
+ * After this we can enter the "loop" traversing the graph. m0_layout_plan_get()
+ * is called and returns M0_LAT_READ plop. In this plop user gets cob fid
+ * (at m0_layout_plop::pl_ent) and m0_layout_io_plop::iop_ext describing where
+ * the unit to be read is located at the cob. Now, having this information the
```
andriytk, line 111: @nikitadanilov how Function Shipping code (user) can know whether we are going to read the real data or the parity data units? FS needs to know where the data units are located in order to send the computation requests there.

andriytk: @nikitadanilov, what if we introduce an additional plop type - `M0_LAT_MUST_READ` which would indicate that the user must try to read the actual data from the server (data or parity unit) and put it in `iop_data` when calling plop_done()? The simple `M0_LAT_READ` would indicate that the plan is working in normal mode (not the degraded one), so the ISC code could just send the computing request to the server without reading the actual data.

Looks like it would solve the problems for ISC mentioned above.

nikitadanilov: This only works for a particular combination of layout type (parity de-clustered layout type) and particular user (ISC). I don't see how he distinction between `READ` and `MUST_READ` makes sense for an arbitrary user of an arbitrary layout type.

I think the notion of layout plan as a graph (represented by dependencies) is generic enough to capture ISC and other users including normal IO. Let's discuss this on a call. I'll leave two dot-graphs for a normal read and a read with failure as examples.

```digraph plop {
	subgraph cluster_0 {
		style=filled;
		color=lightgrey;
		start;
		getlayout [label="layout-index.GET(fid)"];
		getlayout -> start;
	}
	subgraph cluster_1 {
		style=filled;
		color=lightgrey;
		read0 [label="cob0.read(ext0)"];
		read1 [label="cob1.read(ext1)"];
		read2 [label="cob2.read(ext2)"];
		readout0 [label="buf0.ready"];
		readout1 [label="buf1.ready"];
		readout2 [label="buf2.ready"];
		readout0 -> read0;
		readout1 -> read1;
		readout2 -> read2;
		read0 -> getlayout [style=invis];
		read1 -> getlayout [style=invis];
		read2 -> getlayout [style=invis];
	}
	subgraph cluster_2 {
		style=filled;
		color=lightgrey;
		done
		done -> readout0 [style=invis];
		done -> readout1 [style=invis];
		done -> readout2 [style=invis];
	}
}
```

```
digraph plopfail {
	subgraph cluster_0 {
		style=filled;
		color=lightgrey;
		start;
		getlayout [label="layout-index.GET(fid)"];
		getlayout -> start;
	}
	subgraph cluster_1 {
		style=filled;
		color=lightgrey;
		read0 [label="cob0.read(ext0)" color="red"];
		read1 [label="cob1.read(ext1)"];
		read2 [label="cob2.read(ext2)"];
		readout0 [label="buf0.ready" color="blue"];
		readout1 [label="buf1.ready"];
		readout2 [label="buf2.ready"];
		readout0 -> read0;
		readout1 -> read1;
		readout2 -> read2;
		read0 -> getlayout [style=invis];
		read1 -> getlayout [style=invis];
		read2 -> getlayout [style=invis];
	}
	subgraph cluster_degraded {
		style=filled;
		color=lightgrey;
		readparity [label="cob-parity.read(ext-parity)"];
		readparity -> read0 [style=invis];
		compute [label="erasure(buf1, buf2, buf-parity)"];
		readout0p [label="buf0.ready"];
		compute -> readparity;
		compute -> read1;
		compute -> read2;
		readout0p -> compute;
	}
	subgraph cluster_2 {
		style=filled;
		color=lightgrey;
		done
		done -> readout0p [style=invis];
		done -> readout1 [style=invis];
		done -> readout2 [style=invis];
	}
}
```

andriytk: > I don't see how he distinction between READ and MUST_READ makes sense for an arbitrary user of an arbitrary layout type.

@nikitadanilov, could you please elaborate this statement a bit? For this distinction looks generic enough for ISC needs when a clear indication from the implementation is needed whether the request for the computation can be sent to the server side to be done closer to where the data is stored or not.

andriytk: After discussion with @nikitadanilov there seem to be two major issues with the proposed approach: 1) it won't allow using additional layouts for ISC (like encryption or checksums); 2) it introduces double-reading of the parity group units in the normal i/o path in case of an incidental unforeseeable degraded read (when some READ returns error). Therefore, here is an improved proposition which tackles the above issues and still allows to keep the interface simple for the users:

1. Return READ (instead of MUST_READ) even if there are some FUNctions that must be executed on the read unit data (like checksumming or decompression), but all these functions don't require any other READs in order to get the resulting data for OUT_READ. (Straight line dependency in the graph.) This would tell ISC that the processing of this unit can be done at the server side.

2. Introduce m0_layout_plop_ops::po_update(plop) callback which is called when READ is converted to MUST_READ. This would allow the i/o user to avoid the double-reads on an incidental fallback to degraded read of a parity group. (Note: it is assumed that the OUT_READ plop is returned with parity group granularity and the READ plop is not destroyed until its resulting OUT_READ is ready. This guarantees that po_update(plop) can always be called when needed.)

andriytk: See the proposition in detail at commit https://github.com/Seagate/cortx-motr/pull/411/commits/9fc08d67b01e12f01d6ceb937e3a9753b3286626.

nikitadanilov: As I mentioned during the discussion, it is not clear how the proposal handles the difference between compression (that can be handled by ISC) and client-only encryption (that cannot be handled by ISC).

The only way to define `MUST_READ` precisely and generically, is to define it as `CANNOT_DELEGATE`.  I object to the idea that ISC-specific analysis (that results in the distinction between `READ` and `MUST_READ` plops) should be done by the plan implementation (that is, should be implemented separately by each layout type). This ISC-specific analysis should be done by ISC.

Also, not enough details are provided for the proposed ->po_update() call-back:

    - how the user is supposed to know which plop attributes changed?

    - "The implementation invokes this for a still not complete plop..."---what if the user already completed the `READ` plop during normal IO and already freed the buffer? More generally, what are the ownership and liveness assumptions for plops and buffers in this proposal?

layout/plan.h
```diff
@@ -0,0 +1,619 @@
...
+ *
+ * In case of a write to an object with parity de-clustered layout, the plan
+ * contains additional plops corresponding to parity calculations. In case of
+ * partial group write, the plan, in addition, contains read-plops describing
+ * the logic of read-modify-write cycle. Degraded mode introduces more plops and
+ * so on.
+ *
+ * Access plans are an intermediate abstraction designed for two main users:
+ *
+ *   - IO code in client API, including key-value indices and data objects, and
```
huanghua78, line 58: Do you mean, when the "Access plan" APIs are ready, we need some kind of re-write of client code?

layout/plan.h
```diff
@@ -0,0 +1,619 @@
...
+	 * m0_layout_in_object_plop::iob_ready to the set of available data
+	 * extents. More input plops are generated by the implementation until
+	 * the entire range of data is available.
+	 *
+	 * This plop is useful in situations when the entire input data-set is
+	 * not available at once.
+	 *
+	 * @see m0_layout_inout_plop
+	 */
+	M0_LAT_IN_WRITE,
```
huanghua78, line 305: s/M0_LAT_IN_WRITE/M0_LPT_IN_WRITE
?

similar to M0_LPS_INIT.

layout/plan.h
```diff
@@ -0,0 +1,619 @@
...
+	 * The implementation assigns colours to plops to improve locality of
+	 * reference. The caller can optionally map colours to processor cores
+	 * via m0_locality_get() and execute plops on cores corresponding to
+	 * their colours.
+	 */
+	uint64_t                         pl_colour;
+	/**
+	 * Linkage in the list of all plops in the plan.
+	 */
+	struct m0_tlink                  pl_all_link;
```
huanghua78, line 412: Please clarify the link list head which the link is linked to.

layout/plan.h
huanghua78: As told in the introduction PDF, client i/o code may be re-written to use these APIs.

andriytk: Yes, it can be rewritten later when the plan API implementation is ready and stable.

layout/plan.h
andriytk: Fine with me, but it's the original naming from @nikitadanilov. We can update if it's fine with him.

layout/plan.h
andriytk: It's probably in the `m0_layout_plan` structure which will be introduced later during the implementation. Probably in a `plan_internal.h` file. I will update it then.

huanghua78:
@nikitadanilov , can you please review this PR?

huanghua78:
@nikitadanilov , can you please review this PR?

nikitadanilov (approved):

PR #412, opened 2021-02-08
EOS-16866: [DTM0] DIX with DTX.

ivan-alekhin:
The patch introduces DTX into DIX and
the related modules (motr client, dix/cas
requests and so on).
The patch has an update version of the UT
where m0 client sends a kv pair. When
DTM0 is enabled (--enable-dtm0), DTX module
sets the TX record. The transaction id of this
record is printed on the receiver side.

### DIFF
--- a/dtm0/dtx.c
+++ b/dtm0/dtx.c
@@ -1,0 +1,287 @@
+/* -*- C -*- */
+/*
+ * Copyright (c) 2013-2020 Seagate Technology LLC and/or its Affiliates
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * For any questions about this software or licensing,
+ * please email opensource@seagate.com or cortx-questions@seagate.com.
+ *
+ */
+
+
+
+/**
+ * @addtogroup dtm
+ *
+ * @{
+ */
+
+#define M0_TRACE_SUBSYSTEM M0_TRACE_SUBSYS_DTM
+#include "dtm0/dtx.h"
+#include "lib/assert.h" /* M0_PRE */
+#include "lib/memory.h" /* M0_ALLOC */
+#include "lib/errno.h"  /* ENOMEM */
+#include "lib/trace.h"  /* M0_ERR */
+#include "dtm0/service.h" /* m0_dtm0_service */
+#include "reqh/reqh.h" /* reqh2confc */
+#include "conf/helpers.h" /* proc2srv */
+
+static struct m0_sm_state_descr dtx_states[] = {
+	[M0_DDS_INIT] = {
+		.sd_flags     = M0_SDF_INITIAL | M0_SDF_FINAL,
+		.sd_name      = "init",
+		.sd_allowed   = M0_BITS(M0_DDS_INPROGRESS),
+	},
+	[M0_DDS_INPROGRESS] = {
+		.sd_name      = "inprogress",
+		.sd_allowed   = M0_BITS(M0_DDS_EXECUTED, M0_DDS_FAILED),
+	},
+	[M0_DDS_EXECUTED] = {
+		.sd_name      = "executed",
+		.sd_allowed   = M0_BITS(M0_DDS_PERSISTENT),
+	},
+	[M0_DDS_PERSISTENT] = {
+		.sd_name      = "persistent",
+		.sd_allowed   = M0_BITS(M0_DDS_STABLE),
+	},
+	[M0_DDS_STABLE] = {
+		.sd_name      = "stable",
+		.sd_allowed   = M0_BITS(M0_DDS_DONE),
+	},
+	[M0_DDS_DONE] = {
+		.sd_name      = "done",
+		.sd_flags     = M0_SDF_TERMINAL,
+	},
+	[M0_DDS_FAILED] = {
+		.sd_name      = "failed",
+		.sd_flags     = M0_SDF_TERMINAL | M0_SDF_FAILURE
+	}
+};
+
+static struct m0_sm_trans_descr dtx_trans[] = {
+	{ "populated",  M0_DDS_INIT,       M0_DDS_INPROGRESS },
+	{ "executed",   M0_DDS_INPROGRESS, M0_DDS_EXECUTED   },
+	{ "exec-fail",  M0_DDS_INPROGRESS, M0_DDS_FAILED     },
+	{ "persistent", M0_DDS_EXECUTED,   M0_DDS_PERSISTENT },
+	{ "stable",     M0_DDS_PERSISTENT, M0_DDS_STABLE     },
+	{ "prune",      M0_DDS_STABLE,     M0_DDS_DONE       }
+};
+
+static struct m0_sm_conf dtx_sm_conf = {
+	.scf_name      = "dtm0dtx",
+	.scf_nr_states = ARRAY_SIZE(dtx_states),
+	.scf_state     = dtx_states,
+	.scf_trans_nr  = ARRAY_SIZE(dtx_trans),
+	.scf_trans     = dtx_trans,
+};
+
+M0_INTERNAL void m0_dtm0_dtx_domain_init(void)
+{
+	if (!m0_sm_conf_is_initialized(&dtx_sm_conf))
+		m0_sm_conf_init(&dtx_sm_conf);
+}
+
+M0_INTERNAL void m0_dtm0_dtx_domain_fini(void)
+{
+	if (m0_sm_conf_is_initialized(&dtx_sm_conf))
+		m0_sm_conf_fini(&dtx_sm_conf);
+}
+
+static struct m0_dtm0_dtx *m0_dtm0_dtx_alloc(struct m0_dtm0_service *svc,
+					     struct m0_sm_group     *group)
+{
+	struct m0_dtm0_dtx *dtx;
+
+	M0_PRE(svc != NULL);
+	M0_PRE(group != NULL);
+
+	M0_ALLOC_PTR(dtx);
+	if (dtx != NULL) {
+		dtx->dd_dtms = svc;
+		dtx->dd_ancient_dtx.tx_dtx = dtx;
+		m0_sm_init(&dtx->dd_sm, &dtx_sm_conf, M0_DDS_INIT, group);
+	}
+	return dtx;
+}
+
+static void m0_dtm0_dtx_free(struct m0_dtm0_dtx *dtx)
+{
+	M0_PRE(dtx != NULL);
+	m0_sm_fini(&dtx->dd_sm);
+	m0_dtm0_tx_desc_fini(&dtx->dd_txd);
+	m0_free(dtx);
+}
+
+static int m0_dtm0_dtx_prepare(struct m0_dtm0_dtx *dtx)
+{
+	int               rc;
+
+	M0_PRE(dtx != NULL);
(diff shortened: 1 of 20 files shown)


dtm0/dtx.c
```diff
@@ -0,0 +1,276 @@
...
+
+	M0_LOG(DEBUG, "pa: " FID_F " (User) => " FID_F " (DTM) ",
+	       FID_P(pa_fid), FID_P(&rdtms_fid));
+}
+
+static int m0_dtm0_dtx_close(struct m0_dtm0_dtx *dtx)
+{
+	M0_PRE(dtx);
+
+	m0_sm_state_set(&dtx->dd_sm, M0_DDS_INPROGRESS);
```
just-now, line 189: here and elsewhere could you define a M0_PRE(m0_mutex_is_locked(ƒ))  where ƒ-lock has to be defined.

dtm0/dtx.c
```diff
@@ -0,0 +1,276 @@
...
+}
+
+M0_INTERNAL int m0_dtx_dtm0_init(struct m0_dtm0_service *svc,
+				 struct m0_sm_group     *group,
+				 struct m0_dtx         **out)
+{
+	struct m0_dtm0_dtx *dtx;
+	int                 rc;
+
+	if (svc == NULL) {
```
just-now, line 203: shouldn't it be M0_PRE(svc != NULL)?

dtm0/dtx.c
```diff
@@ -0,0 +1,276 @@
...
+	if (rc != 0)
+		return rc;
+
+	*out = &dtx->dd_ancient_dtx;
+	return 0;
+}
+
+M0_INTERNAL void m0_dtx_dtm0_fini(struct m0_dtx **pdtx)
+{
+	if (*pdtx == NULL) {
```
just-now, line 218: remove {} in single line conditionals.

dtm0/dtx.c
```diff
@@ -0,0 +1,276 @@
...
+{
+	return dtx == NULL ? 0 : m0_dtm0_dtx_prepare(dtx->tx_dtx);
+}
+
+M0_INTERNAL int m0_dtx_dtm0_open(struct m0_dtx  *dtx, uint32_t nr)
+{
+	return dtx == NULL ? 0 : m0_dtm0_dtx_open(dtx->tx_dtx, nr);
+}
+
+M0_INTERNAL void m0_dtx_dtm0_assign(struct m0_dtx       *dtx,
```
just-now, line 236: I feel that this code doesn't compile.
function returns void and int at the same time return dtx == NULL ? 0 .

In 99% of cases it's better to review something which at least compiles.

dtm0/dtx.c
```diff
@@ -0,0 +1,276 @@
...
+
+M0_INTERNAL int m0_dtx_dtm0_close(struct m0_dtx *dtx)
+{
+	return dtx == NULL ? 0 : m0_dtm0_dtx_close(dtx->tx_dtx);
+}
+
+M0_INTERNAL int m0_dtx_dtm0_copy_txd(struct m0_dtx *dtx,
+				     struct m0_dtm0_tx_desc *txd)
+{
+	if (dtx == NULL) {
```
just-now, line 252: why do you think dtx can be == NULL in many cases here? if dtx == NULL infrequently (1% of cases or lesser), let's just put M0_PRE(dtx != NULL).
Here and elsewhere.

dtm0/dtx.c
```diff
@@ -0,0 +1,276 @@
...
+	return dtx == NULL ? 0 : m0_dtm0_dtx_assign(dtx->tx_dtx, pa_idx,
+						    pa_fid);
+}
+
+M0_INTERNAL int m0_dtx_dtm0_close(struct m0_dtx *dtx)
+{
+	return dtx == NULL ? 0 : m0_dtm0_dtx_close(dtx->tx_dtx);
+}
+
+M0_INTERNAL int m0_dtx_dtm0_copy_txd(struct m0_dtx *dtx,
```
just-now, line 249: please use consts to indicate srcs and dsts.

dtm0/dtx.h
```diff
@@ -0,0 +1,99 @@
...
+
+#include "dtm0/tx_desc.h" /* m0_dtm0_tx_desc */
+#include "sm/sm.h"        /* m0_sm */
+#include "dtm/dtm.h"      /* m0_dtx */
+
+enum m0_dtm0_dtx_state {
+	/** An empty dtx. */
+	M0_DDS_INIT,
+	/* dtx has a valid tx record. */
+	M0_DDS_INPROGRESS,
```
just-now, line 36: I'm not against this name in an arbitrary sense. Still i _feel_ it doesn't match with EXECUTED, STABLE, DONE, etc. Can we name it like "launched" or "started" or "sent"?

just-now (commented):
looks not a bad patch. 
has all logic there. 
please put some attention into locking of dtx sm (i feel it can be performed from differnt contexts).

dtm0/dtx.c
```diff
@@ -0,0 +1,276 @@
...
+#include "conf/helpers.h" /* proc2srv */
+
+static struct m0_sm_state_descr dtx_states[] = {
+	[M0_DDS_INIT] = {
+		.sd_flags     = M0_SDF_INITIAL | M0_SDF_FINAL,
+		.sd_name      = "init",
+		.sd_allowed   = M0_BITS(M0_DDS_INPROGRESS),
+	},
+	[M0_DDS_INPROGRESS] = {
+		.sd_name      = "dtx-executed",
```
mehjoshi, line 47: .sd_name = "dtx-inprogress"

dtm0/dtx.c
```diff
@@ -0,0 +1,276 @@
...
+		.sd_allowed   = M0_BITS(M0_DDS_EXECUTED, M0_DDS_FAILED),
+	},
+	[M0_DDS_EXECUTED] = {
+		.sd_name      = "dtx-persistent",
```
mehjoshi, line 51: .sd_name = "dtx-executed"

dtm0/service.c
```diff
@@ -306,6 +303,21 @@ M0_INTERNAL bool m0_dtm0_is_a_persistent_dtm(struct m0_reqh_service *service)
 		to_dtm(service)->dos_origin == DTM0_ON_PERSISTENT;
 }
 
+M0_INTERNAL int m0_dtm0_service_find(struct m0_reqh *reqh,
+				     struct m0_dtm0_service **out)
+{
+	struct m0_reqh_service *rh_srv;
+
+	rh_srv = m0_reqh_service_find(&dtm0_service_type, reqh);
+	if (rh_srv == NULL)
```
mehjoshi, line 312: m0_reqh_service_find() signature in service.h returns an int. Should this be: reqh == NULL ? or rather in keeping with our coding style: if(!reqh)

motr/ut/idx_dix.c
```diff
@@ -786,7 +786,7 @@ struct m0_client* st_get_instance()
 #include "dtm0/service.h"
 
 #define M0_FID(c_, k_)  { .f_container = c_, .f_key = k_ }
-static void st_one_dtm0_op_idx_create(void)
```
mehjoshi, line 789: I don't see st_one_dtm0_op_idx_create being used anywhere. If the only reason for making this non-static is to overcome compiler error then consider marking it M0_UNUSED instead.

dtm0/service.c
```diff
@@ -306,6 +303,21 @@ M0_INTERNAL bool m0_dtm0_is_a_persistent_dtm(struct m0_reqh_service *service)
 		to_dtm(service)->dos_origin == DTM0_ON_PERSISTENT;
 }
 
+M0_INTERNAL int m0_dtm0_service_find(struct m0_reqh *reqh,
```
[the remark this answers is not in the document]
ivan-alekhin: thanks, returning a pointer and adding `const` qualifier makes sense here.

to_dtm() will not take care about nullptr. It will panic because bob_of will panic. It has to be a bit more sophisticated than you have described.

motr/idx.c
```diff
@@ -150,6 +154,25 @@ static int idx_op_init(struct m0_idx *idx, int opcode,
 	m0_op_idx_bob_init(oi);
 	m0_ast_rc_bob_init(&oi->oi_ar);
 
+	if (M0_IN(op->op_code, (M0_IC_PUT, M0_IC_DEL))) {
+		/* If DTM0 service start after m0c was initialized,
+		 * we still can get this value. Such a scenario may
+		 * happen in UTs.
+		 */
+		if (m0c->m0c_dtms == NULL) {
```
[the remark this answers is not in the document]
ivan-alekhin: I'm going to remove that completely. If dtms is not found then it is either a UT (which means UT code is responsible for that) or misconfigured system.

dtm0/dtx.h
```diff
@@ -0,0 +1,99 @@
...
+	struct m0_dtm0_tx_desc  dd_txd;
+	struct m0_dtm0_service *dd_dtms;
+};
+
+/* This portion of the API goes against the naming convention because
+ * it operates on m0_dtx (from its user's standpoint) but internally
+ * it uses m0_dtm0_dtx-related code.
+ */
+
+M0_INTERNAL int m0_dtx_dtm0_init(struct m0_dtm0_service *svc,
```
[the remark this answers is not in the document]
ivan-alekhin: thanks, that is good idea to use m0_dtx0 prefix for the provided wrappers around m0_dtx.

dtm0/dtx.c
```diff
@@ -0,0 +1,276 @@
...
+		return rc;
+
+	dtx->dd_txd.dtd_id = (struct m0_dtm0_tid) {
+		.dti_ts  = now,
+		.dti_fid = dtx->dd_dtms->dos_generic.rs_service_fid,
+	};
+
+	M0_POST(m0_dtm0_tid__invariant(&dtx->dd_txd.dtd_id));
+	return rc;
+}
```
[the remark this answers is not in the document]
ivan-alekhin: it can be simplified a bit but it cannot be as simple as you described. In your example, the post condition is checked before the timestamp component of the TID is filled. This post condition must ensure that the invariant is held before exiting the scope.

dtm0/dtx.c
ivan-alekhin: good catch, thanks!

dtm0/dtx.c
```diff
@@ -0,0 +1,276 @@
...
+
+static int m0_dtm0_dtx_open(struct m0_dtm0_dtx  *dtx,
+			    uint32_t             nr)
```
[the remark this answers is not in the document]
ivan-alekhin: Github UI for diffs has bugs. This an example of such a bug.

dtm0/dtx.c
ivan-alekhin: it is much easier to have a single if condition inside this function rather than in 4 functions inside CAS client.

dtm0/dtx.h
ivan-alekhin: If we want to describe it precisely then the name of the state is "request-is-known-but-no-replies-yet". InProgress is a generic alias for this name. We could use Launched/Started/Sent but it has to be a designated state. In this patch, InProgress does not mean something was sent or logged, it just means the transaction has a valid txr (as it is stated in the comment), and it has no idea about replies. If there are any changes to that in the consequent patches then I'll definitely rename it.

dtm0/service.c
```diff
@@ -208,17 +201,21 @@ static int dtm0_service__alloc(struct m0_reqh_service **service,
...
 	if (s == NULL)
-		return -ENOMEM;
+		return M0_ERR(-ENOMEM);
 
 	s->dos_generic.rs_type = stype;
 	s->dos_generic.rs_ops  = ops;
+	rc = dtm0_service__init(s);
+	if (rc != 0)
+		return rc;
+
```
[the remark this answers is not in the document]
ivan-alekhin: I guess it is not possible. How do we assign the output variable in this case? the rc must be checked and then the variable needs to be assigned.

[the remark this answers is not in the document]
just-now: ...
	s->dos_generic.rs_type = stype;
	s->dos_generic.rs_ops  = ops;		s->dos_generic.rs_ops  = ops;
        *service = &s->dos_generic; // no one cares about *service if rc != 0
        return dtm0_service__init(s);

just-now:
Can be merged.
Some logic fixes related such as locking and others will be addressed later.

PR #414, opened 2021-02-09
EOS-16179:[DTM0]

mehjoshi:
Added dtm0 fop request, reply structures, operation and message types,
markers in code where dtm0 operation related credit calculation should
be performed and dtm0 log routines should be called for the basic case
of executing a dtm0 request.
Also added DLD for dtm0 service.

Signed-off-by: Mehul Manharlal Joshi <mehul.joshi@seagate.com>

### DIFF
--- a/cas/client.c
+++ b/cas/client.c
@@ -176,6 +176,7 @@ static int creq_op_alloc(uint64_t           recs_nr,
 	} else {
 		op->cg_rec.cr_nr = recs_nr;
 		op->cg_rec.cr_rec = rec;
+		m0_dtm0_tx_desc_init_none(&op->cg_txd);
 		*out = op;
 	}
 	return M0_RC(0);
--- a/dtm0/fop.c
+++ b/dtm0/fop.c
@@ -20,6 +20,7 @@
  */
 
 
+#include "be/dtm0_log.h"
 #include "dtm0/tx_desc.h"
 #include "dtm0/tx_desc_xc.h"
 #include "lib/errno.h"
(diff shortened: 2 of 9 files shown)


dtm0/fop.c
```diff
@@ -110,46 +115,91 @@ int m0_dtm0_fop_init(void)
...
+static int dtm0_fom_create(struct m0_fop *fop,
+			  struct m0_fom **out, struct m0_reqh *reqh)
+{
+	struct dtm0_fom    *fom;
+	struct m0_fom     *fom0;
+	struct m0_fop     *repfop;
+	struct dtm0_rep_fop *reply;
+
+	if (!dtm0_service_started(fop, reqh))
+		return M0_ERR(-EAGAIN);
```
just-now, line 182: Do we really need this check?
I feel that we never hit -EAGAIN case here. Also, I think that similar check is somewhere during processing on the fom checking that service instance is proper.

just-now (commented):
Looks good.

dtm0/fop.c
mehjoshi: I think this may not be needed .. will remove it.

just-now (approved):
I'm OK to merge.

cas/client.c
```diff
@@ -176,6 +176,7 @@ static int creq_op_alloc(uint64_t           recs_nr,
 	} else {
 		op->cg_rec.cr_nr = recs_nr;
 		op->cg_rec.cr_rec = rec;
+		m0_dtm0_tx_desc_init_none(&op->cg_txd);
```
just-now, line 179: test, ignore this.
cc @just-now

PR #418, opened 2021-02-11
build: bump rpm version number to 2.0.0

huanghua78:

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,5 +1,5 @@
 # Motr version
-m4_define([M0_VERSION_MAJOR],[1])
+m4_define([M0_VERSION_MAJOR],[2])
 m4_define([M0_VERSION_MINOR],[0])
 m4_define([M0_VERSION_PATCH],[0])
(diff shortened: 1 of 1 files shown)


shashank-parulekar (approved):

configure.ac
```diff
@@ -1,5 +1,5 @@
 # Motr version
-m4_define([M0_VERSION_MAJOR],[1])
+m4_define([M0_VERSION_MAJOR],[2])
```
ujjwalpl, line 2: Can we also accept version number as input from RE and use the same. If those values are not passed, we can use these values as default.

ujjwalpl (changes requested):

configure.ac
huanghua78: OK. This will be done in another PR.

mukundkanekar (approved):

configure.ac
huanghua78: https://github.com/Seagate/cortx-motr/pull/422

PR #424, opened 2021-02-12
Fix m0c compilation without DTM0 enabled.

ivan-alekhin:
Problem: When "--enable-dtm0" is not passed
the client code compilation fails.
Solution: Wrap the unused variabe with the DTM0
feature flag.

The fix addressed the compilation issue caused by landing of https://github.com/Seagate/cortx-motr/pull/412


Anilsindhi (approved):
LGTM +1

PR #426, opened 2021-02-12
Fixed UT issue in DTM0 Log module

Anilsindhi:
DTM0 Log UT is failing randomly while updating already added log, seems like specific API is returning uninitialized return code for that case and hence asserting.


huanghua78:
Hello @Anilsindhi , is this PR still valid and needed?

huanghua78:
This is inactive for sometime. If this is still needed, please re-open it.

PR #431, opened 2021-02-16
provisioner: added minimal json file for motr mini provisioner deploy

bansisoni:
- added motr_cluster.json confstore key-value file for motr single/three node mini provisioner deployment
- this file includes only keys required for motr deployment

Signed-off-by: Bansidhar Soni <bansidhar.soni@seagate.com>


huanghua78 (approved):

yeshpal-jain-seagate (approved):
We can keep this only in wiki to avoid changes required, this is also not required on HW so it can be kept only in wiki

bansisoni:
Closing this as it is not require anymore. I will add this in wiki page.

PR #440, opened 2021-02-17
Added the HLD of Motr Lostore Document

VenkyOS:


huanghua78 (approved):
This doc is out-dated, but let's keep it as it is.

PR #454, opened 2021-02-22
provisioning: fix ansible_user var usage

andriytk:
To avoid problems with ansible_user when running ansible ssh-tasks on different platforms (like Linux and macOS) - use "vagrant" username directly instead. In particular, this fixes the following issue:

```
TASK [ssh-jumphost : generate user SSH key pair] *******************************
fatal: [cmu]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'ansible_user' is
undefined\n\nThe error appears to be in '/vagrant/roles/ssh-jumphost/tasks/main.yml': line 24, column 3, but may\nbe elsewh
ere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n---\n- name: generate user S
SH key pair\n  ^ here\n"}
msg:  The task includes an option with an undefined variable. The error was: 'ansible_user' is undefined
The error appears to be in '/vagrant/roles/ssh-jumphost/tasks/main.yml': line 24, column 3, but may
be elsewhere in the file depending on the exact syntax problem.
```


andriytk:
cc @chumakd

chumakd (approved):

huanghua78:
What if I am not using vagrant?

andriytk:
@huanghua78 I suppose the ssh-related tasks are running only under vagrant environment. @chumakd is it right?

huanghua78 (approved):
Looks fine. Tests passed.

PR #458, opened 2021-02-23
provisioning: add StrictHostKeyChecking=no at cmu

andriytk:
Add StrictHostKeyChecking=no to /etc/ssh/ssh_config at cmu
to allow automatic ssh connections to the nodes without questions
for the host authentication confirmations like this one:
```
  $ ssh ssu1.local
  The authenticity of host 'ssu1.local' can't be established.
  ECDSA key fingerprint is SHA256:...
  ECDSA key fingerprint is MD5:...
  Are you sure you want to continue connecting (yes/no)?
```


andriytk:
cc @chumakd

huanghua78 (approved):

chumakd (approved):

PR #459, opened 2021-02-24
Delete .whitesource

mukul-seagate11:


huanghua78:
Hello, @mukul-seagate11 , this file is not needed any longer?

PR #462, opened 2021-02-24
EOS-18124: ISA-L - Implement encode function

sanjognaik:
- Updated m0_parity_math_init() function to use initialization from
  ISA library.
- Implemented isal_encode() function to encode data.
- Updated m0_parity_math_fini() function to free memory allocated for
  coefficient buffers and tables.
- Added debug messages.


sns/parity_math.c
```diff
@@ -500,29 +551,110 @@ static void reed_solomon_encode(struct m0_parity_math *math,
...
+	if (parity_frags == NULL) {
+		M0_ERR_INFO(-ENOMEM, "failed to allocate memory for array of "
+			    "parity fragments");
+		goto fini;
+	}
+
+	data_frags[0] = (uint8_t *)data[0].b_addr;
+	for (i = 1; i < data_count; ++i) {
+		if (block_size != data[i].b_nob) {
+			M0_ERR_INFO(-errno, "data block size mismatch. "
```
huanghua78, line 595: Please don't use the external variable 'errno'.
At this point, the 'external' may contain irrelevant value.
Just return -EINVAL.

sns/parity_math.c
```diff
@@ -500,29 +551,110 @@ static void reed_solomon_encode(struct m0_parity_math *math,
...
+				    "block_size = %u, data[%u].b_nob=%u",
+				    block_size, i, data[i].b_nob);
+			goto fini;
+		}
+		data_frags[i] = (uint8_t *)data[i].b_addr;
+	}
+
+	for (i = 0; i < parity_count; ++i) {
+		if (block_size != parity[i].b_nob) {
+			M0_ERR_INFO(-errno, "parity block size mismatch. "
```
huanghua78, line 605: The same as above.

sns/parity_math.c
```diff
@@ -500,29 +551,110 @@ static void reed_solomon_encode(struct m0_parity_math *math,
...
+				    block_size, i, parity[i].b_nob);
+			goto fini;
+		}
+		parity_frags[i] = (uint8_t *)parity[i].b_addr;
+	}
+
+	M0_LOG(M0_DEBUG, "initialize tables for fast Erasure Code encode.");
+	ec_init_tables(data_count, parity_count,
+		       &math->encode_matrix[data_count * data_count],
+		       math->g_tbls);
```
huanghua78, line 616: Can this initialization be moved into "m0_parity_math_init()"?
Or this is needed for every ec_encode_data() call?

sanjognaik: We need g_tbls for encoding / decoding i.e. in ec_encode_data(). While decoding, the parameters passed to initialize the tables are different from those used in encoding. So to ensure that we have correct values in tables, it required to initialize it before calling ec_encode_data().

sns/parity_math.c
```diff
@@ -500,29 +551,114 @@ static void reed_solomon_encode(struct m0_parity_math *math,
...
+	M0_LOG(M0_DEBUG, "generate erasure codes on given blocks of data.");
+	ec_encode_data(block_size, data_count, parity_count,
+		       math->g_tbls, data_frags, parity_frags);
+
+fini:
+	m0_free(data_frags);
+	m0_free(parity_frags);
+
+	/* TODO: Return error code instead of assert */
+	M0_ASSERT(ret == 0);
```
huanghua78, line 628: Currently, this function returns void, so we ASSERT the ret code here.
In future, error code should be returned and errors should be handled by caller.

huanghua78 (approved):
Looks fine.

sns/parity_math.c
sanjognaik: Yes. We have created separate ticket [EOS-17288](https://jts.seagate.com/browse/EOS-17288) for the same.

sns/parity_math.c
```diff
@@ -500,29 +555,114 @@ static void reed_solomon_encode(struct m0_parity_math *math,
...
+	M0_PRE(data != NULL);
+	M0_PRE(parity != NULL);
+
+	M0_ENTRY();
+
+	data_count = math->pmi_data_count;
+	parity_count = math->pmi_parity_count;
+	block_size = data[0].b_nob;
+
+	M0_ALLOC_ARR(data_frags, data_count);
```
shashank-parulekar, line 583: Since data_count does not change please move the memory allocation to m0_parity_math_init().

sns/parity_math.c
```diff
@@ -500,29 +555,114 @@ static void reed_solomon_encode(struct m0_parity_math *math,
...
+	if (data_frags == NULL) {
+		ret = M0_ERR_INFO(-ENOMEM, "failed to allocate memory for "
+				  "array of data fragments");
+		goto fini;
+	}
+
+	M0_ALLOC_ARR(parity_frags, parity_count);
```
shashank-parulekar, line 590: Since parity_count does not change please move the memory allocation to m0_parity_math_init().

sns/parity_math.c
shashank-parulekar: @sanjognaik Can we use 2 pointers in that case one to be used for encode and other for decode with the encode/decode routines calling the appropriate pointers. This is because we should avoid to initialize things during runtime if possible.

shashank-parulekar (changes requested):

sns/parity_math.c
sanjognaik: Done.

sns/parity_math.c

sns/parity_math.c
sanjognaik: Created separate pointer 'pmi_encode_tbls' for encoding and initialized it in m0_parity_math_init() function now. Will use separate pointer for decode as well in PR for recover function.

mukundkanekar (approved):

sns/parity_math.c
```diff
@@ -500,29 +562,93 @@ static void reed_solomon_encode(struct m0_parity_math *math,
...
+	M0_PRE(math != NULL);
+	M0_PRE(data != NULL);
+	M0_PRE(parity != NULL);
+	M0_PRE(math->pmi_data_count >= 1);
+	M0_PRE(math->pmi_parity_count >= 1);
+	M0_PRE(math->pmi_data_count >= math->pmi_parity_count);
+	M0_PRE(math->pmi_data_count <= SNS_PARITY_MATH_DATA_BLOCKS_MAX);
+
+	uint8_t  *frags_in[math->pmi_data_count];
+	uint8_t  *frags_out[math->pmi_parity_count];
```
huanghua78, line 585: Variable-sized array object. Let's ask @nikitadanilov to review this code block.

huanghua78, line 585: Variable-sized array object. Let's ask @nikitadanilov to review this code block.

shashank-parulekar (approved):

PR #468, opened 2021-03-01
Quick-Start-Guide.rst: fix Prerequisites section

andriytk:
Drop mentioning the explicit kernel version and CentOS version and mentioning
that different kernel versions are supported at the same time - it's very confusing.
Just mention CentOS-7 version for x86_64 platform and ARM64 platform porting
work in progress.

Fix typo in mentioning EPEL repo and put the exact command about how to install it
instead of pointing to the document which is about to be deleted.

### DIFF
--- a/doc/Quick-Start-Guide.rst
+++ b/doc/Quick-Start-Guide.rst
@@ -3,22 +3,18 @@ Quick Start Guide
 =================
 This guide provides information to get Motr component ready.
 
-************
-Prerequisite
-************
+*************
+Prerequisites
+*************
 The prerequisite that is necessary to install the Motr component is mentioned below.
 
-- Kernel Version - 3.10.0-1062.el7
-
-  - To know the version being used, type the following:
-
-    - **uname -r**
-
-  - Different kernel versions that come from Centos7.7 or RHEL7.7 are supported.
+- CentOS-7 for x86_64 platform (ARM64 platform support is in progress).
 
 - **Ansible** is needed. 
 
-  - Please refer to `Install EPEP repo <https://github.com/Seagate/cortx/blob/main/doc/LocalVMSetup.md#4-you-may-need-to-add-epel-repo>`_.
+  - Install EPEL repo:
+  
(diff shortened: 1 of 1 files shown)


gowthamchinna:
run mini-provisioning

doc/Quick-Start-Guide.rst
```diff
@@ -3,22 +3,18 @@ Quick Start Guide
...
-
-  - Different kernel versions that come from Centos7.7 or RHEL7.7 are supported.
+- CentOS-7 for x86_64 platform (ARM64 platform support work is in progress).
 
 - **Ansible** is needed. 
 
-  - Please refer to `Install EPEP repo <https://github.com/Seagate/cortx/blob/main/doc/LocalVMSetup.md#4-you-may-need-to-add-epel-repo>`_.
+  - Install EPEL repo:
+  
+    **$ sudo yum install https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm**
```
mukul-seagate11, line 17: It should be,

sudo yum install https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm** -y

mukul-seagate11 (changes requested):

doc/Quick-Start-Guide.rst
huanghua78: The "-y " means to install rpm without confirmation from user "Yes/No".
But I think this is optional.

andriytk: Yes, Hua is right. It's not a script we are running here and there is no harm if the user will be asked for the confirmation.

huanghua78:
@mukul-seagate11, do you have more comments?

doc/Quick-Start-Guide.rst
mukul-seagate11: OK

mukul-seagate11 (approved):
looks good

PR #484, opened 2021-03-10
EOS-17630: [DTM0] Handle P notice on the client.

ivan-alekhin:
The patch aimed at PERSISTENT (P) notice mechanism
implementation for the happy-path scenario.
P notice is sent by the server side when the corresponding
CAS operation gets comitted, and it is beign handled
by the log through the ast mechanism.
The patch also incudes several minor improvements
for DTM0 traces and fixes for the configuration
of the Motr client UT to enable proper communication
between the DTM0 services on the client and the server
sides of the UT.

### DIFF
--- a/dtm0/fop.c
+++ b/dtm0/fop.c
@@ -84,7 +84,9 @@ static const struct m0_fom_type_ops dtm0_req_fom_type_ops = {
 
 static void dtm0_rpc_item_reply_cb(struct m0_rpc_item *item)
 {
-	struct m0_fop *reply;
+	struct m0_fop *reply = NULL;
+
+	M0_ENTRY("item=%p", item);
 
         M0_PRE(item != NULL);
 	M0_PRE(M0_IN(m0_fop_opcode(m0_rpc_item_to_fop(item)),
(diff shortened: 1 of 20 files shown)


dtm0/fop.c
```diff
@@ -200,6 +206,76 @@ static size_t dtm0_fom_locality(const struct m0_fom *fom)
...
+	m0_dtm0_post__persistent(dtms, &msg.dtd_id.dti_fid, &msg);
+
+	/* TODO: Send notices to the rest of the participants. */
+	m0_dtm0_tx_desc_fini(&msg);
+}
+
+
+
 #define M0_FID(c_, k_)  { .f_container = c_, .f_key = k_ }
 static void dtm0_pong_back(enum m0_be_dtm0_log_credit_op msgtype,
```
just-now, line 280: m0_dtm0_post__persistent and dtm0_pong_back are similar, what about to reduce copy paste?

dtm0/fop.c
```diff
@@ -224,31 +300,49 @@ static void dtm0_pong_back(enum m0_be_dtm0_log_credit_op msgtype,
...
 
 	if (m0_fom_phase(fom) < M0_FOPH_NR) {
 		rc = m0_fom_tick_generic(fom);
 	} else {
 		req = m0_fop_data(fom->fo_fop);
-
-		if (req->dtr_msg == M0_DTML_EXECUTED && m0_dtm0_is_a_persistent_dtm(fom->fo_service)) {
+		M0_LOG(M0_DEBUG, "Processing non-generic phase %d %d, " FID_F
+		       ", %p",
+		       (int) req->dtr_msg,
```
just-now, line 322: (int) -> !!
here and elsewhere.

just-now (commented):
two major concerns:
1) pna_list
2) code looks quite verbose and bulky, therefore in next patches will be working on style.

just-now (approved):
Can be merged without an additional round of review.

General observation.
We didn't use word 'notice' in the design, we use 'message'. Let's discuss if we need not to use 'notice' after some time.

PR #486, opened 2021-03-10
lib: enhance m0_uint128_sscanf to allow dec/oct formats

andriytk:
Enhance m0_uint128_sscanf() to allow parsing the IDs specified in decimal and octal formats also. (Currently, it understands hex format only.)

### DIFF
--- a/lib/misc.c
+++ b/lib/misc.c
@@ -50,7 +50,7 @@ M0_INTERNAL int m0_uint128_cmp(const struct m0_uint128 *u0,
 
 M0_INTERNAL int m0_uint128_sscanf(const char *s, struct m0_uint128 *u128)
 {
-	int rc = sscanf(s, U128X_F, U128_S(u128));
+	int rc = sscanf(s, U128I_F, U128_S(u128));
 	return rc == 2 ? 0 : -EINVAL;
 }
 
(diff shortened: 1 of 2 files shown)


andriytk:
cc @nikitadanilov

lib/misc.c
```diff
@@ -50,7 +50,7 @@ M0_INTERNAL int m0_uint128_cmp(const struct m0_uint128 *u0,
 
 M0_INTERNAL int m0_uint128_sscanf(const char *s, struct m0_uint128 *u128)
 {
-	int rc = sscanf(s, U128X_F, U128_S(u128));
+	int rc = sscanf(s, U128I_F, U128_S(u128));
```
huanghua78, line 53: This patch needs to be thoroughly tested, because this patch will make some result different for the same input.
For example, originally, "123a:7890" (without the leading "0x") will be recognized as valid input. But now, it is not.
"0x123:456" will be treated as two hex numbers, but now only the first part is treated as hex, and the second part is treated as hex.

andriytk: Yes. Luckily, only `go/mio` uses this code atm. :)

huanghua78: OK.

huanghua78 (approved):

madhavemuri (approved):

PR #416, opened 2021-02-09
Replace Galois Arithmetic APIs with respective ISA APIs

sanjognaik:
- Updated Makefile.am and configure.ac to use Intel ISA library.
- Used Intel ISA arithmetic APIs in m0_parity_mul() and m0_parity_div() functions.
- Intel ISA APIs are accessed in user space only.
- Galois APIs are called only in kernel space for now.
- Moved galois table create and release API calls to kernel space only.
- Added option 'run-ub' to m0 script to run unit benchmarking.
- Updated ub tests for parity_math.
- Increased number of iterations for parity_math ub tests to 100.

PR #489, opened 2021-03-11
EOS-18137 sns/parity_math: implement isal_diff() function

sanjognaik:
- Implemented isal_diff() function to calculate differential parity.
- Added MACRO to check if ISA-L code is enabled or not.
- Use Galois library only in kernel space.

### DIFF
--- a/sns/parity_math.c
+++ b/sns/parity_math.c
@@ -410,7 +410,7 @@ static bool parity_math_invariant(const struct m0_parity_math *math)
 {
 	return  _0C(math != NULL) && _0C(math->pmi_data_count >= 1) &&
 		_0C(math->pmi_parity_count >= 1) &&
-		_0C(math->pmi_data_count >= math->pmi_parity_count);
+		_0C(math->pmi_data_count >= math->pmi_parity_count) &&
 		_0C(math->pmi_data_count <= SNS_PARITY_MATH_DATA_BLOCKS_MAX);
 }
 #endif /* ISAL_ENCODE_ENABLED */
--- a/sns/parity_math.c
+++ b/sns/parity_math.c
@@ -36,7 +36,7 @@ enum {
 	IR_INVALID_COL = UINT8_MAX,
 };
 
-#ifndef __KERNEL__
+#if ISAL_ENCODE_ENABLED
 #define ALLOC_ARR_INFO(arr, nr, msg, ret) ({					\
 	M0_ALLOC_ARR(arr, nr);							\
 	if (arr == NULL) 							\
@@ -45,14 +45,14 @@ enum {
 	else 									\
 		(ret) = M0_RC_INFO(0, "allocate memory for " msg);		\
 })
-#endif
+#endif /* ISAL_ENCODE_ENABLED */
 
 /* Forward declarations */
 static void xor_calculate(struct m0_parity_math *math,
                           const struct m0_buf *data,
                           struct m0_buf *parity);
 
-#ifndef __KERNEL__
+#if ISAL_ENCODE_ENABLED
 static void isal_encode(struct m0_parity_math *math,
                         const struct m0_buf *data,
                         struct m0_buf *parity);
@@ -60,7 +60,7 @@ static void isal_encode(struct m0_parity_math *math,
 static void reed_solomon_encode(struct m0_parity_math *math,
                                 const struct m0_buf *data,
                                 struct m0_buf *parity);
-#endif /* __KERNEL__ */
+#endif /* ISAL_ENCODE_ENABLED */
 
 static void xor_diff(struct m0_parity_math *math,
 		     struct m0_buf         *old,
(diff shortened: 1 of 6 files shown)


sns/parity_math.h
```diff
@@ -43,6 +43,12 @@
    @{
 */
 
+/**
+ *  Define the condition here to use Intel ISA library.
+ */
+#define ISAL_ENCODE_ENABLED	(!defined(__KERNEL__))
```
sanjognaik, line 49: Once [PR-416](https://github.com/Seagate/cortx-motr/pull/416) is merged, this can be updated to 
`!defined(__KERNEL__) && defined(HAVE_ISAL)`

sns/parity_math.c
```diff
@@ -382,12 +401,25 @@ static int vandmat_norm(struct m0_matrix *m)
 
 	return 0;
 }
-#endif /* __KERNEL__ */
+#endif /* RS_ENCODE_ENABLED */
+
+#if ISAL_ENCODE_ENABLED
+static inline void parity_math_validate(struct m0_parity_math *math)
```
huanghua78, line 407: Usually, we can this xxx_invariant().
And _0C() can be used to verify conditions.
So this may look like this:
```
static bool parity_math_validate(struct m0_parity_math *math)
{
        return _0C(math != NULL) && _0C(math->pmi_data_count >= 1) &&
                   _0C(math->pmi_parity_count >= 1) &&
                   _0C(math->pmi_data_count >= math->pmi_parity_count) &&
                   _0C(math->pmi_data_count <= SNS_PARITY_MATH_DATA_BLOCKS_MAX);
}
```

huanghua78 (approved):

sns/parity_math.c

sns/parity_math.c
```diff
@@ -914,15 +1003,15 @@ static int isal_gen_recov_coeff_tbl(uint32_t data_count, uint32_t parity_count,
...
-	if (ret != 0)
+	if (temp_mat == NULL)
 		goto fini;
 
 	ALLOC_ARR_INFO(invert_mat, (unit_count * data_count), "invert matrix", ret);
-	if (ret != 0)
+	if (invert_mat == NULL)
 		goto fini;
 
 	/* Construct temp_mat (matrix that encoded remaining frags)
```
shashank-parulekar, line 1017: In the below for-loop confirm the array index used in encode_mat is correct. Since I do not have the background of this I am having a hard time understanding the code. because it seems you take the byte from alive_idx_buf buffer and multiply this value with data_count. Are you sure the encode_mat array is big enough to hold it.
Similarly the decode matrix is doing the same thing.

shashank-parulekar (changes requested):

sns/parity_math.c
sanjognaik: In our case, data_count = 4 and parity_count = 2.
unit_count = (data_count + parity_count) = 6

The encode_matrix is of size (unit_count * data_count) which is 24 bytes for us i.e. 6 rows and 4 columns
The decode_matrix is also of same size (unit_count * data_count) i.e. 24 bytes.

In below for loop, the values present in alive_idx_buf are 0 < idx < unit_count i.e.  0 < idx < 6
Hence for 'val = (data_count * idx + j)' calculation, the range would be 0 < val < 24.

Same is true for decode_matrix, temp_matrix and inver_matrix.

sns/parity_math.c
```diff
@@ -440,19 +471,19 @@ M0_INTERNAL int m0_parity_math_init(struct m0_parity_math *math,
 		ALLOC_ARR_INFO(math->pmi_encode_matrix,
```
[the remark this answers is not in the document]
sanjognaik: Done

sns/parity_math.c
```diff
@@ -541,6 +572,67 @@ static void xor_calculate(struct m0_parity_math *math,
...
+
+	block_size = new[index].b_nob;
+
+	ALLOC_ARR_INFO(diff_data, block_size, "differential data block", ret);
+	if (diff_data == NULL)
+		goto fini;
+
+	M0_LOG(M0_DEBUG, "calculate differential data.");
+	for (i = 0; i < block_size; i++) {
+		diff_data[i] = ((uint8_t *)old[index].b_addr)[i] ^
```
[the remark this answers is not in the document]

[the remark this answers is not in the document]
sanjognaik: Hi @shashank-parulekar, as per suggestion, I have added assert here for checking buffer size alignment and used uint_fast32_t instead of uint64_t.

sns/parity_math.c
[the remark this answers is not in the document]
shashank-parulekar: I still do not see this change. Maybe you need to push this change.

sns/parity_math.c
```diff
@@ -541,6 +572,77 @@ static void xor_calculate(struct m0_parity_math *math,
...
+	 * 8-bytes (especially since block size currently is 4K) and this assert
+	 * is added with the hope that any deviation from this assumption is
+	 * caught during development instead of on the field.
+	 */
+	M0_ASSERT_INFO(m0_is_aligned(block_size, alignment),
+		       "block_size=%u is not %u-bytes aligned",
+		       block_size, alignment);
+
+	ALLOC_ARR_INFO(diff_data, block_size, "differential data block", ret);
+	if (diff_data == NULL)
```
shashank-parulekar, line 611: This might not work since you would have asserted in the above macro ALLOC_ARR_INFO. See if you can avoid using this macro.

shashank-parulekar: Sorry my mistake. Ignore this comment.

sns/parity_math.c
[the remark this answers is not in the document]
sanjognaik: Done

sns/parity_math.c
```diff
@@ -463,6 +490,9 @@ M0_INTERNAL int m0_parity_math_init(struct m0_parity_math *math,
 		ec_init_tables(data_count, parity_count,
 			       &math->pmi_encode_matrix[data_count * data_count],
 			       math->pmi_encode_tbls);
+
+		M0_LOG(M0_DEBUG, "use Intel ISA for parity calculation.");
```
shashank-parulekar, line 494: You need to take this log at the start of this else statement so that if any of the error paths are taken we would still know this was the Intel ISA path. In the current code if we have an error during allocation we would not know we had entered this path.
I will approve this PR but make this small change and no need to wait for approvals for this small change.

shashank-parulekar (approved):

PR #490, opened 2021-03-11
client-io: simplify pargrp_iomap_populate

andriytk:
Simplify pargrp_iomap_populate() by introducing and using m0_ivec_cursor_step_to() function.

### DIFF
--- a/lib/vec.h
+++ b/lib/vec.h
@@ -92,7 +92,7 @@ struct m0_vec_cursor {
 	/** Segment that the cursor is currently in. */
 	uint32_t             vc_seg;
 	/** Offset within the segment that the cursor is positioned at. */
-	m0_bindex_t          vc_offset;
+	m0_bcount_t          vc_offset;
 };
 
 /**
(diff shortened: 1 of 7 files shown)


andriytk:
cc @nikitadanilov

lib/vec.h
```diff
@@ -92,7 +92,7 @@ struct m0_vec_cursor {
 	/** Segment that the cursor is currently in. */
 	uint32_t             vc_seg;
 	/** Offset within the segment that the cursor is positioned at. */
-	m0_bindex_t          vc_offset;
```
huanghua78, line 95: I think we should keep  the original code.
This is offset, and offset is an index in the object.

lib/vec.h
```diff
@@ -700,7 +708,7 @@ struct m0_ivec_varr_cursor {
...
```
huanghua78, line 703: The same here.

lib/vec.h
andriytk: Offset is on how much to shift, isn't it? So it's a count, not an index.

andriytk:
I think the original code and this patch (which closely follows it) have a bug - if the segments in the indexvec under cursor overlap `m0_ivec_cursor_step_to(cursor, grpend)` may return the result which will be `>= grpsize` and the `rmw` flag will be false. But still the overlapping segments may not span the whole parity group in which case the `rmw` flag should be true.

I will prepare another patch...

andriytk:
See https://github.com/Seagate/cortx-motr/pull/491.

PR #500, opened 2021-03-16
EOS-18762 : Use two local endpoints for two way connection establishment in libfabric transport

upendrapatwardhan:
Signed-off-by: Upendra Patwardhan <upendra.patwardhan@seagate.com>

### DIFF
--- a/net/libfab/libfab.c
+++ b/net/libfab/libfab.c
@@ -138,6 +138,7 @@ static void libfab_txep_event_check(struct m0_fab__ep *txep,
 static void libfab_rxep_comp_read(struct fid_cq *cq, struct m0_fab__ep *ep);
 static int libfab_txep_init(struct m0_fab__active_ep *aep,
 			    struct m0_fab__tm *tm);
+static inline struct m0_fab__active_ep *libfab_aep_get(struct m0_fab__ep *ep);
 
 /* libfab init and fini() : initialized in motr init */
 M0_INTERNAL int m0_net_libfab_init(void)
@@ -504,6 +505,8 @@ static uint32_t libfab_handle_connect_request_events(struct m0_fab__tm *tm)
 			libfab_ep_ntop(cd->fcd_netaddr, &en);
 			libfab_fab_ep_find(tm, &en, cd->fcd_straddr, &ep);
 			rc = libfab_conn_accept(ep, tm, entry.info);
+			if (rc != FI_SUCCESS)
+				M0_LOG(M0_ERROR, "Conn accept failed %d", rc);
 			fi_freeinfo(entry.info);
 		}
 	} else if (rc == -FI_EAVAIL) {
@@ -552,16 +555,12 @@ static void libfab_rxep_comp_read(struct fid_cq *cq, struct m0_fab__ep *ep)
 
 	if (cq != NULL) {
 		rc = libfab_check_for_comp(cq, &buf, &len);
-		if (buf != NULL) {
-			if (rc > 0) {
-				if (buf->fb_length == 0)
-					buf->fb_length = len;
-				buf->fb_ev_ep = ep;
-				libfab_ep_get(ep);
-				libfab_buf_done(buf, 0);
-			}
-			else if (rc != -FI_EAGAIN)
-				libfab_buf_done(buf, -ECANCELED);
+		if (rc > 0) {
+			if (buf->fb_length == 0)
+				buf->fb_length = len;
+			buf->fb_ev_ep = ep;
+			libfab_ep_get(ep);
+			libfab_buf_done(buf, 0);
 		}
 	}
 }
@@ -582,7 +581,7 @@ static void libfab_poller(struct m0_fab__tm *tm)
 		usleep(500);
 		while(1) {
 			m0_mutex_lock(&tm->ftm_endlock);
-			if (tm->ftm_shutdown == true)
+			if (tm->ftm_shutdown)
 				break;
 			else if (m0_mutex_trylock(&tm->ftm_net_ma->ntm_mutex) 
 									 != 0) {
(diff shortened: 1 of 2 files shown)


net/libfab/libfab.c
```diff
@@ -508,6 +522,50 @@ static uint32_t libfab_handle_connect_request_events(struct m0_fab__tm *tm)
...
+ */
+static void libfab_rxep_comp_read(struct fid_cq *cq, struct m0_fab__ep *ep)
+{
+	struct m0_fab__buf *buf = NULL;
+	m0_bindex_t         len = 0;
+	int                 rc;
+
+	if (cq != NULL) {
+		rc = libfab_check_for_comp(cq, &buf, &len);
+		if (buf != NULL) {
```
huanghua78, line 555: For a function call, we usually check the return value first, and then check output parameters.

net/libfab/libfab.c
```diff
@@ -489,7 +503,7 @@ static uint32_t libfab_handle_connect_request_events(struct m0_fab__tm *tm)
 			cd = (struct m0_fab__conn_data*)entry.data;
 			libfab_ep_ntop(cd->fcd_netaddr, &en);
 			libfab_fab_ep_find(tm, &en, cd->fcd_straddr, &ep);
-			rc = libfab_active_ep_create(ep, tm, entry.info);
+			rc = libfab_conn_accept(ep, tm, entry.info);
```
huanghua78, line 506: result should be checked?

net/libfab/libfab.c
```diff
@@ -560,39 +617,12 @@ static void libfab_poller(struct m0_fab__tm *tm)
 
 		m0_tl_for(m0_nep, &tm->ftm_net_ma->ntm_end_points, net) {
 			xep = libfab_ep_net(net);
-			aep = (xep->fep_listen == NULL) ? xep->fep_send :
-			      xep->fep_listen->pep_tx_ep;
+			aep = (xep->fep_listen == NULL) ? xep->fep_aep :
+			       xep->fep_listen->pep_aep;
```
huanghua78, line 621: I propose to define an inline function for this, because same code is duplicated in multiple places.

net/libfab/libfab.c
```diff
@@ -641,18 +671,21 @@ static int libfab_ep_find(struct m0_net_transfer_mc *tm, const char *name,
...
+	int                       rc = 0;
 
 	M0_ENTRY();
 
 	m0_tl_for(m0_nep, &tm->ntm_end_points, net) {
-		xep = libfab_ep_net(net);
-		if (libfab_ep_cmp(xep, name, epn) == true) {
-			*epp = &xep->fep_nep;
+		ep = libfab_ep_net(net);
+		if (libfab_ep_cmp(ep, name, epn) == true) {
```
huanghua78, line 687: According to "doc/coding-style",
```
137   * use difference between NULL, 0 and "false" to emphasize whether
138     an expression is used as a pointer, integer (including success
139     or failure code) or boolean:
140 
141           if (p == NULL) { /* assumes that p is a pointer */
142           } else if (q != 0) { /* q is an integer */
143           } else if (r) { /* r is a boolean */
144           }
145 
146       Specifically, never use if (r == true);

```
this line should be:
```
if (libfab_ep_cmp(ep, name, epn)) {
```

net/libfab/libfab.c
```diff
@@ -662,8 +695,24 @@ static int libfab_ep_find(struct m0_net_transfer_mc *tm, const char *name,
...
 			rc = libfab_ep_create(tm, name, epn, epp);
 		else {
-			sprintf(ep, "libfab:%s:%s", epn->fen_addr, epn->fen_port);
-			rc = libfab_ep_create(tm, ep, epn, epp);
+			sprintf(ep_str, "libfab:%s:%s", epn->fen_addr,
+				epn->fen_port);
+			rc = libfab_ep_create(tm, ep_str, epn, epp);
+		}
+	} else {
+		if (name != NULL && epn != NULL) {
```
huanghua78, line 703: From this checking, I can see that "epn" might be NULL.
So, it is necessary to checking if it is NULL on line 698.

net/libfab/libfab.c
```diff
@@ -920,14 +1012,13 @@ static int libfab_passive_ep_create(struct m0_fab__ep *ep,
...
-	if (ep->fep_listen->pep_rx_ep == NULL)
+	M0_ALLOC_PTR(ep->fep_listen->pep_aep);
+	if (ep->fep_listen->pep_aep == NULL)
 		return M0_RC(-ENOMEM);
 
 	pep = ep->fep_listen;
+	ep->fep_listen->pep_aep->aep_rxep = NULL;
+	ep->fep_listen->pep_aep->aep_txep = NULL;
 
 	if (strlen(ep->fep_name.fen_port) != 0){
```
huanghua78, line 1023: An space is needed before the opening brace:
```
if (strlen(ep->fep_name.fen_port) != 0) {
```

huanghua78: And please check this for all other places.

net/libfab/libfab.c
upendrapatwardhan: Done

net/libfab/libfab.c
upendrapatwardhan: We can only add an entry to the error log in this case. The error handling is done inside the libfab_conn_accept() API.
Added an entry to the error log.

net/libfab/libfab.c
upendrapatwardhan: Done

net/libfab/libfab.c
upendrapatwardhan: Done

net/libfab/libfab.c
upendrapatwardhan: Added M0_ASSERT(epn != NULL)
In this function either one of the arguments from name or epn must not be NULL.

net/libfab/libfab.c
upendrapatwardhan: Done

net/libfab/libfab.c
huanghua78: I don't understand, if error happens here, why does the function "m0_net_libfab_init()" still return 0?

huanghua78 (commented):
I don't have objections.
Please have more people to review this, because this involve some new design and implementation of transport.

jamohite (approved):

PR #508, opened 2021-03-18
EOS-18209 : balloc PoC - updated balloc to start/prefer allocations from lowest group with free space.

mukundkanekar:
Signed-off-by: Mukund Kanekar <mukund.kanekar@seagate.com>

### DIFF
--- a/balloc/balloc.c
+++ b/balloc/balloc.c
@@ -964,6 +964,7 @@ static int balloc_init_internal(struct m0_balloc *bal,
 
 	bal->cb_be_seg = seg;
 	bal->cb_group_info = NULL;
+	bal->cb_lfreegroup = 0;
 	m0_mutex_init(&bal->cb_sb_mutex.bm_u.mutex);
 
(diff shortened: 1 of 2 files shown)


balloc/balloc.c
```diff
@@ -964,6 +964,7 @@ static int balloc_init_internal(struct m0_balloc *bal,
 
 	bal->cb_be_seg = seg;
 	bal->cb_group_info = NULL;
+	bal->cb_lfreegroup = 0;
```
shashank-parulekar, line 967: Currently we can check how the current code works and if it shows promise then we will need to move this cb_lfreegroup in superblock so that it is preserved across cluster reboots.

shashank-parulekar (approved):

balloc/balloc.c
mukundkanekar: Yes, this is raw code for PoC. 
Will do that in final code if we get significant improvements with this PoC.

huanghua78:
This task is done.
https://jts.seagate.com/browse/EOS-18209

PR #510, opened 2021-03-18
Clearing up optional builds

hessio:
Signed-off-by: Patrick Hession <patrick.hession@seagate.com>

## Problem

- The current guide does not make it clear enough that you can optionally do RPM generation or build from source.

## Changes

- Specified that you should not do RPM generation if you have built from source.

## Tested

- Tested on VMWare Workstation Pro 16

### DIFF
--- a/doc/Quick-Start-Guide.rst
+++ b/doc/Quick-Start-Guide.rst
@@ -70,7 +70,7 @@ Perform the below mentioned procedure to build the source code.
 RPM Generation
 ===============
 
-You can build RPMs instead of building from source. Do not do this if you have followed the above steps for build from source.
+You can also build RPMs by running the following command.
 
 - **$ make rpms**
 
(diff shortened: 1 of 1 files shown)


doc/Quick-Start-Guide.rst
```diff
@@ -66,11 +66,11 @@ Perform the below mentioned procedure to build the source code.
    - **$ scripts/m0 rebuild**
 
    **Note**: The **scripts/m0 rebuild** command includes some clean up.
-   
+ 
 RPM Generation
 ===============
 
-If you want to make sure about generating RPMs, run the below mentioned command.
+You can build RPMs instead of building from source. Do not do this if you have followed the above steps for build from source.
```
huanghua78, line 73: This is still misleading. Building binary executables and building RPMs are orthogonal. 
Users can do either of them, or both of them.
So, I would like to state this as:
```
You can also build RPMs by running the following command.
```

hessio: Yes but I have found that you get conflicts when you do both, which causes an error.

hessio: I have changed to your suggestion.

huanghua78 (approved):

PR #513, opened 2021-03-19
Fix user-space build on CentOS-8

andriytk:


andriytk:
Included at https://github.com/Seagate/cortx-motr/pull/523.

PR #517, opened 2021-03-19
Update Quick-Start-Guide.rst

vaibhavparatwar:
Minor changes to headings of each sections, added section names which were missing in the list at the start of document

### DIFF
--- a/doc/Quick-Start-Guide.rst
+++ b/doc/Quick-Start-Guide.rst
@@ -21,16 +21,20 @@ Procedure
 **********
 The below mentioned procedure must be followed to install the Motr component and to perform tests.
 
-1. Cloning the Source Code
+1. Clone the Source Code
 
-2. Building the Source Code
+2. Build the Source Code
 
-3. Running the Tests
+3. Generate RPMs
+
+4. Run the Tests
+
+5. Troubleshooting
 
 **Note**: The procedure mentioned above must be initiated in accordance with the order of listing.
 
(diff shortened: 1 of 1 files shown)


cortx-admin:

doc/Quick-Start-Guide.rst
```diff
@@ -21,16 +21,20 @@ Procedure
...
 
-2. Building the Source Code
+2. Build the Source Code
 
-3. Running the Tests
+3. Generate RPMs
+
+4. Run the Tests
+
+5. Troubleshooting
```
huanghua78, line 32: Generating RPMs is not a necessary step to run the tests.
Actually, in this guide, RPMs are only generated, not installed. Users/developers can install the result RPMs.
But running those UT and ST still requires the source code tree, because some of the UT/ST/scripts are only for developers in development environments, not for users in production environment.


And of course, troubleshooting can be referenced in any moment. It is not necessary to follow this order.

vaibhavparatwar: Ok. Got it. Needs to rearrange the sections then.. RPM generation section can go after Run test section.. while following steps, it may seem like it is a necessary step as it comes before the Running test section

huanghua78: Please refer to https://github.com/Seagate/cortx-motr/pull/510

huanghua78:
Duplicated.

PR #522, opened 2021-03-19
added state diagrams of deletion and iteration operation

RadhaGulhane13:
Signed-off-by: RadhaGulhane13 <radha.gulhane@seagate.com>


huanghua78:
On Line 67: 
 *   ancestor of a node is either its parent of the parent of an ancestor. The

s/ *   ancestor of a node is either its parent of the parent of an ancestor. The
  / *   ancestor of a node is either its parent or the parent of an ancestor. The

huanghua78:
Line 119:
 * node containing the target key. In the check is successful, the lookup

s
/ * node containing the target key. In the check is successful, the lookup
/ * node containing the target key. If the check is successful, the lookup

shashank-parulekar (changes requested):

btree/btree.c
```diff
@@ -235,10 +235,116 @@
...
+ *                                                |
+ *                                                v
+ *                                              DONE
+ * @endverbatim
  *
+ * Phases Description: 
+ * step 1. NEXTDOWN: travese down by searching for given key till we reach leaf node 
+ * step 2. LOAD : load left , right sibling  only if there are chances of underflow at the node (i.e. number of keys == min)
+ * step 3. CHECK : check traversal path of node and right, left sibling
+ *         if traverse path is invalid and checks fails it will start traversing again
```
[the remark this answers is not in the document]
RadhaGulhane13: okay

shashank-parulekar (approved):

RadhaGulhane13:
> On Line 67:
> 
> * ancestor of a node is either its parent of the parent of an ancestor. The
> 
> s/ * ancestor of a node is either its parent of the parent of an ancestor. The
> / * ancestor of a node is either its parent or the parent of an ancestor. The

changes are done as per suggestions:
 
image

btree/btree.c
```diff
@@ -230,10 +230,115 @@
...
+ *                                               v              |
+ *                                             CLEANUP<----------+
+ *                                                |
+ *                                                v
+ *                                              DONE
+ * @endverbatim
  *
+ * Phases Description: 
+ * step 1. NEXTDOWN: travese down by searching for given key till we reach leaf node 
+ * step 2. LOAD : load left , right sibling  only if there are chances of underflow at the node (i.e. number of keys == min)
```
huanghua78, line 270: In case of variable length of key and/or variable length of value, the number of keys/values that an internal node or a left node can contain, are not constant. 
How do we define the minimal of key/val?

btree/btree.c
```diff
@@ -230,10 +230,115 @@
  *
  * Deletion (DEL)
  * ..............
+ * There are basically 2 cases for deletion
+ * 1. No underflow after deletion
+ * 2. Underflow after deletion
+ *  a. Balance by borrowing key from sibling
+ *  b. Balance by merging with sibling
+ *    b.i. No underflow at internal node
```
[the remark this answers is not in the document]
RadhaGulhane13: If we merge with sibling , there will be decrement in number of keys of parent by one , so underflow can occur at parent internal node.
below is an example of such case:
image

btree/btree.c
```diff
@@ -230,10 +230,115 @@
...
+ * If next_sibling_flag == 1, we will also load next sibling and , we need to fetch key next to given key
+ * As we are releasing lock for finding next node, updates such as(insertion of new keys, merging due to deletion can happen), 
+ * so in such cases, we will load both earlier node and node to the right of it
+ * 
+ * Phases Description: 
+ * step 1. NEXTDOWN: this state will load nodes searching by the given key if next_sibling_flag == 0, 
+ *                   but if next_sibling_flag == 1 then this state will load the leftmost child nodes during its downward traversal.
+ * step 2. LOAD: this function will get called only when next_sibling_flag == 1, 
+ *               and functionality of this function is to load next node so, it will search and LOAD next sibling 
+ * step 3. CHECK: check function will check the traversal path for node with key and traversal path for node next node if it is also loaded
```
huanghua78, line 324: Please check this line. I don't understand this:
```
traversal path for node next node if it is also loaded
```

btree/btree.c
```diff
@@ -230,10 +230,115 @@
...
+ *                if any of the path is invalid and checks fails it will start traversing again
```
huanghua78, line 325: s/checks/check
?

btree/btree.c
RadhaGulhane13: it should be : traversal path for  next node if it is also loaded

btree/btree.c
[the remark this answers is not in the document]
huanghua78: I see. So, we may say: b.ii. Underflow at its parent node
?

btree/btree.c
RadhaGulhane13: check

btree/btree.c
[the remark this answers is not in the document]
RadhaGulhane13: yes

btree/btree.c
RadhaGulhane13: In such cases, if we have node size defined , we can consider node size instead of number of keys to determine if underflow can occur

btree/btree.c
```diff
@@ -314,17 +313,17 @@
...
- * As we are releasing lock for finding next node, updates such as(insertion of new keys, merging due to deletion can happened), 
- * so in such cases, we will load both earlier node and node next to it
+ * As we are releasing lock for finding next node, updates such as(insertion of new keys, merging due to deletion) can happen, 
+ * so in such cases, we will load both earlier node and node to the right of it
  * 
  * Phases Description: 
- * step 1. NEXTDOWN: while loading next node, it will traverse and load leftmost child, 
- *                   else it will traverse and load nodes by searching for given key. 
+ * step 1. NEXTDOWN: this state will load nodes searching by the given key if next_sibling_flag == 0, 
+ *                   but if next_sibling_flag == 1 then this state will load the leftmost child nodes during its downward traversal.
```
shashank-parulekar, line 321: Change
this state will load nodes searching by the given key if next_sibling_flag == 0, 
                    but if next_sibling_flag == 1 then this state will load the leftmost child nodes during its downward traversal.
to
this state will load nodes searching for the given key,
                    but if next_sibling_flag == 'True' then this state will also load the leftmost child nodes during its downward traversal.

btree/btree.c
```diff
@@ -314,17 +313,17 @@
...
  * step 2. LOAD: this function will get called only when next_sibling_flag == 1,
```
shashank-parulekar, line 322: Change
get called only when next_sibling_flag == 1
to
get called only when next_sibling_flag == 'True'

btree/btree.c
```diff
@@ -314,17 +313,17 @@
...
- *                   else it will traverse and load nodes by searching for given key. 
+ * step 1. NEXTDOWN: this state will load nodes searching by the given key if next_sibling_flag == 0, 
+ *                   but if next_sibling_flag == 1 then this state will load the leftmost child nodes during its downward traversal.
  * step 2. LOAD: this function will get called only when next_sibling_flag == 1, 
  *               and functionality of this function is to load next node so, it will search and LOAD next sibling 
- * step 3. CHECK: check function will check the traversal path for node with key and traversal path for node next node if it is also loaded
- *                if any of the path is invalid and checks fails it will start traversing again  
- *                else, if next_sibling_flag ==1, check will go to NEXTKEY to fetch next key, else to ACT for callback 
+ * step 3. CHECK: check function will check the traversal path for node with key and traversal path for next sibling node if it is also loaded
+ *                if traverse path of any of the node has changed, repeat traversal again after UNLOCKING the tree
```
shashank-parulekar, line 325: Wherever the flag is set to '1' replace that with 'True' and '0' with 'False'

shashank-parulekar (changes requested):

PR #527, opened 2021-03-22
[DTM0] EOS-18811: Finalize DTX when it enters DONE state.

ivan-alekhin:
The patch adds finalization of DTX after the client finished
working with it.
Also, the patch addresses CR comments for the previous patch
in the series.


just-now (commented):
Looks good to me.

just-now (approved):
+1

PR #535, opened 2021-03-23
m0crate: fix segv crash when MOTR_CONFIG is missing

andriytk:
When the MOTR_CONFIG section is missing in the configuration
file or it's named differently (for example MERO_CONFIG) -
m0crate just crashes with the following stack:
```
  0  0x000000000040c05e in atoi (__nptr=0x223a0f0 "8") at /usr/include/stdlib.h:280
  1  copy_value (load=0x2226010, max_workload=<optimized out>, index=0x7fff0f6bc524,
      key=<optimized out>, value=<optimized out>) at motr/m0crate/parser.c:281
  2  0x000000000040ccec in parse_yaml_file (load=load@entry=0x2226010, max_workload=max_workload@entry=32,
      index=index@entry=0x7fff0f6bc524, config_file=<optimized out>) at motr/m0crate/parser.c:588
  3  0x0000000000403eca in main (argc=3, argv=0x7fff0f6bc648) at motr/m0crate/crate.c:1345
```
Which is a bit confusing because the problem is actually at
frame 1 (not 0):
```
  (gdb) f 1
  1  copy_value (load=0x2226010, max_workload=<optimized out>, index=0x7fff0f6bc524,
      key=<optimized out>, value=<optimized out>) at motr/m0crate/parser.c:281
  281				conf->layout_id = atoi(value);
  (gdb) p conf
  $1 = (struct crate_conf *) 0x0
```
Solution: check the conf pointer at copy_value() and print
a proper error msg when MOTR_CONFIG section is missing in the
config file.

### DIFF
--- a/motr/m0crate/parser.c
+++ b/motr/m0crate/parser.c
@@ -227,24 +227,32 @@ static int parse_int(const char *value, enum config_key_val tag)
 #define workload_index(t) (t->u.cw_index)
 #define workload_io(t) (t->u.cw_io)
 
+const char conf_section_name[] = "MOTR_CONFIG";
+
 int copy_value(struct workload *load, int max_workload, int *index,
 		char *key, char *value)
 {
+	int                       value_len = strlen(value);
 	struct workload          *w = NULL;
 	struct m0_fid            *obj_fid;
 	struct m0_workload_io    *cw;
 	struct m0_workload_index *ciw;
-	int                       value_len = strlen(value);
 
-	if (!strcmp(value, "MOTR_CONFIG")) {
+	if (m0_streq(value, conf_section_name)) {
 		if (conf != NULL) {
-			cr_log(CLL_ERROR, "YAML file error. More than one config sections");
+			cr_log(CLL_ERROR, "YAML file error: "
+			       "more than one config sections\n");
 			return -EINVAL;
 		}
 
 		conf = m0_alloc(sizeof(struct crate_conf));
 		if (conf == NULL)
 			return -ENOMEM;
+	}
+	if (conf == NULL) {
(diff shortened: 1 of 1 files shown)


huanghua78:
Another problem is that:
```
313                 case WORKLOAD_TYPE:
314                         (*index)++;
315                         w = &load[*index];
316                         if (atoi(value) == INDEX) {
317                                 w->cw_type = CWT_INDEX;
318                                 w->u.cw_index = m0_alloc(SIZEOF_CWIDX);
319                                 if (w->u.cw_io == NULL)
320                                         return -ENOMEM;
321                         } else {
322                                 w->cw_type = CWT_IO;
323                                 w->u.cw_io = m0_alloc(SIZEOF_CWIO);
324                                 if (w->u.cw_io == NULL)
325                                         return -ENOMEM;
326                         }
327                         return workload_init(w, w->cw_type);
328                 case SEED:
329                         w = &load[*index];
330                         if (w->cw_type == CWT_IO) {
331                                 cw = workload_io(w);
332                                 if (strcmp(value, "tstamp"))
333                                         w->cw_rstate = atoi(value);
```


The m0crate assumes that the "WORKLOAD_TYPE" is the first key for WORKLOAD_SPEC.
`*index` is only increased from initial value (-1) to a valid number on Line 314.
Otherwise, panic will happen, e.g. on line 329, because the "*index" is an invalid number (-1 in this case).

andriytk:
@huanghua78 good catch! Should be fixed now.

huanghua78 (approved):
Looks fine. Thank you, @andriytk

somnathbghule (approved):
Fix looks fine, Thank you @andriytk.

PR #543, opened 2021-03-26
[DTM0] EOS-17644: add the client to all2all integration test

sergey-shilov:
Now the m0crate is used to put key/val pairs with DTX
enabled to check that whole chain works.

Some workarounds are added to start the DTM0 service
on the client side and wait for connections from the
cluster nodes. Some fixes provided.

Also the patch for Hare with partial DTM0 support is
attached to have the ability to run the integration
test.

### DIFF
--- a/dtm0/it/all2all/m0crate.yaml
+++ b/dtm0/it/all2all/m0crate.yaml
@@ -1,0 +1,35 @@
+CrateConfig_Sections: [MOTR_CONFIG, WORKLOAD_SPEC]
+MOTR_CONFIG:
+   MOTR_LOCAL_ADDR:  10.230.248.89@tcp:12345:4:1
+   MOTR_HA_ADDR:    10.230.248.89@tcp:12345:1:1
+   PROF: 0x7000000000000001:0x5f
+   LAYOUT_ID: 1                           # Layout id defines the unit size.
+   IS_OOSTORE: 1                           # Is oostore-mode?
+   IS_READ_VERIFY: 0                       # Enable read-verify?
+   TM_RECV_QUEUE_MIN_LEN: 2         # Minimum length of the receive queue, default is 2
+   MAX_RPC_MSG_SIZE: 131072         # Maximum rpc message size, default is 131072 (128k)
+   PROCESS_FID: 0x7200000000000001:0x36
+   IDX_SERVICE_ID: 1
+   ADDB_INIT: 1
+
+WORKLOAD_SPEC:
+  WORKLOAD:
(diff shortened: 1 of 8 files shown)


dtm0/it/all2all/m0crate.yaml
```diff
@@ -0,0 +1,35 @@
+CrateConfig_Sections: [MOTR_CONFIG, WORKLOAD_SPEC]
+MOTR_CONFIG:
+   MOTR_LOCAL_ADDR:  10.230.248.89@tcp:12345:4:1
```
vaibhavparatwar, line 3: Just curious question : is there a reason behind MOTR_LOCAL_ADDR and MOTR_HA_ADDR to be hardcoded?

ivan-alekhin (commented):
The code looks good. Some improvements can be added in next patches.
I checked out the branch and checked the test. The test passed.
Let's rebase this branch. After rebasing, please run the `idx-dix-mt:dtm0` test. It should not work with this version of the patch because we have added start/stop of DTM0 service into the client. In the UT, start/stop is done explicitly. Let's fix this too if it is possible.

dtm0/it/all2all/m0crate.yaml
sergey-shilov: Yes, they need to be specified. Another solution is to use a template that is pre-processed before running of m0crate (like it is done by io_workload script), but I would say that it may be treated as further improvements.

vaibhavparatwar: Ok. May be we should make note of it somewhere.. or have a readme so the test runner knows that they need to specify these IP addresses in m0crate.yaml file ? If it (instruction/Note) is  already available, please ignore the comment.

ivan-alekhin (approved):
As per today's discussion, let's merge it as it is. The UT issue will be addressed in the next patch.

dtm0/it/all2all/m0crate.yaml
sergey-shilov: Good point, added README.md.

just-now (commented):
Later on, please do:

move config to *.in file:
mv dtm0/it/all2all/m0crate.yaml  dtm0/it/all2all/m0crate.yaml.in


in all2all, add:
LOCAL_IP=$(ifconfig eth0 | grep 'inet ' | awk '{ print $2 }')
cat dtm0/it/all2all/m0crate.yaml.in | sed s"/###LOCAL_IP###/$LOCAL_IP/g" > m0crate.yaml

PR #523, opened 2021-03-20
Fix build on CentOS-8

andriytk:
Fix build on CentOS-8.3.

Note: gccxml is deprecated and is no longer available on CentOS-8, use castxml instead.

Closes #502.

Here is CentOS-8 VM with all dependencies installed ready to build and run Motr - https://github.com/andriytk/cortx-motr/releases/tag/centos83ova.

PR #544, opened 2021-03-27
Fix build on Gentoo Linux

pasis:
Gentoo Linux has rolling updates and thus provides latest toolchain versions. This patchset contains of a subset of the patches required to build Motr on Gentoo system with latest software versions. It doesn't contain patches conflicting with #523 and should be treated as addition to that PR.

### DIFF
--- a/lib/user_space/timer.c
+++ b/lib/user_space/timer.c
@@ -67,8 +67,8 @@ M0_TL_DEFINE(tid, static, struct m0_timer_tid);
    gettid(2) implementation.
    Thread-safe, async-signal-safe.
  */
-static pid_t gettid() {
-
+static pid_t _gettid()
+{
 	return syscall(SYS_gettid);
 }
 
@@ -132,7 +132,7 @@ M0_INTERNAL int m0_timer_thread_attach(struct m0_timer_locality *loc)
 
 	M0_PRE(loc != NULL);
 
-	tid = gettid();
+	tid = _gettid();
 	M0_ASSERT(locality_tid_find(loc, tid) == NULL);
 
 	M0_ALLOC_PTR(tt);
@@ -157,7 +157,7 @@ M0_INTERNAL void m0_timer_thread_detach(struct m0_timer_locality *loc)
 
 	M0_PRE(loc != NULL);
 
-	tid = gettid();
+	tid = _gettid();
 	tt = locality_tid_find(loc, tid);
 	M0_ASSERT(tt != NULL);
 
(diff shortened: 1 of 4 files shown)


cortx-admin:

madhavemuri (approved):

motr/cob.c
```diff
@@ -512,10 +512,13 @@ static void cob_fail_op(struct m0_op *op, int rc)
...
+		/* fallthrough */
 	case M0_ES_CREATING:
 		m0_sm_move(&op->op_entity->en_sm, 0, M0_ES_OPEN);
+		/* fallthrough */
 	case M0_ES_OPEN:
 		m0_sm_move(&op->op_entity->en_sm, 0, M0_ES_CLOSING);
+		/* fallthrough */
 	default:
 		m0_sm_move(&op->op_entity->en_sm, 0, M0_ES_INIT);
 	}
```
andriytk, line 524: @nikitadanilov should we really fallthrough here?

andriytk (approved):

madhavemuri (approved):

huanghua78:
https://github.com/Seagate/cortx-motr/pull/523 is merged.

lib/user_space/timer.c
```diff
@@ -67,8 +67,8 @@ M0_TL_DEFINE(tid, static, struct m0_timer_tid);
    gettid(2) implementation.
    Thread-safe, async-signal-safe.
  */
-static pid_t gettid() {
-
+static pid_t m0_gettid()
```
huanghua78, line 70: What function name is this gettid() conflicting with?
The system call gettid()?

Usually m0_ prefix is used for M0_INTERNAL or M0_EXTERNAL.
For static function, we don't use the m0_ prefix.

Can we change it to _gettid()?

huanghua78:
@pasis can you please provide a OVA with Gentoo Linux for evaluation?

lib/user_space/timer.c
pasis: Done.

pasis: And yes, it conflicts with glibc version of gettid().

pasis:
> @pasis can you please provide a OVA with Gentoo Linux for evaluation?

@huanghua78, I use gentoo as my main system and I've built it from scratch (gentoo installation differs from binary distributions). The only image I know is here https://www.osboxes.org/gentoo/. But the image is old and requires updating the system what can be an issue for non-gentoo users. So I'm afraid I cannot help with the image.

huanghua78 (approved):

huanghua78:
OK to test

PR #546, opened 2021-03-29
[DTM0] EOS-19055: Fix DTM0 UTs (clnt and svc)

ivan-alekhin:
[DTM0] EOS-19055: Fix DTM0 UT for m0 client.

The patch addresses the issue where the UT
used to test DTM0 with M0 client failed.
The root cause was the part where the client
starts the DTM0 service: the FID of the client
matches with the FID of the server, and because
of that the wrong DTM0 service FID was selected,
that led to the situation where the server-side
DTM0 service sends P messages to itself (rather
than to the client-side DTM0 service).
Also the patch alleviates the failures of the DTM0
service UT: it still fails but only at the point
of finalization (connections are not terminated
properly and rconf complaints about objects
that disappeared without proper notice).

DTM0: Remove event-driven connectivity in UT.

The patch fixes the problem where the DTM0 service
UT does not work properly when the HA based
mechanism is used to establish connections
between the client and the server.
Also, the patch addresses the question on the
hardcoded-dtm0s-fid fault point by replacing it
with m0_dtm0_in_ut which a better choice for
UT-specific branches in non-UT code.

### DIFF
--- a/motr/ut/idx_dix.c
+++ b/motr/ut/idx_dix.c
@@ -788,7 +788,6 @@ struct m0_client* st_get_instance()
 struct dtm0_ut_ctx {
 	struct m0_fid           duc_cli_svc_fid;
 	struct m0_fid           duc_srv_svc_fid;
-	struct m0_reqh_service *duc_cli_svc;
 	struct m0_reqh_service *duc_srv_svc;
 	struct m0_idx           duc_idx;
 	struct m0_container     duc_realm;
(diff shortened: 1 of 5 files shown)


motr/ut/idx_dix.c
```diff
@@ -804,34 +803,30 @@ static int duc_setup(void)
 							    0x1a);
 	struct m0_fid            srv_dtm0_fid = M0_FID_INIT(0x7300000000000001,
 							    0x1c);
-	struct m0_reqh_service  *cli_srv;
 	struct m0_reqh_service  *srv_srv;
 	struct m0_reqh          *srv_reqh;
 	int                      rc;
 	struct m0_container     *realm = &duc.duc_realm;
 
+	m0_fi_enable("m0_client_init", "hardcoded-dtm0s-fid");
```
sergey-shilov, line 811: We already have m0_dtm0_in_ut() function, propose to enable it in addition to hardcoded fid to avoid sleep 15secs during client start. Also may be we can use m0_dtm0_in_ut() instead of hardcoded-dtm0s-fid FI.

ivan-alekhin: Unfortunately, `m0_dtm0_in_ut` leads to the situation where we dtm0/service.c has a bit different behavior. `process_clink_cb` is called twice with the same event for the same process:

```c
$4 = (struct m0_fid *) 0x7fffedbb9d90 <dix_ut_sctx+33584>
(gdb) p /x current_proc_fid[0]
$5 = {f_container = 0x7200000000000001, f_key = 0x5}
(gdb) p /x evented_proc_fid[0]
$6 = {f_container = 0x7200000000000001, f_key = 0x6}
(gdb) p process
$7 = (struct dtm0_process *) 0x2b08790
(gdb) p evented_proc_state
$8 = M0_NC_ONLINE
```

It causes the situation where we are trying to initialize the same RPC link twice inside `m0_dtm0_service_process_connect`.

So that I deliberately decided not to use `m0_dtm0_in_ut`, and use a separate FP instead Because of that, the client UT uses explicit `m0_dtm0_service_process_connect` call (see `duc_setup`) to establish the backward connection (server -> client).

ivan-alekhin:
I have rebased the branch and added the patch proposed by @sergey-shilov . The UTs and the integration test pass successfully.

PR #547, opened 2021-03-29
EOS-17644 dtm0: override m0crate cfg in all2all test

sergey-shilov:
Now hctl status output is used to get cluster params
and set them in m0crate cfg file before running m0crate.


just-now (approved):
looks good.

ivan-alekhin (approved):

PR #548, opened 2021-03-30
[EOS-18074] Adding libfab Transport for all the STs

VinothVijayakumar-dev:
[EOS-18074] Adding libfabric transport in ST scripts
	- update XPT variable to libfab, if libfab is available

Signed-off-by: Vinoth.V <vinoth.v@seagate.com>


console/st/console-st.sh
```diff
@@ -45,6 +44,14 @@ CONF_PROFILE='<0x7000000000000001:0>'
 NODE_UUID=02e94b88-19ab-4166-b26b-91b51f22ad91  # required by `common.sh'
 . $M0_SRC_DIR/m0t1fs/linux_kernel/st/common.sh  # modload_galois
 
+cmd=$(whereis fi_info | cut -d ':' -f2)
```
madhavemuri, line 47: Move this logic to m0t1fs/linux_kernel/st/common.sh

Also, add an option to run-st to specify the transport type, which will help in testing both the transports at the same time while in jenkins or running ST's in VM.

conf/st
```diff
@@ -53,6 +53,8 @@ M0_SRC_DIR=${M0_SRC_DIR%/*/*}
 
 M0CONFGEN=$M0_SRC_DIR/utils/m0confgen
 
+XPRT=$(m0_default_xpt)
```
huanghua78, line 56: Sorry, I didn't see where the variable or function "m0_default_xpt" is defined.

Is it defined by command line argument?
or environment variable?

Please add comments to say how it is defined, and how people can change it.

scripts/m0
```diff
@@ -41,6 +41,26 @@ M0_SRC_DIR="${M0_SRC_DIR%/*/*}"
 
 SUDO="${SUDO-sudo -E}"  # the absence of ':' is important
 
+#choose the transport 
+_transport_select() {
```
huanghua78, line 45: Usually, we don't use interactive way to get such selection from user's input.

We can use command line arguments, or environment variables. Please refer to,
```
253 Environment variables:
254     CONFIGURE_OPTS      Extra options for './configure' command.
255     MAKE_OPTS           Extra options for 'make' command.
256 

```

For example, we can use the "NET_XPRT" as the env variable to pass user's selection.
A user can do:
```
NET_XPRT=libfab ./scripts/m0 run-st 
```

or
```
NET_XPRT=lnet ./scripts/m0 run-st 
```

Then in "scripts/m0", we can check this variable to select a network transport.
If this variable is not defined, we can default to lnet or libfab.

huanghua78 (approved):

upendrapatwardhan (approved):

PR #550, opened 2021-03-30
Addb2 instrumentation for DTM0

mehjoshi:

### DIFF
--- a/dtm0/fop.h
+++ b/dtm0/fop.h
@@ -80,7 +80,8 @@ struct dtm0_fom {
 	struct m0_fom dtf_fom;
 };
 
-M0_INTERNAL void m0_dtm0_on_committed(struct m0_reqh               *reqh,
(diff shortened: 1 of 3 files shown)


just-now (commented):
Looks like the patch is not complete.

dtm0/fop.h
```diff
@@ -80,7 +80,8 @@ struct dtm0_fom {
 	struct m0_fom dtf_fom;
 };
 
-M0_INTERNAL void m0_dtm0_on_committed(struct m0_reqh               *reqh,
+M0_INTERNAL void m0_dtm0_on_committed(struct m0_fom                *fom,
```
huanghua78, line 83: Do you think we can make this "*fom" as const?

huanghua78:
closing stale PR. 
If this is still needed, please file a new PR.

PR #554, opened 2021-03-31
[DTM0] Use rt checks agains DTM0 configure flag.

ivan-alekhin:
The patch renames the DTM0 configure flag
into ENABLE_DTM0, sets this flag to zero when
"enable-dtm0" was not defined at the configure
step, and, finally, this patch converts
the DTM0-related pre-processor conditions into
runtime (rt in the title) conditions.

### DIFF
--- a/dix/req.c
+++ b/dix/req.c
@@ -1365,14 +1365,9 @@ static void dix__rop(struct m0_dix_req *req, const struct m0_bufvec *keys,
 	M0_PRE(req->dr_indices_nr == 1);
 
 	keys_nr = keys->ov_vec.v_nr;
-#ifndef DTM0
 	M0_PRE(keys_nr != 0);
-#else
 	/* We support only one record per request in DTM0. */
-	M0_PRE(ergo(M0_IN(req->dr_type, (DIX_CREATE, DIX_DELETE,
-					 DIX_PUT, DIX_DEL)),
-		    keys_nr == 1));
-#endif
+	M0_PRE(ergo(req->dr_dtx != NULL, keys_nr == 1));
 	M0_ALLOC_PTR(rop);
 	if (rop == NULL) {
 		dix_req_failure(req, M0_ERR(-ENOMEM));
@@ -1853,12 +1848,11 @@ static uint32_t dix_rop_tgt_iter_max(struct m0_dix_req    *req,
 		 */
 		return m0_dix_liter_P(iter);
 	else
-#ifndef DTM0
-		return m0_dix_liter_N(iter) + 2 * m0_dix_liter_K(iter);
-#else
-		/* Skip spares in DTM0 */
-		return m0_dix_liter_N(iter) + m0_dix_liter_K(iter);
-#endif
+		/* Skip spares in ENABLE_DTM0 */
+		if (ENABLE_DTM0)
+			return m0_dix_liter_N(iter) + m0_dix_liter_K(iter);
+		else
+			return m0_dix_liter_N(iter) + 2 * m0_dix_liter_K(iter);
 }
 
 static void dix_rop_tgt_iter_next(const struct m0_dix_req *req,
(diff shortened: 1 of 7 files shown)


dix/req.c
```diff
@@ -1365,14 +1365,11 @@ static void dix__rop(struct m0_dix_req *req, const struct m0_bufvec *keys,
...
 	keys_nr = keys->ov_vec.v_nr;
-#ifndef DTM0
-	M0_PRE(keys_nr != 0);
-#else
+	M0_PRE(ergo(!ENABLE_DTM0, keys_nr != 0));
 	/* We support only one record per request in DTM0. */
-	M0_PRE(ergo(M0_IN(req->dr_type, (DIX_CREATE, DIX_DELETE,
-					 DIX_PUT, DIX_DEL)),
+	M0_PRE(ergo(ENABLE_DTM0 && M0_IN(req->dr_type, (DIX_CREATE, DIX_DELETE,
+							DIX_PUT, DIX_DEL)),
```
sergey-shilov, line 1371: Propose to use the following condition
=====================================
	/* We support only one record per request in DTM0. */
	M0_PRE(ergo(req->dr_dtx != NULL && M0_IN(req->dr_type,
						 (DIX_CREATE, DIX_DELETE,
						  DIX_PUT, DIX_DEL)),
		    keys_nr == 1));
=====================================
as we don't have any restrictions for the operations that go without dtx.

dix/req.c
```diff
@@ -1853,12 +1850,11 @@ static uint32_t dix_rop_tgt_iter_max(struct m0_dix_req    *req,
...
-		return m0_dix_liter_N(iter) + 2 * m0_dix_liter_K(iter);
-#else
-		/* Skip spares in DTM0 */
-		return m0_dix_liter_N(iter) + m0_dix_liter_K(iter);
-#endif
+		/* Skip spares in ENABLE_DTM0 */
+		if (ENABLE_DTM0)
+			return m0_dix_liter_N(iter) + m0_dix_liter_K(iter);
+		else
+			return m0_dix_liter_N(iter) + 2 * m0_dix_liter_K(iter);
```
sergey-shilov, line 1857: Why do we need to substitute #ifndef here with runtime check?

dix/req.c
```diff
@@ -2133,10 +2129,8 @@ static void dix_rop_units_set(struct m0_dix_req *req)
 			unit = &rec_op->dgp_units[j];
 			dix_rop_tgt_iter_next(req, rec_op, &tgt,
 					      &unit->dpu_is_spare);
-#ifdef DTM0
-			/* We do not operate with spares in DTM0. */
-			M0_ASSERT(!unit->dpu_is_spare);
-#endif
+			M0_ASSERT_INFO(ergo(ENABLE_DTM0, !unit->dpu_is_spare),
+				       "We do not operate with spares in DTM0");
```
sergey-shilov, line 2133: Why do we need to substitute #ifndef here with runtime check?

motr/client_init.c
```diff
@@ -1618,40 +1613,39 @@ int m0_client_init(struct m0_client **m0c_p,
...
-		 */
-		cli_svc_fid  = M0_FID_INIT(0x7300000000000001, 0x1a);
-	}
-
 #ifndef __KERNEL__
-	(void) m0_dtm__client_service_start(&m0c->m0c_reqh, &cli_svc_fid);
-#endif
-	m0c->m0c_dtms = m0_dtm0_service_find(&m0c->m0c_reqh);
-	M0_ASSERT(m0c->m0c_dtms != NULL);
+	if (ENABLE_DTM0) {
```
sergey-shilov, line 1617: Why do we need to substitute #ifndef here with runtime check? That makes the resulting binary more heavy with the branch of code that never happen for particular build.

motr/idx.c
```diff
@@ -157,20 +156,16 @@ static int idx_op_init(struct m0_idx *idx, int opcode,
 	m0_op_idx_bob_init(oi);
 	m0_ast_rc_bob_init(&oi->oi_ar);
 
-#if defined(DTM0)
-	if (M0_IN(op->op_code, (M0_IC_PUT, M0_IC_DEL))) {
-		m0c = entity->en_realm->re_instance;
+	if (ENABLE_DTM0 && M0_IN(op->op_code, (M0_IC_PUT, M0_IC_DEL))) {
```
sergey-shilov, line 159: Why do we need to substitute #ifndef here with runtime check? That makes the resulting binary more heavy with the branch of code that never happen for particular build.

ivan-alekhin:
> General question: since we leave a build flag as is why do we need to get rid of pre-processor macros substituting them with runtime checks? Configuration definitions make some branches of code never happen, it is the exact use case for pre-processor macros to have the resulting binary more light weight. For other places we can use check for NULL dtx field.

For the record, here are the results of our discussions:

The change (s/ifdef/if) is an intermediate step between the point where DTM0 exists as an optional code (under "ifdefs") and the point where DTM0 is fully integrated into the upstream codebase (a few "ifs" + checks of the conf cache). The main advantage of the changes proposed in the patch is that the DTM0-related code will be compiled whether "enable-dtm0" is specified or not. It will help us to make sure that the DTM0-related code paths won't be broken at least from compilation perspective by changes in the upstream. Besides, it will greatly help us with the follow-up transition to the "runtime-only" way (conf cache, a global variable or anything like that).

sergey-shilov (approved):
Looks good.

PR #555, opened 2021-03-31
[DTM0] EOS-18958: Add ADDB2 instrumentation

mehjoshi:

### DIFF
--- a/dtm0/fop.c
+++ b/dtm0/fop.c
@@ -20,7 +20,14 @@
  */
 
 
+#define M0_TRACE_SUBSYSTEM M0_TRACE_SUBSYS_DTM0
+#include "lib/trace.h"
+#include "addb2/addb2.h"
 #include "be/dtm0_log.h"
+#include "dtm0/addb2.h"
+#include "dtm0/fop.h"
+#include "dtm0/fop_xc.h"
+#include "dtm0/service.h"
 #include "dtm0/tx_desc.h"
 #include "dtm0/tx_desc_xc.h"
 #include "lib/errno.h"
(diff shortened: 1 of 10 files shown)


motr/idx.c
```diff
@@ -162,6 +164,9 @@ static int idx_op_init(struct m0_idx *idx, int opcode,
 		oi->oi_dtx = m0_dtx0_alloc(m0c->m0c_dtms, oi->oi_sm_grp);
 		if (oi->oi_dtx == NULL)
 			return M0_ERR(-ENOMEM);
+		did = m0_sm_id_get(&oi->oi_dtx->tx_dtx->dd_sm);
+		cid = m0_sm_id_get(&op->op_sm);
+		M0_ADDB2_ADD(M0_AVI_CLIENT_TO_DIX, cid, did);
```
ivan-alekhin, line 169: client-to-dix? shouldn't it be client-to-dtx?

dtm0/fop.c
```diff
@@ -235,6 +270,11 @@ static void m0_dtm0_send_msg(struct m0_dtm0_service *dtms,
 	req               = m0_fop_data(fop);
 	req->dtr_msg      = msg_type;
 	rc = m0_dtm0_tx_desc_copy(txd, &req->dtr_txr) ?: m0_rpc_post(item);
+
+	phase_sm_id = m0_sm_id_get(&fom->fo_sm_phase);
+	rpc_sm_id   = m0_sm_id_get(&item->ri_sm);
+	M0_ADDB2_ADD(M0_AVI_FOM_TO_TX, phase_sm_id, rpc_sm_id);
```
ivan-alekhin, line 276: Why do we re-use fom-to-tx here? shouldn't it be something like "fom-to-pmsg"?

ivan-alekhin (commented):
@just-now , is it a temporary solution? I mean the cases where we are re-using already defined mappings. Are we going to define our own mappings later on?

just-now (approved):
+1

dtm0/fop.c
just-now: Let's leave this mapping as is, today, for new scripts mappings are type-less. Eventually we'll remove all typed mappings from the instrumentation schema.

motr/idx.c
just-now: shouldn't, please look at my comment above.

PR #565, opened 2021-04-07
EOS-14402: BE log header doesn't always point to a valid log record

gshipra:
Problem:
If log IO fails just after the log header is updated just before log record overwrites
old record the header was pointing for, then the log header may point to a garbage
instead of pointing to a valid log record. When the engine starts again
recovery won't happen properly.

Soution
1. Add a counter to log header and record header structure.
2. Intiate log header counter to 0 when log is created.
3. Populate the record header counter with log header counter
at the time of record write.
4. Check the record header counter and log header counter at the
time of creating a list of records for recovery. If they do not match
than skip the record and further records.
5. Increment the log header counter at the time of be_engine start, after
recovery completes.

Testing
All UT tests passed.

### DIFF
--- a/be/engine.c
+++ b/be/engine.c
@@ -814,6 +814,10 @@ M0_INTERNAL int m0_be_engine_start(struct m0_be_engine *en)
 	if (rc != 0)
 		return M0_ERR(rc);
 
+	rc = be_log_header_inc_generation(&(en->eng_log), &(en->eng_log.lg_header));
+	if (rc != 0)
+		return M0_ERR(rc);
+
 	recovery_time = m0_time_now();
 	be_engine_try_recovery(en);
 
--- a/be/fmt.h
+++ b/be/fmt.h
@@ -200,7 +200,7 @@ struct m0_be_fmt_log_store_header {
 struct m0_be_fmt_group_cfg;
 
 struct m0_be_fmt_log_header {
-	uint64_t    flh_serial;
+	uint64_t    flh_serial_num;
 	m0_bindex_t flh_discarded;
 	m0_bindex_t flh_group_lsn;
 	m0_bcount_t flh_group_size;
@@ -208,7 +208,7 @@ struct m0_be_fmt_log_header {
 
 #define BFLH_F "(flh_serial=%"PRIu64" flh_discarded=%"PRIu64" " \
 		"flh_group_lsn=%"PRIu64" flh_group_size=%"PRIu64")"
-#define BFLH_P(log_hdr) (log_hdr)->flh_serial, (log_hdr)->flh_discarded, \
+#define BFLH_P(log_hdr) (log_hdr)->flh_serial_num, (log_hdr)->flh_discarded, \
 			(log_hdr)->flh_group_lsn, (log_hdr)->flh_group_size
 
 struct m0_be_fmt_group {
@@ -262,6 +262,7 @@ struct m0_be_fmt_log_record_header {
 	m0_bindex_t                            lrh_prev_pos;
 	m0_bindex_t                            lrh_prev_size;
 	uint64_t                               lrh_io_nr_max;
+	uint64_t			       lrh_serial_num;
 	struct m0_be_fmt_log_record_header_io_size lrh_io_size;
 } M0_XCA_RECORD M0_XCA_DOMAIN(be);
 
--- a/be/log.c
+++ b/be/log.c
@@ -108,7 +108,7 @@ static int be_log_level_enter(struct m0_module *module)
 		return m0_be_fmt_log_header_init(&log->lg_header, NULL);
 	case M0_BE_LOG_LEVEL_HEADER:
 		if (log->lg_create_mode) {
-			m0_be_log_header__set(&log->lg_header, 0, 0, 0);
+			m0_be_log_header__set(&log->lg_header, 0, 0, 0, 0);
 			rc = be_log_header_write(log, &log->lg_header);
 		} else {
 			rc = m0_be_log_header_read(log, &log->lg_header);
@@ -663,6 +663,7 @@ m0_be_log_record_io_prepare(struct m0_be_log_record *record,
 	struct m0_bufvec         bvec;
 	struct m0_be_log_io     *lio;
 	struct m0_be_log        *log    = record->lgr_log;
+	struct m0_be_fmt_log_header *lg_header = &log->lg_header;
 	struct m0_buf           *buf;
 	m0_bcount_t              size   = 0;
 	m0_bcount_t              size_lio;
@@ -744,6 +745,7 @@ m0_be_log_record_io_prepare(struct m0_be_log_record *record,
 		header->lrh_discarded = record->lgr_last_discarded;
 		header->lrh_prev_pos  = record->lgr_prev_pos;
 		header->lrh_prev_size = record->lgr_prev_size;
+		header->lrh_serial_num = lg_header->flh_serial_num;
 		size_fmt = m0_be_fmt_log_record_header_size(header);
 		bvec     = M0_BUFVEC_INIT_BUF(&lio->lio_buf.b_addr, &size_fmt);
 		m0_bufvec_cursor_init(&cur, &bvec);
@@ -872,15 +874,17 @@ M0_INTERNAL uint32_t m0_be_log_bshift(struct m0_be_log *log)
 M0_INTERNAL void m0_be_log_header__set(struct m0_be_fmt_log_header *hdr,
 				       m0_bindex_t                  discarded,
 				       m0_bindex_t                  lsn,
-				       m0_bcount_t                  size)
+				       m0_bcount_t                  size,
+				       uint64_t                     serial_num)
 {
-	M0_ENTRY("discarded=%"PRId64" lsn=%"PRId64" size=%"PRId64,
-		 discarded, lsn, size);
+	M0_ENTRY("discarded=%"PRId64" lsn=%"PRId64" size=%"PRId64
+		 " serial_num=%"PRId64, discarded, lsn, size, serial_num);
 
 	m0_be_fmt_log_header_reset(hdr);
 	hdr->flh_discarded  = discarded;
 	hdr->flh_group_lsn  = lsn;
 	hdr->flh_group_size = size;
+	hdr->flh_serial_num = serial_num;
 }
 
 M0_INTERNAL bool m0_be_log_header__is_eq(struct m0_be_fmt_log_header *hdr1,
@@ -888,7 +892,8 @@ M0_INTERNAL bool m0_be_log_header__is_eq(struct m0_be_fmt_log_header *hdr1,
 {
 	return hdr1->flh_discarded  == hdr2->flh_discarded &&
 	       hdr1->flh_group_lsn  == hdr2->flh_group_lsn &&
-	       hdr1->flh_group_size == hdr2->flh_group_size;
+	       hdr1->flh_group_size == hdr2->flh_group_size &&
+	       hdr1->flh_serial_num == hdr2->flh_serial_num;
 }
 
 static void be_log_header_update(struct m0_be_log *log)
@@ -916,7 +921,8 @@ static void be_log_header_update(struct m0_be_log *log)
 		index = log->lg_unplaced_pos;
 		size  = log->lg_unplaced_size;
 	}
-	m0_be_log_header__set(&log->lg_header, log->lg_discarded, index, size);
+	m0_be_log_header__set(&log->lg_header, log->lg_discarded, index,
+			      size, log->lg_header.flh_serial_num);
 }
 
 static int be_log_header_write(struct m0_be_log            *log,
@@ -931,6 +937,17 @@ static int be_log_header_write(struct m0_be_log            *log,
 	rc  = m0_be_fmt_log_header_encode_buf(log_hdr, buf);
 	if (rc == 0)
 		be_log_header_io_sync(log, M0_BE_LOG_STORE_IO_WRITE);
+	return M0_RC(rc);
+}
+
+M0_INTERNAL int be_log_header_inc_generation(struct m0_be_log            *log,
+					     struct m0_be_fmt_log_header *log_hdr)
+{
+	M0_ENTRY("be_log_header_inc_generation=%"PRId64, log_hdr->flh_serial_num);
+	m0_be_log_header__set(log_hdr, log_hdr->flh_discarded, log_hdr->flh_group_lsn,
+			      log_hdr->flh_group_size, log_hdr->flh_serial_num + 1);
+	int rc = be_log_header_write(log, log_hdr);
+	M0_LEAVE("be_log_header_inc_generation=%"PRId64, log_hdr->flh_serial_num);
 	return M0_RC(rc);
 }
 
@@ -949,7 +966,8 @@ M0_INTERNAL bool m0_be_log_header__repair(struct m0_be_fmt_log_header **hdrs,
 	}
 	m0_be_log_header__set(out, hdrs[i]->flh_discarded,
 				   hdrs[i]->flh_group_lsn,
-				   hdrs[i]->flh_group_size);
+				   hdrs[i]->flh_group_size,
+				   hdrs[i]->flh_serial_num);
 
 	return need_repair || !m0_be_log_header__is_eq(hdrs[0], hdrs[2]);
 }
@@ -1065,6 +1083,7 @@ static void be_log_record_header_copy(struct m0_be_fmt_log_record_header *dest,
 	dest->lrh_discarded = src->lrh_discarded;
 	dest->lrh_prev_pos  = src->lrh_prev_pos;
 	dest->lrh_prev_size = src->lrh_prev_size;
+	dest->lrh_serial_num = src->lrh_serial_num;
 	for (i = 0; i < src->lrh_io_size.lrhs_nr; ++i)
 		dest->lrh_io_size.lrhs_size[i] = src->lrh_io_size.lrhs_size[i];
 	dest->lrh_io_size.lrhs_nr = src->lrh_io_size.lrhs_nr;
--- a/be/log.h
+++ b/be/log.h
@@ -465,7 +465,8 @@ M0_INTERNAL uint32_t m0_be_log_bshift(struct m0_be_log *log);
 M0_INTERNAL void m0_be_log_header__set(struct m0_be_fmt_log_header *hdr,
 				       m0_bindex_t                  discarded,
 				       m0_bindex_t                  lsn,
-				       m0_bcount_t                  size);
+				       m0_bcount_t                  size,
+				       uint64_t                     serial_num);
 M0_INTERNAL bool m0_be_log_header__is_eq(struct m0_be_fmt_log_header *hdr1,
 					 struct m0_be_fmt_log_header *hdr2);
 M0_INTERNAL bool m0_be_log_header__repair(struct m0_be_fmt_log_header **hdrs,
@@ -510,6 +511,9 @@ m0_be_log_recovery_discarded(struct m0_be_log *log);
 
 M0_INTERNAL bool m0_be_log_contains_stob(struct m0_be_log        *log,
                                          const struct m0_stob_id *stob_id);
+
+M0_INTERNAL int be_log_header_inc_generation(struct m0_be_log            *log,
+					     struct m0_be_fmt_log_header *log_hdr);
 /** @} end of be group */
 
 #endif /* __MOTR_BE_LOG_H__ */
--- a/be/recovery.c
+++ b/be/recovery.c
@@ -245,16 +245,27 @@ M0_INTERNAL int m0_be_recovery_run(struct m0_be_recovery *rvr)
 		rc = be_recovery_log_record_iter_new(&iter);
 		rc = rc ?: m0_be_log_record_prev(log, prev, iter);
 		if (rc == 0) {
-			log_record_iter_tlink_init_at(iter, &rvr->brec_iters);
-			prev = iter;
-			M0_LOG(M0_DEBUG, "Backward movement : prev header "
-					  BFLRH_F, BFLRH_P(&(prev->lri_header)));
+			if (iter->lri_header.lrh_serial_num == log_hdr.flh_serial_num) {
+				log_record_iter_tlink_init_at(iter, &rvr->brec_iters);
+				prev = iter;
+				M0_LOG(M0_ERROR, "record header=%"PRIu64"log header=%"PRIu64,
+				       prev->lri_header.lrh_serial_num, log_hdr.flh_serial_num);
+				M0_LOG(M0_DEBUG, "Backward movement : prev header "
+						 BFLRH_F, BFLRH_P(&(prev->lri_header)));
+			}
+			else {
+				M0_LOG(M0_ERROR, "log header and record header serial num do not match"
+						 "record header=%"PRIu64"log header=%"PRIu64,
+						 prev->lri_header.lrh_serial_num, log_hdr.flh_serial_num);
+				rc = -EINVAL;
+				break;
+			}
 		} else
 			be_recovery_log_record_iter_destroy(iter);
 	}
 	M0_ASSERT(ergo(rc == 0, prev->lri_header.lrh_pos == last_discarded));
 
-	if (rc != 0){
+	if (rc != 0 && rc != -EINVAL){
 		M0_LOG(M0_ERROR, "Error: rc=%d", rc);
 		goto err;
 	}
--- a/be/ut/log.c
+++ b/be/ut/log.c
@@ -397,7 +397,8 @@ static void be_ut_log_header_repair_test(int header_nr,
 					 m0_bcount_t size_old,
 					 m0_bindex_t lsn_new,
 					 m0_bcount_t size_new,
-					 int valid_index)
+					 int valid_index,
+					 uint64_t serial_num)
 {
 	struct m0_be_fmt_log_header **hdrs;
 	struct m0_be_fmt_log_header   valid = {};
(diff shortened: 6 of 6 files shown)


gshipra:
@sg-shankar @madhavemuri  @atulsdeshmukh2312 - Please review the code.

be/log.c
```diff
@@ -934,6 +940,17 @@ static int be_log_header_write(struct m0_be_log            *log,
 	return M0_RC(rc);
 }
 
+M0_INTERNAL int be_log_header_inc_generation(struct m0_be_log            *log,
```
huanghua78, line 943: Usually, we name a function as:
xxx_noun_verb().

So, in this case, we prefer:
be_log_header_generation_inc()

be/log.c
```diff
@@ -934,6 +940,17 @@ static int be_log_header_write(struct m0_be_log            *log,
...
+					     struct m0_be_fmt_log_header *log_hdr)
+{
+	M0_ENTRY("be_log_header_inc_generation=%"PRId64, log_hdr->flh_serial_num);
+	m0_be_log_header__set(log_hdr, log_hdr->flh_discarded, log_hdr->flh_group_lsn,
+			      log_hdr->flh_group_size, log_hdr->flh_serial_num + 1);
+	int rc = be_log_header_write(log, log_hdr);
+	M0_LEAVE("be_log_header_inc_generation=%"PRId64, log_hdr->flh_serial_num);
```
huanghua78, line 950: Is this function protected by some mutex lock?

If not, concurrent calling of this function will cause race-conditions, and result in wrong serial number.

be/fmt.h
```diff
@@ -200,15 +200,15 @@ struct m0_be_fmt_log_store_header {
 struct m0_be_fmt_group_cfg;
 
 struct m0_be_fmt_log_header {
-	uint64_t    flh_serial;
+	uint64_t    flh_serial_num;
 	m0_bindex_t flh_discarded;
 	m0_bindex_t flh_group_lsn;
```
huanghua78, line 205: Please add comments for these field members.
Especially for "flh_serial_num" and "flh_group_lsn".
What are the differences?

be/recovery.c
```diff
@@ -245,16 +245,27 @@ M0_INTERNAL int m0_be_recovery_run(struct m0_be_recovery *rvr)
...
 		if (rc == 0) {
-			log_record_iter_tlink_init_at(iter, &rvr->brec_iters);
-			prev = iter;
-			M0_LOG(M0_DEBUG, "Backward movement : prev header "
-					  BFLRH_F, BFLRH_P(&(prev->lri_header)));
+			if (iter->lri_header.lrh_serial_num == log_hdr.flh_serial_num) {
+				log_record_iter_tlink_init_at(iter, &rvr->brec_iters);
+				prev = iter;
+				M0_LOG(M0_ERROR, "record header=%"PRIu64"log header=%"PRIu64,
+				       prev->lri_header.lrh_serial_num, log_hdr.flh_serial_num);
```
huanghua78, line 252: I think this line exceeds 80 chars. Please fix it.

And please fix all the same issues.

be/recovery.c
```diff
@@ -245,16 +245,27 @@ M0_INTERNAL int m0_be_recovery_run(struct m0_be_recovery *rvr)
...
+				rc = -EINVAL;
+				break;
+			}
 		} else
 			be_recovery_log_record_iter_destroy(iter);
 	}
 	M0_ASSERT(ergo(rc == 0, prev->lri_header.lrh_pos == last_discarded));
 
-	if (rc != 0){
+	if (rc != 0 && rc != -EINVAL){
```
huanghua78, line 268: If the serial number mismatch, rc will be -EINVAL. That will not be considered as an error, and we will proceed?
If so, the rc should be reset to 0 ?

be/log.c
gshipra: Done.

be/log.c
gshipra: This function is called from m0_be_engine_start with be_engine_lock(en) held.

be/fmt.h
gshipra: done.

be/recovery.c
gshipra: done.

be/recovery.c
gshipra: done.

gshipra:
Adding @nikitadanilov  for code review.

andriytk:
Has the problem been ever reproduced, or it's only a theoretical observation?

be/engine.c
```diff
@@ -814,6 +814,10 @@ M0_INTERNAL int m0_be_engine_start(struct m0_be_engine *en)
 	if (rc != 0)
 		return M0_ERR(rc);
 
+	rc = be_log_header_generation_inc(&(en->eng_log), &(en->eng_log.lg_header));
```
chumakd, line 817: parens are redundant:
```c
rc = be_log_header_generation_inc(&en->eng_log, &en->eng_log.lg_header);
```

chumakd: When uncertain, I always look at `man 7 operator` for C operators precedence reference.

gshipra:
> Has the problem been ever reproduced, or it's only a theoretical observation?

This problem has not been ever reproduced. If some one can suggest me the steps, I can try to reproduce.

andriytk:
Is it possible to add some background information into the issue description above about how the log update works? It will help to understand the problem and the proposed solution.

andriytk:
Please add some background information into the issue description above about how the log update works. It will help to understand the problem and the proposed solution.

gshipra:
> Please add some background information into the issue description above about how the log update works. It will help to understand the problem and the proposed solution.

Yes, I will add it as per my understanding of the code.

andriytk:
From https://github.com/Seagate/cortx-motr/blob/05b719b85d806fa97e83adec4f3fea1245ab17e3/be/log.h#L105-L117 looks like the log header is needed to be updated each time when the log is rotated and tries to re-write the old record the header was pointing to. And if I correctly understand the problem, we update the header with the new record information - the one we are going to write to the log, but which is not written yet. So if the failure happens in-between - the header will point to some garbage in the log segment.

If my understanding is correct, then why don't we just update the log header with the previous log record which is known to be written successfully already?

huanghua78:
If the log header is updated for every record, this will impact performance dramatically?
We will lose the sequential appending to a log, because we need to update the header.

andriytk:
> If the log header is updated for every record, this will impact performance dramatically?

@huanghua78 not for every record, but each time when the log is rotated (see my previous comment). And the rotation of the log file happens not so often.

be/log.c
huanghua78: Please add assertion of lock at the beginning of this function.

huanghua78:
> If the log header is updated for every record, this will impact performance dramatically?

>@huanghua78 not for every record, but each time when the log is rotated (see my previous comment). And the rotation of the log file happens not so often.

OK. Thank you, @andriytk.

huanghua78:
I agree that, we must reproduce the issue (maybe need to write some special test script) and verify that this patch fixes the issue.

andriytk:
> why don't we just update the log header with the previous log record which is known to be written successfully already by the time we update the log header?

Of course, this implies that the previous record is not overwritten with the new one, i.e. that the log file is big enough to accommodate at least two records in it. Which is usually the case.

gshipra:
> I agree that, we must reproduce the issue (maybe need to write some special test script) and verify that this patch fixes the issue.

@andriytk @huanghua78 - If some one can guide me how to reproduce this defect, I can try reproducing it.

madhavemuri:
> > I agree that, we must reproduce the issue (maybe need to write some special test script) and verify that this patch fixes the issue.
> 
> @andriytk @huanghua78 - If some one can guide me how to reproduce this defect, I can try reproducing it.

@gshipra : Can you try with  fault injection mechanism in motr to reproduce the issue ?

gshipra:
> > > I agree that, we must reproduce the issue (maybe need to write some special test script) and verify that this patch fixes the issue.
> > 
> > 
> > @andriytk @huanghua78 - If some one can guide me how to reproduce this defect, I can try reproducing it.
> 
> @gshipra : Can you try with fault injection mechanism in motr to reproduce the issue ?

I tried to reproduce the problem by using fault injection and via filling the logs to make it write log header

M0_INTERNAL void m0_be_log_record_io_launch(struct m0_be_log_record *record,
                                            struct m0_be_op         *op)
{
if (record->lgr_write_header) {
                m0_be_op_reset(&log->lg_header_write_op);
                m0_be_op_set_add(&record->lgr_record_op,
                                 &log->lg_header_write_op);
                be_log_header_io(log, M0_BE_LOG_STORE_IO_WRITE,
                                 &log->lg_header_write_op);

		// do not write reorc, return from here.
                if (M0_FI_ENABLED("skip_record"))
                {
                        m0_be_log_sched_unlock(&log->lg_sched);
                        return;
                }
        }

}

void m0_be_ut_recovery(void)
{
be_ut_recovery_log_fill(&ctx, 3, 3, false); //write 3 record and discard 3
be_ut_recovery_log_fill(&ctx, 244, 0, false); // write 244 and discard nothing
m0_fi_enable("m0_be_log_record_io_launch", "skip_record"); // skip writing record, write only log header
// write 3 records, log wrap would happen at 1st record and we will only write log header, writing record will be skipped for 1st record
be_ut_recovery_log_fill(&ctx, 3, 0, false); 
m0_fi_disable("m0_be_log_record_io_launch", "skip_record");

be_ut_recovery_log_reopen(&ctx);
// recovery recovered only 244 records, last 3 records are not recovered. As we missed writing the record for one instance, records linked list is screwed.
count = be_ut_recovery_iter_count(&ctx); 
M0_UT_ASSERT(count == 244);
}

gshipra:
I have written some test code for reproducing the problem.
It has two parts. 
1. Fault injection code to write only log header and not record
2. UT code to induce log wrap and cause fault injection.

Test code for fault injection at the time of log header

M0_INTERNAL void m0_be_log_record_io_launch(struct m0_be_log_record *record,
                                            struct m0_be_op         *op)
{
if (record->lgr_write_header) {
                m0_be_op_reset(&log->lg_header_write_op);
                m0_be_op_set_add(&record->lgr_record_op,
                                 &log->lg_header_write_op);
                be_log_header_io(log, M0_BE_LOG_STORE_IO_WRITE,
                                 &log->lg_header_write_op);

		// do not write reorc, return from here.
                if (M0_FI_ENABLED("skip_record"))
                {
                        m0_be_log_sched_unlock(&log->lg_sched);
                        return;
                }
        }

}


Test code for forcing log wrap and failing recod writing after writing log header:
m0_be_ut_recovery()
{
**//write 3 records and discard 3 records**
be_ut_recovery_log_fill(&ctx, 3, 3, false);
// write 244 records and discard nothing
be_ut_recovery_log_fill(&ctx, 244, 0, false);
//log wrap will happen at next record write,enable fault injection
to write only log header and not record
m0_fi_enable("m0_be_log_record_io_launch", "skip_record");
//log wrap happens at this record
be_ut_recovery_log_fill(&ctx, 1, 0, false);
m0_fi_disable("m0_be_log_record_io_launch", "skip_record");

// close and open log 
be_ut_recovery_log_reopen(&ctx);
count = be_ut_recovery_iter_count(&ctx);
// recovery recovers only 244 records
M0_UT_ASSERT(count == 244);

// write 5 more records
be_ut_recovery_log_fill(&ctx, 5, 0, false);
be_ut_recovery_log_reopen(&ctx);
count = be_ut_recovery_iter_count(&ctx);
// recovery recocers 5 records
M0_UT_ASSERT(count == 5);
}


WHAT IS INDEX AND SIZE AT TIME OF LOG WRAP

(gdb) bt
#0  be_log_header_update (log=0x7fffffffb440) at be/log.c:908
#1  0x00007fffebd93386 in m0_be_log_record_io_prepare (record=0x8b0740, opcode=SIO_WRITE, size_reserved=4096) at be/log.c:772
#2  0x00007fffed5ad98d in be_ut_recovery_log_fill (ctx=0x7fffffffb440, record_nr=1, discard_nr=0, last_incomplete=false) at be/ut/recovery.c:169
#3  0x00007fffed5ade10 in m0_be_ut_recovery () at be/ut/recovery.c:257
#4  0x00007fffed76ca8f in run_test (test=0x6832c8 <be_ut+4488>, max_name_len=43) at ut/ut.c:390
#5  0x00007fffed76cd81 in run_suite (suite=0x682140 <be_ut>, max_name_len=43) at ut/ut.c:459
#6  0x00007fffed76d025 in tests_run_all (m=0x7ffff7dccae0 <ut.11897>) at ut/ut.c:513
#7  0x00007fffed76d083 in m0_ut_run () at ut/ut.c:539
#8  0x00000000004048e4 in main (argc=3, argv=0x7fffffffe5a8) at ut/m0ut.c:508
(gdb) n
909             m0_bcount_t              size  = log->lg_prev_record_size;
(gdb)
911             m0_mutex_lock(&log->lg_record_state_lock);
(gdb)
912             record = record_tlist_head(&log->lg_records);
(gdb)
914             if (record != NULL && record->lgr_state != LGR_DONE &&
(gdb)
915                 record->lgr_position != 0) {
(gdb)
914             if (record != NULL && record->lgr_state != LGR_DONE &&
(gdb)
916                     index = record->lgr_prev_pos;
(gdb)
917                     size  = record->lgr_prev_size;
(gdb)
919             while (record != NULL && record->lgr_state == LGR_DONE) {
(gdb)
924             m0_mutex_unlock(&log->lg_record_state_lock);
(gdb)
926             if (log->lg_unplaced_exists && log->lg_unplaced_pos < index) {
(gdb)
927                     index = log->lg_unplaced_pos;
(gdb)
928                     size  = log->lg_unplaced_size;
(gdb)
931             M0_LOG(M0_ERROR, " size=%"PRIu64" index=%"PRIu64, size, index);
(gdb) p index
$1 = 12288
(gdb) p size
$2 = 4096

[root@ssc-vm-2585 cortx-motr]# grep "be_log_header_update" out.txt
3300  8722115.971.152874  7fffffffaf80  ERROR    be/log.c:931:be_log_header_update                   size=4096 index=0 -- at start
33355  8722116.240.633430  7fffffffaf80  ERROR   be/log.c:931:be_log_header_update                   size=4096 index=12288 -- at log wrap
33792  8722116.246.833912  7fffffffb290  ERROR   be/log.c:931:be_log_header_update                   size=4096 index=12288 -- at log close
85411  8722116.771.139240  7fffffffaf80  ERROR   be/log.c:931:be_log_header_update                   size=4096 index=1011712 -- after first recovery of 244 records
86073  8722116.778.281510  7fffffffb290  ERROR   be/log.c:931:be_log_header_update                   size=4096 index=1011712 -- at log close



FIRST RECOVERY RUN WITH 244 RECORDS:

35190  8722116.307.145716  7fffffffafd0  CALL    be/recovery.c:195:m0_be_recovery_run                > rvr = 0x7fffffffd8c0, log = 0x7fffffffb440
35602  8722116.310.039220  7fffffffafd0  DEBUG   be/recovery.c:203:m0_be_recovery_run                log_discarded=12288
35712  8722116.311.111494  7fffffffafd0  DEBUG   be/recovery.c:211:m0_be_recovery_run                Forward movement : prev header (pos=12288 size=4096 discarded=12288 prev_pos=8192 prev_size=4096 io_nr_max=2)
62442  8722116.603.826892  7fffffffafd0  DEBUG   be/recovery.c:211:m0_be_recovery_run                Forward movement : prev header (pos=1007616 size=4096 discarded=12288 prev_pos=1003520 prev_size=4096 io_nr_max=2)
62554  8722116.605.145888  7fffffffafd0  DEBUG   be/recovery.c:287:m0_be_recovery_run                Recovery scan complete : last_record_pos=1007616 last_record_size=4096 current position=1011712 discarded position=12288

**Position of the record which could not be written because of fault injection is : 47 * 4096 = 1011712
For this round of recovery last record pos is - last_record_pos=100761
First 3 Records were discarded thus discarded position is 12288**

SECOND RECOVERY RUN WITH 5 RECORDS

87483  8722116.835.144360  7fffffffafd0  CALL    be/recovery.c:195:m0_be_recovery_run                > rvr = 0x7fffffffd8c0, log = 0x7fffffffb440
87895  8722116.838.195792  7fffffffafd0  DEBUG   be/recovery.c:203:m0_be_recovery_run                log_discarded=1011712
88005  8722116.839.071324  7fffffffafd0  DEBUG   be/recovery.c:211:m0_be_recovery_run                Forward movement : prev header (pos=1011712 size=4096 discarded=1011712 prev_pos=1007616 prev_size=4096 io_nr_max=2)
88115  8722116.839.951020  7fffffffafd0  DEBUG   be/recovery.c:211:m0_be_recovery_run                Forward movement : prev header (pos=1015808 size=4096 discarded=1011712 prev_pos=1011712 prev_size=4096 io_nr_max=2)
88225  8722116.840.790792  7fffffffafd0  DEBUG   be/recovery.c:211:m0_be_recovery_run                Forward movement : prev header (pos=1019904 size=4096 discarded=1011712 prev_pos=1015808 prev_size=4096 io_nr_max=2)
88335  8722116.841.642790  7fffffffafd0  DEBUG   be/recovery.c:211:m0_be_recovery_run                Forward movement : prev header (pos=1024000 size=4096 discarded=1011712 prev_pos=1019904 prev_size=4096 io_nr_max=2)
88445  8722116.842.540518  7fffffffafd0  DEBUG   be/recovery.c:211:m0_be_recovery_run                Forward movement : prev header (pos=1028096 size=4096 discarded=1011712 prev_pos=1024000 prev_size=4096 io_nr_max=2)
88557  8722116.843.567354  7fffffffafd0  DEBUG   be/recovery.c:287:m0_be_recovery_run                Recovery scan complete : last_record_pos=1028096 last_record_size=4096 current position=1032192 discarded position=1011712
88558  8722116.843.567900  7fffffffafd0  CALL    be/recovery.c:316:m0_be_recovery_run                <
88585  8722116.843.635116  7fffffffad90  DEBUG   be/recovery.c:348:m0_be_recovery_log_record_get     iter pos=1011712 size=4096 discarded=1011712
88679  8722116.844.198044  7fffffffad90  DEBUG   be/recovery.c:348:m0_be_recovery_log_record_get     iter pos=1015808 size=4096 discarded=1011712
88771  8722116.844.816850  7fffffffad90  DEBUG   be/recovery.c:348:m0_be_recovery_log_record_get     iter pos=1019904 size=4096 discarded=1011712
88863  8722116.845.324796  7fffffffad90  DEBUG   be/recovery.c:348:m0_be_recovery_log_record_get     iter pos=1024000 size=4096 discarded=1011712
88955  8722116.845.840874  7fffffffad90  DEBUG   be/recovery.c:348:m0_be_recovery_log_record_get     iter pos=1028096 size=4096 discarded=1011712

**[eos-14402.docx](https://github.com/Seagate/cortx-motr/files/6337434/eos-14402.docx)**

From Traces this can be concluded that failure to write log record after writing log header
does not do any harm.

From Code perspective:

When we do log wrap, we update log header with the first undone record or first unplaced record, which ever index is least.
Log header does not have pointer to last or current record to be written. Rather it maintains the pointer to first position
which is unplaced/undone.

At the time of recovery we check read the log header and get the position of first record.
Next we read each and every record till the end and make sure that next record has valid position.

I believe, no fix is needed.

andriytk:
> Rather it maintains the pointer to first position which is unplaced/undone.

OK, but what if the record at this position will be corrupted - how the recovery will find the next record after it?

And it can be corrupted there because we update the log header with the new position before writing the record at this position. Don't we?

gshipra:
> > Rather it maintains the pointer to first position which is unplaced/undone.
> 
> OK, but what if the record at this position will be corrupted - how the recovery will find the next record after it?
> 
> And it can be corrupted there because we update the log header with the new position before writing the record at this position. Don't we?



> **And it can be corrupted there because we update the log header with the new position before writing the record at this position. Don't we?**

We update log header at the time of wrap with the first undone or unplaced position. This is **not** the same position as where we will be writing new record. If you check the traces for the wrap situation - log header is updated with 12288, first undone/non discarded/unplaced record. Whereas the new record location ( record which we will fail writing) is 247*4096 = 1011712.

My test code test the situation -
Log wrap happens.
update log header
write record fails because of power cycle (we do it via fault injection)
system started again and recovery happens (we do it via log close and open)
recovery does not take the record at 1011712 in consideration.

andriytk:
> We update log header at the time of wrap with the first undone or unplaced position. This is not the same position as where we will be writing new record.

Ah, OK. So if it's some position which has been written previously already - then we are safe indeed. That's actually how I proposed to fix the issue, BTW (if it was existing).

Now, @max-seagate was mentioning elsewhere that there might be another issue when the log file is too small and the log record is too big, so that we may not find that "first undone or unplaced position" in an intact state. In other words, this is a rigid situation (which should not happen in practice) when only one log record can be accommodated in the log. Because currently, there is no limit to the log record, so it's still possible that due to some misconfiguration such a situation can happen.

So to avoid such a situation and not allow it to happen we should make sure that when we reserve the space in the log we don't overwrite the last successfully written log record with our new record. In other words, to make sure that there are always at least two valid log records in the log available at any moment in time.

Please check and investigate this and create another PR, if needed.

Thank you.

gshipra:
> > We update log header at the time of wrap with the first undone or unplaced position. This is not the same position as where we will be writing new record.
> 
> Ah, OK. So if it's some position which has been written previously successfully already - then we are safe indeed. That's actually how I proposed to fix the issue above, BTW (if it was existing).
> 
> Now, @max-seagate was mentioning elsewhere that there might be another issue when the log file is too small and the log record is too big, so that we may not find that "first undone or unplaced position" in an intact state. In other words, this is a rigid situation (which should not happen in practice) when only one log record can be accommodated in the log. Because currently, there is no limit to the log record, so it's still theoretically possible that due to some misconfiguration such a situation would happen.
> 
> So to not allow this to happen we should make sure that when we reserve the space in the log for the next record we don't overwrite the last successfully written log record with it. In other words, to make sure that there are always at least two valid log records in the log available at any moment in time.
> 
> Please investigate this and create another PR, if needed.
> 
> Thank you.

@andriytk @max-seagate  - Is there an existing defect for 'at least two records space' ? If not than I will create one and investigate further.

be/engine.c
huanghua78: Can this function return failure?
If yes, do we need to cleanup before return err?

be/fmt.h
```diff
@@ -200,15 +200,18 @@ struct m0_be_fmt_log_store_header {
 struct m0_be_fmt_group_cfg;
 
 struct m0_be_fmt_log_header {
-	uint64_t    flh_serial;
+	/* flh_serial_num is a incarnation number in log header.
+	 * This number is incremented after every engine restart
+	 */
```
huanghua78, line 205:
```
/**
 * some comments.
 */
```
can be a correct Doxygen comments.

be/fmt.h
```diff
@@ -262,6 +265,11 @@ struct m0_be_fmt_log_record_header {
 	m0_bindex_t                            lrh_prev_pos;
 	m0_bindex_t                            lrh_prev_size;
 	uint64_t                               lrh_io_nr_max;
+	/* lrh_serial_num is populated with flh_serial_num of log header.
+	 * These two fields are compared at the time of recovery
+	 * to check the validity of record.
+	 */
+	uint64_t			       lrh_serial_num;
```
huanghua78, line 272:
```
/**
 * This is a Doxygen comment.
 */
```
is t he correct format.

Please add explanations, what are the correct comparison result?
In what cases, we say a record is valid?

be/log.c
```diff
@@ -663,6 +663,7 @@ m0_be_log_record_io_prepare(struct m0_be_log_record *record,
 	struct m0_bufvec         bvec;
 	struct m0_be_log_io     *lio;
 	struct m0_be_log        *log    = record->lgr_log;
+	struct m0_be_fmt_log_header *lg_header = &log->lg_header;
```
huanghua78, line 666: This variable breaks the alignment of other variable names.

be/log.c
```diff
@@ -916,7 +921,8 @@ static void be_log_header_update(struct m0_be_log *log)
 		index = log->lg_unplaced_pos;
 		size  = log->lg_unplaced_size;
 	}
-	m0_be_log_header__set(&log->lg_header, log->lg_discarded, index, size);
+	m0_be_log_header__set(&log->lg_header, log->lg_discarded, index,
+			      size, log->lg_header.flh_serial_num);
```
huanghua78, line 925: Are you sure, the "log->lg_header.flh_serial_num" is valid at this moment?

This is confusing. "log->lg_header" is updated with the serial number from itself.

be/log.c
huanghua78: Why do we need the second parameter?
It is embedded as "struct m0_be_log::lg_header".
So, we can get it from the first param.

huanghua78: So, this function only increases the serial number, but also synchronized it onto persistent storage.
This needs to be added into comments.

huanghua78: Also we don't call it a "generation" number, but a serial number.

be/log.h
```diff
@@ -510,6 +511,9 @@ m0_be_log_recovery_discarded(struct m0_be_log *log);
 
 M0_INTERNAL bool m0_be_log_contains_stob(struct m0_be_log        *log,
                                          const struct m0_stob_id *stob_id);
+
+M0_INTERNAL int be_log_header_generation_inc(struct m0_be_log            *log,
+					     struct m0_be_fmt_log_header *log_hdr);
```
huanghua78, line 516: Please add Doxygen comments for these functions (this and others in this file).
We need all functions to have comments.

be/recovery.c
```diff
@@ -245,16 +245,25 @@ M0_INTERNAL int m0_be_recovery_run(struct m0_be_recovery *rvr)
...
+				prev = iter;
+				M0_LOG(M0_DEBUG, "Backward movement : prev header "
+						 BFLRH_F, BFLRH_P(&(prev->lri_header)));
+			}
+			else {
+				M0_LOG(M0_ERROR, "serial num do not match"
+						 "record header=%"PRIu64
+						 "log header=%"PRIu64,
+						 prev->lri_header.lrh_serial_num,
+						 log_hdr.flh_serial_num);
```
huanghua78, line 259: What actions shall we take in this situation?

I only see some error messages are generated, without any other actions.

gshipra:
@huanghua78 plase check defect page for analysis and my previous comments in this pull request. This is not a defect and we do not need any fix. This PR is out of my forked repo. I will close it.

huanghua78:
> @huanghua78 plase check defect page for analysis and my previous comments in this pull request. This is not a defect and we do not need any fix. This PR is out of my forked repo. I will close it.


OK. Thank you.

PR #566, opened 2021-04-07
utils: enable kernel module tracing

andriytk:
Previously, it was disabled due to licensing issues in
m0ctl.ko which uses GPL-exported kernel function(s).
But GPLv3-licensed projects do allow to use Apache 2.0 code
in them. (See more at
https://www.apache.org/licenses/GPL-compatibility.html)
So we can make m0ctl.ko module GPLv3 and still keep all the
rest code under Apache 2.0. Including m0tr.ko, which does
not use any code from m0ctl.ko.

Solution: make m0ctl.ko GPLv3 and enable its code which
allows to use m0trace with m0tr.ko.

### DIFF
--- a/utils/linux_kernel/core_debugfs.c
+++ b/utils/linux_kernel/core_debugfs.c
@@ -19,8 +19,6 @@
  *
  */
 
-#if 0
-
 #include <linux/kernel.h>    /* pr_info */
 #include <linux/debugfs.h>   /* debugfs_create_dir */
 #include <linux/module.h>    /* THIS_MODULE */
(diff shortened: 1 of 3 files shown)


andriytk:
cc @johnbent

huanghua78:
This change must be discussed with @johnbent and approved by @johnbent.

andriytk:
Linux kernel module cannot be licensed under GPLv3 (and GPLv2 is not compatible with Apache 2.0 licence), so I'm closing this.

PR #567, opened 2021-04-07
EOS-19208: remove tx and segment dependency from btree code for benchmark.

NikhilBirgade:

### DIFF
--- a/motr/m0crate/btree_without_tx_seg.c
+++ b/motr/m0crate/btree_without_tx_seg.c
@@ -63,6 +63,10 @@ static struct m0_rwlock *btree_rwlock(struct m0_be_btree *tree);
 
 static struct be_btree_key_val *be_btree_search(struct m0_be_btree *btree,
 						void *key);
+
+#define M0_BE_TX_CAPTURE_PTR(seg, tx, ptr)
+#define M0_BE_CREDIT_DEC(cr_user, tx)
+
 M0_INTERNAL void m0_be_alloc_aligned(struct m0_be_allocator *a,
 				     struct m0_be_tx *tx,
 				     struct m0_be_op *op,
@@ -102,6 +106,11 @@ M0_INTERNAL bool m0_be_seg_contains(const struct m0_be_seg *seg,
 				    const void *addr)
 {
 	return true;
+}
+
+M0_INTERNAL void m0_be_tx_capture(struct m0_be_tx *tx, 
+				  const struct m0_be_reg *reg)
+{
 }
 
 static void btree_root_set(struct m0_be_btree *btree,
@@ -149,6 +158,7 @@ static inline void mem_free(const struct m0_be_btree *btree,
 static inline void mem_update(const struct m0_be_btree *btree,
 			      struct m0_be_tx *tx, void *ptr, m0_bcount_t size)
 {
+	m0_be_tx_capture(tx, &M0_BE_REG(btree->bb_seg, size, ptr));
 }
 
 static inline void *mem_alloc(const struct m0_be_btree *btree,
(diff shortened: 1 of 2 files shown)


huanghua78:
This is only used for benchmarking purpose?
Not intended to be merged?

NikhilBirgade:
> This is only used for benchmarking purpose?
> Not intended to be merged?

The plan is to integrate btree benchmarking tool with m0crate. It need to be merged in that case.

huanghua78:
My question here.
The btree_without_tx_seg.c is copied from btree.c and modified to have no tx+seg?
Can we save it as a patch, and generate it when building?

Because this will not duplicate the original code. Otherwise, we have to make sure to update this source code when original btree.c is updated.

huanghua78 (approved):
Looks fine.

motr/m0crate/btree_without_tx_seg.c
```diff
@@ -578,11 +480,6 @@ static void be_btree_insert_newkey(struct m0_be_btree      *btree,
 	struct m0_be_bnode *old_root;
 	struct m0_be_bnode *new_root;
 
-	M0_PRE(btree_invariant(btree));
```
shashank-parulekar, line 581: you should have kept the invariants if they are not related to segment or tx operations.

shashank-parulekar (changes requested):

motr/m0crate/btree_without_tx_seg.c
NikhilBirgade: Updated code to include invariants

shashank-parulekar (approved):

motr/m0crate/btree_without_tx_seg.c
```diff
@@ -247,8 +270,7 @@ static bool btree_backlink_invariant(const struct m0_be_btree_backlink *h,
 		_0C(m0_cookie_dereference(&h->bli_tree, &cookie) == 0) &&
 		_0C(m0_be_seg_contains(seg, cookie)) &&
 		_0C(parent == cookie) &&
-		_0C(M0_BBT_INVALID < h->bli_type && h->bli_type < M0_BBT_NR) &&
-		_0C(h->bli_gen == seg->bs_gen);
+		_0C(M0_BBT_INVALID < h->bli_type && h->bli_type < M0_BBT_NR);
```
nikitadanilov, line 273: Would it be difficult to create a dummy m0_be_seg structure, and point every btree node to it? I'd like to keep code as much in sync with the original, as possible.

motr/m0crate/btree_without_tx_seg.c
```diff
@@ -446,7 +468,6 @@ static void btree_node_free(struct m0_be_bnode       *node,
 {
 	/* invalidate header so that recovery tool can skip freed records */
 	node->bt_header.hd_magic = 0;
-	M0_BE_TX_CAPTURE_PTR(btree->bb_seg, tx, &node->bt_header.hd_magic);
```
nikitadanilov, line 449: Instead #define M0_BE_TX_CAPTURE_PTR() to a no-op.

motr/m0crate/btree_without_tx_seg.c
```diff
@@ -1382,19 +1403,6 @@ static void btree_save(struct m0_be_btree        *tree,
...
-			M0_BE_CREDIT_DEC(M0_BE_CU_BTREE_DELETE, tx);
-			/* fallthrough */
-		case BTREE_SAVE_INSERT:
-			M0_BE_CREDIT_DEC(M0_BE_CU_BTREE_INSERT, tx);
-			break;
-		case BTREE_SAVE_UPDATE:
-			M0_BE_CREDIT_DEC(M0_BE_CU_BTREE_UPDATE, tx);
-			break;
-	}
-
```
nikitadanilov, line 1397: Again, use #defines.

nikitadanilov (changes requested):

motr/m0crate/btree_without_tx_seg.c
NikhilBirgade: Added #define as noop.

motr/m0crate/btree_without_tx_seg.c
NikhilBirgade: Added #define as noop.

motr/m0crate/btree_without_tx_seg.c
NikhilBirgade: Re added code for seg. It will be responsibility of btree interface user to pass a dummy m0_be_seg while creating tree. 
For eg. m0crate will add dummy seg to btree while tree creation and same dummy seg will be used for further node creation.

shashank-parulekar (approved):

PR #570, opened 2021-04-09
Timeout error fix

yeshpal-jain-seagate:
Sometimes on dell server, it was seen s3servers getting timedout.
Reverting BELOG size which has reduced performance.


huanghua78:
@yeshpal-jain-seagate have you tested and verified this patch on target platform?

madhavemuri:
> @yeshpal-jain-seagate have you tested and verified this patch on target platform?

Yes, these changes are tested on R2 setup.

PR #575, opened 2021-04-13
[btree]Fixed new btree code and added a couple of ut's - EOS 19061

shashank-parulekar:
Fixed the implementation of fixed-format code for new btree. Also added a couple of ut's to debug the btree node code implementation.


btree/btree.c
```diff
@@ -826,46 +915,67 @@ struct node_op {
  * are set by the caller, and the function sets ->s_rec.
  */
 struct slot {
-	struct nd          *s_node;
-	int                 s_idx;
-	struct m0_btree_rec s_rec;
+	const struct nd     *s_node;
```
nikitadanilov, line 918: Field name alignment, see coding-style.

btree/btree.c
```diff
@@ -1015,34 +1135,51 @@ static bool node_shift_is_valid(int shift)
 	return shift >= NODE_SHIFT_MIN && shift < NODE_SHIFT_MIN + 0x10;
 }
 
-static bool segaddr_is_valid(const struct segaddr *addr)
+/**
+ * This function should be called right after the allocation to make sure that
+ * the allocated memory starts at a properly aligned address.
+ *
+ * @author 530902 (06-Apr-21)
```
nikitadanilov, line 1142: This should be something more human readable. :-) But in general, authorship is recorded in the git logs, there is no need to duplicate it in the source.

btree/btree.c
```diff
@@ -1015,34 +1135,51 @@ static bool node_shift_is_valid(int shift)
...
+ *
+ * @param addr - Start address of the allocated space provided by the allocator.
+ *
+ * @return bool - True if the input address is properly aligned.
+ */
+static bool addr_is_aligned(const void *addr)
+{
+	return (((size_t)addr & ((1 << NODE_SHIFT_MIN) - 1)) == 0);
```
nikitadanilov, line 1150: s/1/1ULL/

btree/btree.c
```diff
@@ -1015,34 +1135,51 @@ static bool node_shift_is_valid(int shift)
...
-	return (void *)(addr->as_core >> 9);
+	M0_PRE(segaddr_is_valid(seg_addr));
+	return (void *)(seg_addr->as_core & ~((1 << NODE_SHIFT_MIN) - 1));
 }
 
 static int segaddr_shift(const struct segaddr *addr)
 {
 	M0_PRE(segaddr_is_valid(addr));
-	return addr->as_core & 0xf;
+	return ((addr->as_core & 0xf) + NODE_SHIFT_MIN);
```
nikitadanilov, line 1179: Too many parentheses.

btree/btree.c
```diff
@@ -1139,54 +1298,84 @@ static struct nd *node_try(struct td *tree, struct segaddr *addr)
...
+#endif
 
 static int64_t node_alloc(struct node_op *op, struct td *tree, int size,
-			  struct node_type *nt, struct m0_be_tx *tx, int nxt)
+			  const struct node_type *nt, int ksize, int vsize,
+			  struct m0_be_tx *tx, int nxt)
 {
-	return segops->so_node_alloc(op, tree, size, nt, tx, nxt);
+	int  nxt_state;
+	/**
```
nikitadanilov, line 1308: Doxygen does not work within function bodies.

btree/btree.c
```diff
@@ -1262,17 +1451,76 @@ static const struct seg_ops mem_seg_ops = {
...
 struct ff_head {
-	struct m0_format_header ff_fmt;
-	struct node_header      ff_seg;
-	uint16_t                ff_used;
-	uint8_t                 ff_shift;
-	uint8_t                 ff_level;
-	uint16_t                ff_ksize;
-	uint16_t                ff_vsize;
-	struct m0_format_footer ff_foot;
+	struct m0_format_header ff_fmt;   /* Node Header */
```
nikitadanilov, line 1458: Please use doxygen ("/*<" - style, if you want everything on the same line).

btree/btree.c
```diff
@@ -1262,17 +1451,76 @@ static const struct seg_ops mem_seg_ops = {
...
+	struct node_header      ff_seg;   /* Node type */
+	uint16_t                ff_used;  /* Count of records */
+	uint8_t                 ff_shift; /* node size as pow-of-2 */
+	uint8_t                 ff_level; /* Level in Btree */
+	uint16_t                ff_ksize; /* Size of key in bytes */
+	uint16_t                ff_vsize; /* Size of value in bytes */
+	struct m0_format_footer ff_foot;  /* Node Footer */
+	/**
```
nikitadanilov, line 1466: Doxygen won't attach this anywhere. It would be better to put this information in comment right before ff_head.

btree/btree.c
```diff
@@ -1283,7 +1531,8 @@ static void *ff_key(const struct nd *node, int idx)
 	struct ff_head *h    = ff_data(node);
 	void           *area = h + 1;
 
-	M0_PRE(0 <= idx && idx < h->ff_used);
+	M0_PRE(ergo(!(h->ff_used == 0 && idx == 0),
+		   (0 <= idx && idx <= h->ff_used)));
```
nikitadanilov, line 1535: It's better to introduce ff_invariant(). Add ->nt_invariant() to node_type, call it from node_invariant() and call node_invariant() on entry to every node_*() function.

btree/btree.c
```diff
@@ -1312,9 +1562,27 @@ static bool ff_invariant(const struct nd *node)
...
+}
+
+static void ff_fini(const struct nd *node)
+{
+}
+
 static int ff_count(const struct nd *node)
 {
-	int used = ff_data(node)->ff_used;
+	struct ff_head *h   = ff_data(node);
```
nikitadanilov, line 1584: Why do you need a separate "h"?

btree/btree.c
```diff
@@ -1390,7 +1659,7 @@ static bool ff_isfit(struct slot *slot)
 
 	M0_PRE(ff_invariant(slot->s_node));
 	M0_PRE(ff_rec_is_valid(slot));
-	return h->ff_ksize + h->ff_vsize >= ff_space(slot->s_node);
+	return h->ff_ksize + h->ff_vsize <= ff_space(slot->s_node);
```
nikitadanilov, line 1662: Ah, good catch!

btree/btree.c
```diff
@@ -1421,22 +1690,32 @@ static void ff_find(struct slot *slot, const struct m0_btree_key *key)
...
+	if (h->ff_used == 0 ||
+	    (memcmp(ff_key(slot->s_node, 0),
+		    key->k_data.ov_buf[0], h->ff_ksize) >= 0)) {
+		slot->s_idx = 0;
+		return;
+	} else if (memcmp(ff_key(slot->s_node, (h->ff_used - 1)),
+			  key->k_data.ov_buf[0], h->ff_ksize) < 0) {
+		slot->s_idx = h->ff_used;
+		return;
+	}
```
nikitadanilov, line 1702: Hmm, why? This just introduced 2 additional memcmp() in hot path.

btree/btree.c
```diff
@@ -1421,22 +1690,32 @@ static void ff_find(struct slot *slot, const struct m0_btree_key *key)
...
+
+	do {
 		int m    = (i + j) / 2;
 		int diff = memcmp(ff_key(slot->s_node, m),
 				  key->k_data.ov_buf[0], h->ff_ksize);
-		M0_ASSERT(i < m && m < j);
```
nikitadanilov, line 1428: What's wrong with this? We definitely want to shrink the interval on each iteration.

btree/btree.c
```diff
@@ -1476,34 +1755,400 @@ static void generic_move(struct nd *src, struct nd *tgt,
 		if (nr == 0 || (nr == NR_EVEN &&
 				(node_space(tgt) <= node_space(src))))
 			break;
+
+		/** Get the record at src index in rec. */
 		rec.s_node = src;
 		rec.s_idx  = srcidx;
 		node_rec(&rec);
+
+		/**
```
nikitadanilov, line 1764: No doxygen within function bodies.

nikitadanilov (changes requested):

btree/btree.c
```diff
@@ -546,6 +559,13 @@ enum op_flags {
 	OF_COOKIE  = M0_BITS(3)
 };
 
+enum btree_type {
```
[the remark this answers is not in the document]
nikitadanilov: Tree type is something like "ad-extmap", "cas", "balloc", etc. Tree type has a function that returns node format of a new tree node (when it is allocated).

btree/btree.c
```diff
@@ -714,7 +749,7 @@ struct td {
 	 * read-only after the tree root is loaded into memory.
 	 */
 	struct m0_rwlock            t_lock;
-	struct segaddr              t_root;
+	struct nd                  *t_root;
```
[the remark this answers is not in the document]
nikitadanilov: Two points: proper node_get() function also needs code to page fault the node asynchronously (from another thread). That requires some additional discussion. I think that for the time being we should continue with the mock seg_ops and concentrate on the state machines.

Also, I was wrong about the root node: it is never relocated (because it can be embedded in a larger ambient structure, for one thing), so keeping a pointer to its node descriptor is fine, provided m0_btree_open() pins the root node descriptor in memory and m0_btree_close() releases it. But this moves the complexity to m0_btree_open(): it now has to be a non-blocking state machine that loads the root node from the memory.

btree/btree.c
nikitadanilov: *Identifiers* should be aligned, that is, all "s_"-s should start at the same column.

btree/btree.c
nikitadanilov: Parentheses around a top-level expression are always redundant.

btree/btree.c
nikitadanilov: Yes, the point was that comments within a function body should start with "/* " rather than "/** ".

btree/btree.c
nikitadanilov: Hmm.. which cases the old code does not handle? If used is 0, both versions return 0. If the target key is less than the leftmost key, both versions return 0.

btree/btree.c
nikitadanilov: If used == 2, then i = 0, j = 2 on entry, the loop is entered and m = (0+2)/2 = 1.

I guess we understand the required behaviour of ->nt_find() differently. Note that in addition to insertions it should also work for lookups. Let's leave it as it was for now, work out state machine details and then we will know the exact requirements for t->nt_find().

btree/btree.c
```diff
@@ -735,35 +770,97 @@ enum {
...
+	 */
 	bool (*nt_isfit)(struct slot *slot);
+
+	/**
+	 *  Node changes related to last record have completed; any post
+	 *  processing related to the record needs to be done in this function.
+	 */
 	void (*nt_done) (struct slot *slot, struct m0_be_tx *tx, bool modified);
+
+	/** Make space in the node for inserting new entry at specific index */
```
nikitadanilov, line 836: Please keep the grammatical structure consistent: 3rd person singular present tense indicative mood active voice: "makes", "changes", etc.

nikitadanilov (changes requested):

btree/btree.c
```diff
@@ -896,87 +1018,126 @@ struct m0_btree_oimpl {
 	struct level    i_level[0];
 };
 
+static struct td        trees[M0_TREE_COUNT];
```
huanghua78, line 1021: Why do we only support such amount of trees?

btree/btree.c
```diff
@@ -735,35 +775,97 @@ enum {
...
+ *  Different types of btree node formats are supported. While the basic btree
+ *  operations remain the same, the differences are encapsulated in the nodes
+ *  contained in the btree.
+ *  Each supported node type provides the same interface to implement the btree
+ *  operations so that the node-specific changes are captured in the node
+ *  implementation.
+ */
 struct node_type {
 	uint32_t                   nt_id;
 	const char                *nt_name;
```
NikhilBirgade, line 796: space alignment between char and *nt_name according to line 797.

btree/btree.c
```diff
@@ -826,46 +929,65 @@ struct node_op {
...
 static int64_t    node_get  (struct node_op *op, struct td *tree,
 			     struct segaddr *addr, int nxt);
 static void       node_put  (struct nd *node);
 static struct nd *node_try  (struct td *tree, struct segaddr *addr);
-static int64_t    node_alloc(struct node_op *op, struct td *tree, int shift,
-			     struct node_type *nt, struct m0_be_tx *tx, int nxt);
-static int64_t    node_free (struct node_op *op, struct nd *node,
+#endif
+static int64_t    node_alloc(struct node_op *op, struct td *tree, int size,
+			  const struct node_type *nt, int ksize, int vsize,
```
NikhilBirgade, line 953: Alignment with previous line's opening bracket.

btree/btree.c
```diff
@@ -826,46 +929,65 @@ struct node_op {
...
+			  struct m0_be_tx *tx, int nxt);
+#ifndef __KERNEL__
+static int64_t    node_free(struct node_op *op, struct nd *node,
 			     struct m0_be_tx *tx, int nxt);
```
NikhilBirgade, line 957: Alignment with previous line's opening bracket.

btree/btree.c
```diff
@@ -896,87 +1018,126 @@ struct m0_btree_oimpl {
 	struct level    i_level[0];
 };
 
+static struct td        trees[M0_TREE_COUNT];
+static uint64_t         trees_in_use[(M0_TREE_COUNT + sizeof(uint64_t) - 1) /
```
NikhilBirgade, line 1022: Sorry, I did not understand this formula. Can you please explain?

btree/btree.c
```diff
@@ -1476,34 +1986,379 @@ static void generic_move(struct nd *src, struct nd *tgt,
...
+	op.no_opc = NOP_ALLOC;
+	tree_create(&op, &tt, 10, NULL, 0);
+
+	tree = op.no_tree;
+
+	op.no_opc = NOP_ALLOC;
+	node_alloc(&op, tree, 10, nt, 8, 8, NULL, 0);
+	node1 = op.no_node;
+
+	while(run_loop--) {
```
NikhilBirgade, line 2316: space between while and  (

btree/btree.c
```diff
@@ -1351,35 +1865,33 @@ static void ff_rec(struct slot *slot)
...
 static void ff_node_key(struct slot *slot)
 {
-	struct nd      *node = slot->s_node;
-	struct ff_head *h    = ff_data(node);
+	const struct nd  *node = slot->s_node;
+	struct ff_head   *h    = ff_data(node);
 
-	M0_PRE(ff_invariant(node));
-	M0_PRE(slot->s_idx < h->ff_used);
+	M0_PRE(ergo(!(h->ff_used == 0 && slot->s_idx == 0),slot->s_idx <= h->ff_used));
```
NikhilBirgade, line 1883: 80 char limit.

btree/btree.c
```diff
@@ -735,35 +776,97 @@ enum {
...
+
+	/** Initializes newly allocated node */
+	void (*nt_init)(const struct nd *node, int shift, int ksize, int vsize);
+
+	/** Cleanup of the node if any before deallocation */
+	void (*nt_fini)(const struct nd *node);
+
+	/** Returns count of records in the node */
 	int  (*nt_count)(const struct nd *node);
+
```
RadhaGulhane13, line 808: It should be /** Returns count of **keys** in the node */ instead of records

btree/btree.c
```diff
@@ -1388,55 +1912,55 @@ static bool ff_isfit(struct slot *slot)
...
+static bool ff_find(struct slot *slot, const struct m0_btree_key *find_key)
 {
 	struct ff_head *h = ff_data(slot->s_node);
-	int i = 0;
-	int j = h->ff_used;
-	M0_PRE(ff_invariant(slot->s_node));
-	M0_PRE(key->k_data.ov_vec.v_count[0] == h->ff_ksize);
-	M0_PRE(key->k_data.ov_vec.v_nr == 1);
+	int             i = -1;
+	int             j = h->ff_used;
```
RadhaGulhane13, line 1940: int j = node_count(slot->s_node)
because, at internal node, we don't want to consider last key. last key can be null or any random value

btree/btree.c
```diff
@@ -1476,34 +1998,387 @@ static void generic_move(struct nd *src, struct nd *tgt,
 		if (nr == 0 || (nr == NR_EVEN &&
```
RadhaGulhane13, line 1998: one more base case condition needs to be added i.e. srcidx != -1, in case nr==NR_MAX,  srcidx may get -1 when we move all the items from source to target

btree/btree.c
```diff
@@ -1101,111 +1316,345 @@ struct seg_ops {
```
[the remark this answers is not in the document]
shashank-parulekar: Done.

btree/btree.c
RadhaGulhane13: Yes but there are some cases, where we might want to move al records from src to tgt, e.g. in case of deletion, for merging , we might want to transfer all records to tgt node  and delete src node , Also, in case of insertion, whenever overflow will occur at root node, as we do not want to change pointer to root node, we might have to transfer all the data from root to some other node and fill the root with new records.

RadhaGulhane13: at line 1997, 1998:
we need to set srcidx, tgtidx according to left / internal node, current implementation will work for leaf node, but for internal node it will not work, For example, if dir ==D_RIGHT and src=[(K0,c0),....(Kn, Cn),(' ',Cn+1)], according to current implementation, it will start moving record from nth index instead of n+1 th index, which may lead to inconsistent state.

shashank-parulekar: Done

RadhaGulhane13 (approved):
Everything looks good !

ydb242 (approved):
Looks good!

PR #580, opened 2021-04-15
EOS-18145 sns: update ub and ut for parity_math functions

sanjognaik:
Add m0_parity_math_mt_ub as part of run-ub
- m0_parity_math_mt_ub is added in run-ub list to run multi-threaded
  unit benchmarking.
- Improved m0_parity_math_mt_ub with more tests.
- Added functions to spoil failed units and compare the recovered units
  with the original after recovery.

Add unit test for mixed failures
- Added unit test 'reed_solomon_recover_with_fail_vec_rand' to test
  parity_math encode and recover functionality for mixed failures i.e.
  failures in random data fragments as well as in parity fragments.

Unit benchmarking changes
- Updated results_print() function in ub.c to print results with correct
  columns

### DIFF
--- a/sns/ut/parity_math_mt_ub.c
+++ b/sns/ut/parity_math_mt_ub.c
@@ -28,6 +28,7 @@
 #include "lib/assert.h"
 #include "lib/memory.h"
 #include "lib/thread.h"
+#include "lib/arith.h"	     /* m0_rnd64 */
 
 #include "lib/ub.h"
 #include "ut/ut.h"
@@ -35,6 +36,12 @@
 #include "sns/matvec.h"
 #include "sns/ls_solve.h"
 #include "sns/parity_math.h"
+
+#define MAX_NUM_THREADS 30
+#define KB(x)		((x) * 1024)
+#define MB(x)		(KB(x) * 1024)
+
+static uint64_t seed = 42;
 
 struct tb_cfg {
 	uint32_t  tc_data_count;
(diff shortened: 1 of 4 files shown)


sns/ut/parity_math_mt_ub.c
```diff
@@ -44,55 +51,62 @@ struct tb_cfg {
...
 	/* allocate and set fail info */
-	cfg->tc_fail = m0_alloc(cfg->tc_fail_count * sizeof(uint8_t));
+	M0_ALLOC_ARR(cfg->tc_fail, cfg->tc_fail_count);
 	M0_ASSERT(cfg->tc_fail != NULL);
 
-	for (i = 0; i < parity_count; ++i)
-		cfg->tc_fail[i] = 1; /* maximal possible fails */
+	fail_count = (m0_rnd64(&seed) % parity_count) + 1;
+	for (i = 0; i < fail_count; ++i){
+		uint32_t idx = m0_rnd64(&seed) % cfg->tc_fail_count;
```
huanghua78, line 107: The result "idx" might be duplicated with previous iteration.
Is that OK?

huanghua78: Here, we are generating "fail_count" of failure idx.
But these number might be duplicated. So, the cfg->tc_fail[] array may have fewer number of failure index.

sanjognaik: Updated the logic to avoid duplicate failure index.

huanghua78 (approved):

mukundkanekar (approved):

PR #581, opened 2021-04-15
[DTM0] Fix use-after-free in STABLE.

ivan-alekhin:
There is a race condition between the user
thread and the locality thread where
a motr client operation gets stabilized
and the dix_req is getting destroyed.

Here is the scheme:

```
User thread:
	await on STABLE for op1
	...
	fini_entity(op1->entity) <- frees the entity
				    and poisons it.

Locality:
	set op1 state to STABLE <- unblocks the "awaiting".
	dix_req_destroy(op1->dix_req) <- enity is finalized already.
		...
		read op1->entity->id <- reads poisoned bytes.
```

The patch reorders the calls, so that the dix_req is getting
destroyed right before the state of the operation is changed
to "STABLE". In this case, we are always operating on a
valid entity structure inside the dix_req_destroy call.


sergey-shilov (approved):

just-now (commented):
Really good work! Thanks Ivan.

PR #583, opened 2021-04-15
[DTM0] EOS-19130: Fix panic during disconnect.

ivan-alekhin:
The patch fixes the problem where m0d services
were failing because DTM0-related connections
were not properly finalized.

### DIFF
--- a/dtm0/service.c
+++ b/dtm0/service.c
@@ -42,11 +42,12 @@
 #include "reqh/reqh.h"    /* m0_reqh2confc */
 
 static int dtm0_service__ha_subscribe(struct m0_reqh_service *service);
-static int dtm0_service__ha_unsubscribe(struct m0_reqh_service *service);
+static void dtm0_service__ha_unsubscribe(struct m0_reqh_service *service);
 
 static struct m0_dtm0_service *to_dtm(struct m0_reqh_service *service);
 static int dtm0_service_start(struct m0_reqh_service *service);
 static void dtm0_service_stop(struct m0_reqh_service *service);
+static void dtm0_service_prepare_to_stop(struct m0_reqh_service *service);
 static int dtm0_service_allocate(struct m0_reqh_service **service,
 				 const struct m0_reqh_service_type *stype);
 static void dtm0_service_fini(struct m0_reqh_service *service);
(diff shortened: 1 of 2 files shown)


dtm0/service.c
```diff
@@ -62,9 +62,10 @@ static const struct m0_reqh_service_type_ops dtm0_service_type_ops = {
 };
 
 static const struct m0_reqh_service_ops dtm0_service_ops = {
-	.rso_start = dtm0_service_start,
-	.rso_stop  = dtm0_service_stop,
-	.rso_fini  = dtm0_service_fini
+	.rso_start           = dtm0_service_start,
+	.rso_stop            = dtm0_service_stop,
+	.rso_fini            = dtm0_service_fini,
+	.rso_prepare_to_stop = dtm0_service__ha_unsubscribe,
```
sergey-shilov, line 68: Propose to name consistently, dtm0_service_prepare_to_stop() which calls dtm0_service__ha_unsubscribe(). Possibly dtm0_service_prepare_to_stop() will be extended.

dtm0/service.c
```diff
@@ -479,7 +497,7 @@ static bool process_clink_cb(struct m0_clink *clink)
 	 * trigger for handling HA messages in this case.
 	 */
 	if (evented_proc_state != M0_NC_TRANSIENT &&
-	    !ready_to_process_ha_msgs && !m0_dtm0_in_ut()) {
+	    !ready_to_process_ha_msgs) {
```
sergey-shilov, line 500: Why m0_dtm0_in_ut() check is not needed anymore?

dtm0/service.c
ivan-alekhin: makes sense, I'll change that.

dtm0/service.c
ivan-alekhin: It has been checked already at the start:

```c
	if (m0_dtm0_in_ut()) {
		M0_LOG(M0_DEBUG, "No HA handling in UT.");
		goto out;
	}
```

I don't think it is necessary to check this again. That's why I have removed it.

sergey-shilov: Agreed.

dtm0/service.c
```diff
@@ -233,33 +234,50 @@ m0_dtm0_service_process_connect(struct m0_reqh_service *s,
...
+	M0_SET0(&process->dop_rlink);
 
 	return M0_RC(rc);
 }
 
+M0_INTERNAL int
+m0_dtm0_service_process_disconnect(struct m0_reqh_service *s,
+				   struct m0_fid          *remote_srv)
+{
+	struct dtm0_process *process;
```
just-now, line 270: nit: `struct dtm0_process *process = dtm0_service_process__lookup(s, remote_srv);`

just-now (approved):

dtm0/service.c
ivan-alekhin: done, and I also folded the return statements.

sergey-shilov (approved):

PR #594, opened 2021-04-20
EOS-19527 : Credits for btree Operations

RadhaGulhane13:


RadhaGulhane13:
There is one  parameter: const struct m0_btree  *tree, which is passed in some of the functions for credit but it is never been used as currently I have defined max node size, max key size, etc as constant  , but in future , we might want to get it from tree parameter therefore I haven't removed this tree parameter but if get these value in different way , we can remove tree parameter.

btree/btree.c
```diff
@@ -1501,6 +1502,253 @@ static void generic_move(struct nd *src, struct nd *tgt,
...
 }
 
+
+/**
+ *  -------------------------------------------------------
+ *  Section START - Credit calculation for BTree Operations
+ *  -------------------------------------------------------
+ */
+
+//In case of fix key, fix value node format: max_ksize, max_vsize will be fix key, value size else estimate max sizes
```
huanghua78, line 1512: Replace "fix" with "fixed".
No C++ style comments.

btree/btree.c
```diff
@@ -1501,6 +1502,253 @@ static void generic_move(struct nd *src, struct nd *tgt,
...
+m0_bcount_t max_ksize; 
+m0_bcount_t max_vsize;
+m0_bcount_t max_node_size;
+m0_bcount_t internal_vsize; //for internal node, value size is fix
+
+/** threshold can be used in btree_move_key_value_credit() to identify 
+ * what fraction of node needs to be move in case of overflow/underflow
+ */
```
huanghua78, line 1520: A Doxygen comment is like the following:
```
/**
 * This is some Doxygen comments for a function.
 *  We will use this as an example.
 *  Please follow the format in this example.
 */
int some_function(int a);
```

btree/btree.c
```diff
@@ -1501,6 +1502,253 @@ static void generic_move(struct nd *src, struct nd *tgt,
...
+enum threshold{
```
huanghua78, line 1521:
```
enum threshold {
```
a space is needed after the identifier.

btree/btree.c
```diff
@@ -1501,6 +1502,253 @@ static void generic_move(struct nd *src, struct nd *tgt,
...
+ * @accum.
+ */
+static void btree_node_update_credit(struct m0_be_tx_credit *accum,
+				     m0_bcount_t             nr)
+{
+	struct m0_be_tx_credit cred = {};
+
+	/* @todo: why multiplication with 2 is needed? */
+	m0_be_tx_credit_mac(&cred,
+			    &M0_BE_TX_CREDIT_TYPE(max_node_size), 2);
```
huanghua78, line 1556: I think this is wrong.
1. For a btree, its node size should be known at this moment. We need to take a "btree" as parameter and then get its node size.
2. if the node size is unknown, we may use the `max_node_size`. But, the use of M0_BE_TX_CREDIT_TYPE() here is wrong. This is only to measure the size of a "m0_bcount_t". This is apparently wrong.

btree/btree.c
```diff
@@ -1501,6 +1502,253 @@ static void generic_move(struct nd *src, struct nd *tgt,
...
+{
+	struct m0_be_tx_credit cred = {};
+
+	btree_node_alloc_credit(tree, &cred);
+	btree_node_update_credit(&cred, 1);  /* @todo: is btree_node_update_credit needed? */
+	m0_be_tx_credit_mac(accum, &cred, nr);
+}
+
+/**
+ * Calculates the credit needed to move some records between two nodes and adds this credit to
```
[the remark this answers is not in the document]
huanghua78: Please fix all places with the same type of issue.

btree/btree.c
huanghua78: What threshold?
Please have a more meaningful name, and add comments.

btree/btree.c
```diff
@@ -1501,6 +1502,253 @@ static void generic_move(struct nd *src, struct nd *tgt,
...
+	
+	btree_insert_leaf_node_credit(tree, ksize, vsize, accum);
+
+	struct m0_be_tx_credit cred = {};
+	btree_insert_internal_node_credit(tree, &cred);
+	m0_be_tx_credit_mac(accum, &cred, height - 1);
+
+	btree_insert_new_root_credit(tree, accum);
+
+	m0_be_tx_credit_mul(accum, nr);
```
huanghua78, line 1653: There are two issues:
1. the `accum` is passed as an arguments. We can not multiply it here directly. It may have other operations. We need to define a local variable to hold the credit.
2. Are you sure all the credit, from the root, to internal node, to leaf node, are all needed to be multiplied by 'nr'? YEvery insertion will trigger the whole path to the root to be updated?

btree/btree.c
```diff
@@ -1501,6 +1502,253 @@ static void generic_move(struct nd *src, struct nd *tgt,
...
+}
+
+static void btree_mem_free_credit(const struct m0_btree  *btree,
+				  m0_bcount_t             size,
+				  struct m0_be_tx_credit *accum)
+{
+	m0_be_allocator_credit(NULL, M0_BAO_FREE_ALIGNED, 0, 0, accum);
```
huanghua78, line 1660: If this is a free operation, no credit is needed at all?

btree/btree.c
```diff
@@ -1501,6 +1502,253 @@ static void generic_move(struct nd *src, struct nd *tgt,
...
+}
+
+static void btree_delete_key_value_credit(const struct m0_btree  *tree,
+					  m0_bcount_t             ksize,
+					  m0_bcount_t             vsize,
+					  struct m0_be_tx_credit *accum)
+{
+	btree_mem_free_credit(tree, ksize + vsize, accum);
+	m0_be_tx_credit_add(accum,
+			    &M0_BE_TX_CREDIT_TYPE(struct m0_btree_rec));
```
huanghua78, line 1684: Deleting key-value may trigger node merge?

btree/btree.c
```diff
@@ -1501,6 +1502,253 @@ static void generic_move(struct nd *src, struct nd *tgt,
...
+
+	btree_delete_leaf_node_credit(tree, ksize, vsize, accum);
+
+	struct m0_be_tx_credit cred = {};
+	btree_delete_internal_node_credit(tree, &cred);
+	m0_be_tx_credit_mac(accum, &cred, height - 2);
+	
+	delete_root_credit(tree, accum);
+
+	m0_be_tx_credit_mul(accum, nr);
```
huanghua78, line 1749: Similar issues with "insert_credit()"

btree/btree.c
RadhaGulhane13: okay

btree/btree.c
RadhaGulhane13: In some conditions, such as underflow/overflow, there will be some threshold point, after that we will move records between two nodes. To figure out how many records or fraction of node needs to be moved, we can use threshold . e.g, in case of deletion, there can be underflow, so in worst case , we might need to move ~50% of records(in case of merging), therefore, MIN_MOVE will be used, and later on we can update these values, once we decide the conditions for underflow, overflow

btree/btree.c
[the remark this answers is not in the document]
RadhaGulhane13: okay

btree/btree.c
```diff
@@ -1501,6 +1502,253 @@ static void generic_move(struct nd *src, struct nd *tgt,
...
+ * @accum.
+ */
+static void m0_btree_create_node_credit(const struct m0_btree  *tree,
+					m0_bcount_t             nr,
+					struct m0_be_tx_credit *accum)
+{
+	struct m0_be_tx_credit cred = {};
+
+	btree_node_alloc_credit(tree, &cred);
+	btree_node_update_credit(&cred, 1);  /* @todo: is btree_node_update_credit needed? */
```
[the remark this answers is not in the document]
RadhaGulhane13: I will fix it

btree/btree.c
```diff
@@ -1501,6 +1502,253 @@ static void generic_move(struct nd *src, struct nd *tgt,
...
+	m0_be_tx_credit_mac(accum, &cred, nr);
+}
+
+/**
+ * Calculates the credit needed to move some records between two nodes and adds this credit to
+ * @accum.
+ */
+static void btree_move_key_value_credit(struct m0_be_tx_credit *accum, enum threshold threshold)
```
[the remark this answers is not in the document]
RadhaGulhane13: okay

btree/btree.c
```diff
@@ -1501,6 +1502,253 @@ static void generic_move(struct nd *src, struct nd *tgt,
...
+{
+	/*depending on thershold, calculate fraction which we want to move other node*/
+	m0_bcount_t size = max_node_size * threshold/10;
```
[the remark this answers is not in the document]
RadhaGulhane13: Yes we can use percentage as well, I thought, ultimately last zero is going to get cancelled, therefore, calculated with base 10, but percentage might provide more readability, I will change the code and calculate based on percentage

btree/btree.c
RadhaGulhane13: 1. okay I will update it 
2. As,  for any insert operation we need to reach leaf node to insert any node, every insert will trigger whole path, and at the time of credit calculation, we might not have information about the btree, height, etc, all the credit calculation is based on worst cases of credits requirement (e.g, in worst case of insert, we might required to split every leaf , internal node and allocate new node at each level ), therefore, all insert/delete operation is multiplied with nr

btree/btree.c
RadhaGulhane13: size needs to be passed, I will make changes

btree/btree.c
RadhaGulhane13: I have consider credits required for  merging btree_delete_leaf_node_credit() and btree_delete_internal_node_credit(), more specifically, btree_move_key_value_credit(accum, MIN_MOVE); provide credits for merging/ rebalancing two nodes.

btree/btree.c
RadhaGulhane13: Explained in insertion part
[link](https://github.com/Seagate/cortx-motr/pull/594#discussion_r629852782)

huanghua78: Also, we need to have a local variable to estimate the credit, and then multiply it with "nr":
```
	uint32_t height = BTREE_HEIGHT_MAX;
	struct m0_be_tx_credit cred = {};
	struct m0_be_tx_credit local_accum = {};


	btree_delete_leaf_node_credit(tree, ksize, vsize, &cred);

	btree_delete_internal_node_credit(tree, &local_accum);
	m0_be_tx_credit_mac(&local_accum, &cred, height - 2);

	delete_root_credit(tree, &local_accum);

	m0_be_tx_credit_mul(&local_accum, nr);


	m0_be_tx_credit_add(accum, &local_accum);

```

RadhaGulhane13: Yes I got it I will  make changes

btree/btree.c
```diff
@@ -1501,6 +1502,263 @@ static void generic_move(struct nd *src, struct nd *tgt,
...
+
+/**
+ * Calculates the credit needed to move some records between two nodes and
+ * adds this credit to @accum.
+ */
+static void btree_move_key_value_credit(struct m0_be_tx_credit *accum,
+					enum node_threshold th)
+{
+	/*depending on node_threshold, calculate fraction which needs to be
+	moved to other node*/
```
huanghua78, line 1586: It's better to have :
```
/* 
 * depending on node_threshold, calculate fraction which needs to be
 * moved to other node.
 */
```

for multiple lines of comments.

btree/btree.c
```diff
@@ -1501,6 +1502,263 @@ static void generic_move(struct nd *src, struct nd *tgt,
...
+	m0_bcount_t size = max_node_size*th/100;
```
huanghua78, line 1587:
```
max_node_size * th / 100;
```
Space is needed between operand and operator.

btree/btree.c
```diff
@@ -1501,6 +1502,263 @@ static void generic_move(struct nd *src, struct nd *tgt,
...
+	btree_insert_key_value_credit(tree, max_ksize, internal_vsize, &cred);
+	m0_be_tx_credit_mac(accum, &cred, 2);	
+}
+
+/**
+ * Calculates total credits required to perform @nr insert operations
+ * @param nr     Number of insert operations.
+ * @param ksize  Key data size.
+ * @param vsize  Value data size.
+*/
```
huanghua78, line 1641: Please add comment:
@param accum the accumulated credit will be updated to have credit for insert operation.

RadhaGulhane13: okay

btree/btree.c
RadhaGulhane13: yes

RadhaGulhane13:
Closing this PR as this work needs to be start again due to implementation changes in btree operation.

PR #595, opened 2021-04-20
[DTM0] Do not release cs_params at the wrong moment.

ivan-alekhin:
Due to an "if (COND) X; Y;"-like typo, the "cs_params"
field was finalized inside service_decode.
The patch addresses this issue. Also, it adds more
traces into the function that uses these arguments,
and it fixes some minor coding style-related issues
there.

### DIFF
--- a/conf/objs/service.c
+++ b/conf/objs/service.c
@@ -90,9 +90,10 @@ service_encode(struct m0_confx_obj *dest, const struct m0_conf_obj *src)
 	}
 
 	rc = arrfid_from_dir(&d->xs_sdevs, s->cs_sdevs);
-	if (rc != 0)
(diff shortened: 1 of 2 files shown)


conf/objs/service.c
```diff
@@ -90,9 +90,10 @@ service_encode(struct m0_confx_obj *dest, const struct m0_conf_obj *src)
 	}
 
 	rc = arrfid_from_dir(&d->xs_sdevs, s->cs_sdevs);
-	if (rc != 0)
+	if (rc != 0) {
```
huanghua78, line 93: Nice catch.

huanghua78 (approved):

sergey-shilov (approved):

PR #596, opened 2021-04-21
pool: improve struct m0_pool_version doc

andriytk:
Extend struct m0_pool_version documentation with a few more clarifying sentences.


pool/pool.h
```diff
@@ -98,7 +98,14 @@ struct m0_pool {
 };
 
 /**
- * Pool version is the subset of devices from the filesystem.
+ * Pool version is the subset of devices from the filesystem
+ * configuration: racks, controllers, disks, etc. If represent
+ * the whole cluster hardware configuration as a tree, the pool
+ * version would be its sub-tree.
```
huanghua78, line 104: If the whole cluster hardware configuration is represented as a tree, the pool version would be its sub-tree.

pool/pool.h
```diff
@@ -98,7 +98,14 @@ struct m0_pool {
...
+ *
+ * Mero objects and indexes are assigned some pool version when
```
huanghua78, line 106: s/Mero/Motr

pool/pool.h
andriytk: Accepted, thanks.

pool/pool.h
andriytk: Right, fixed. :)

huanghua78 (approved):

PR #603, opened 2021-04-23
doc: restored lnet transport pictures

andriytk:
Restored LNet pictures from the old HLD pdf found on my laptop.


huanghua78:
These images are already fixed in https://github.com/Seagate/cortx-motr/blob/main/doc/Motr_LNet_Transport.rst

Yes, they are messed up in PDF file. So, the fixed PDF file can be accepted.

andriytk:
@huanghua78 I don't see them fixed in the mentioned rst doc. They look broken to me there.

nikitadanilov (approved):
"c2_" is outdated, but I guess this is still an improvement.

huanghua78:
Hi @andriytk , I tried to modify the image files to update it with m0_ prefix. After I did it, I found you'd done it.

huanghua78 (approved):

andriytk:
@huanghua78, sorry. I updated my last comment with the information that I will try to do it almost at the same time when I wrote it. But I'm afraid github does not send notification emails for the comments updates.

PR #618, opened 2021-05-01
scripts/systemd: fix m0d startup on Ubuntu

andriytk:
EnvironmentFile is hardcoded at /etc/sysconfig directory
which is absent in Ubuntu/Debian, so the startup fails.

Solution: drop EnvironmentFile configuration, it's redundant
for m0d@.service. motr-server script which is used to startup
m0d service can automatically determine the location of the
environment files and use them.


andriytk:
cc @chumakd

huanghua78 (approved):

madhavemuri (approved):

PR #630, opened 2021-05-04
client: fix error handling at pargrp_iomap_populate()

andriytk:
Return immediately if `pi_seg_process()` fails.

The regression was introduced at b99b9a13dcd7911b19e302ceff822ec33bb71a32.


andriytk:
The patch was added to PR https://github.com/Seagate/cortx-motr/pull/616.

PR #632, opened 2021-05-05
[DTM0] EOS-18924: Handle persistent messages (srv side)

mehjoshi:
This PR is for merging commits related to handling of persistent messages in dtm0.

### DIFF
--- a/cas/service.c
+++ b/cas/service.c
@@ -20,6 +20,7 @@
  */
 
 
+#include "dtm0/tx_desc.h"
 #define M0_TRACE_SUBSYSTEM M0_TRACE_SUBSYS_CAS
 #include "be/op.h"
 #include "be/tx_credit.h"
@@ -1052,29 +1053,28 @@ static int cas_dtm0_logrec_credit_add(struct m0_fom *fom0)
 	return M0_RC(rc);
 }
 
-static int cas_dtm0_logrec_add(struct m0_fom *fom0, struct m0_dtm0_tx_desc *txd,
+static int cas_dtm0_logrec_add(struct m0_fom *fom0,
 			       enum m0_dtm0_tx_pa_state state)
 {
 	/* log the dtm0 logrec before completing the cas op */
 	struct m0_dtm0_service *dtms =
 		m0_dtm0_service_find(fom0->fo_service->rs_reqh);
 	struct m0_dtm0_tx_desc *msg = &cas_op(fom0)->cg_txd;
-	struct m0_buf	        buf = {};
+	struct m0_buf           buf = {};
 	int                     i;
-	int			rc;
+	int                     rc;
 
 	for (i = 0; i < msg->dtd_ps.dtp_nr; ++i) {
 		if (m0_fid_eq(&msg->dtd_ps.dtp_pa[i].p_fid,
-					&dtms->dos_generic.rs_service_fid)) {
-			msg->dtd_ps.dtp_pa[i].p_state =
-				(uint32_t)M0_DTPS_PERSISTENT;
+			      &dtms->dos_generic.rs_service_fid)) {
+			msg->dtd_ps.dtp_pa[i].p_state = state;
 			break;
 		}
 	}
 	rc = m0_xcode_obj_enc_to_buf(&M0_XCODE_OBJ(m0_cas_op_xc, cas_op(fom0)),
 				     &buf.b_addr, &buf.b_nob) ?:
-		m0_dtm0_logrec_update(dtms->dos_log, &fom0->fo_tx.tx_betx, msg,
-				      &buf);
+	     m0_dtm0_logrec_update(dtms->dos_log, &fom0->fo_tx.tx_betx, msg,
+				   &buf);
 	m0_buf_free(&buf);
 
 	return rc;
@@ -1232,7 +1232,7 @@ static int cas_fom_tick(struct m0_fom *fom0)
 		 */
 		if (phase == M0_FOPH_TXN_COMMIT_WAIT &&
 		    m0_fom_phase(fom0) == M0_FOPH_FINISH && is_dtm0_used) {
-			rc = m0_dtm0_on_committed(fom0, &cas_op(fom0)->cg_txd);
+			rc = m0_dtm0_on_committed(fom0, &cas_op(fom0)->cg_txd.dtd_id);
 			if (rc != 0)
 				M0_LOG(M0_WARN, "Could not send PERSISTENT "
 				       "messages out");
@@ -1691,7 +1691,6 @@ static int cas_fom_tick(struct m0_fom *fom0)
 		break;
 	case CAS_DTM0:
 		rc = cas_dtm0_logrec_add(&fom->cf_fom,
-					 &cas_op(&fom->cf_fom)->cg_txd,
 					 M0_DTPS_PERSISTENT);
 		if (rc != 0)
 			cas_fom_failure(fom, M0_ERR(rc), opc == CO_CUR);
--- a/dtm0/fop.c
+++ b/dtm0/fop.c
@@ -353,7 +353,7 @@ M0_INTERNAL int m0_dtm0_on_committed(struct m0_fom            *fom,
 
 static int dtm0_fom_tick(struct m0_fom *fom)
 {
-	int                       rc;
+	int                       rc = 0;
 	struct   m0_dtm0_service *svc;
 	struct   m0_buf           buf = {};
 	struct   dtm0_rep_fop    *rep;
--- a/dtm0/fop.c
+++ b/dtm0/fop.c
@@ -343,22 +343,6 @@ M0_INTERNAL int m0_dtm0_on_committed(struct m0_fom                *fom,
 	return M0_RC(rc);
 }
 
-/*
- * Wait for the BE transaction to be persisted.
- */
-static int op_sync_wait(struct m0_fom *fom)
-{
-    struct m0_be_tx     *tx;
-
-    tx = m0_fom_tx(fom);
-    if (m0_be_tx_state(tx) < M0_BTS_LOGGED) {
-	M0_LOG(M0_DEBUG, "fom wait for tx to be logged");
-	m0_fom_wait_on(fom, &tx->t_sm.sm_chan, &fom->fo_cb);
-	return M0_FSO_WAIT;
-    }
-    return M0_FSO_AGAIN;
-}
-
 static int dtm0_fom_tick(struct m0_fom *fom)
 {
 	int                       rc;
@@ -389,12 +373,9 @@ static int dtm0_fom_tick(struct m0_fom *fom)
 						      &dtm0logrec_cred);
 				m0_be_tx_credit_add(&fom->fo_tx.tx_betx_cred, &dtm0logrec_cred);
 			}
-			if (phase == M0_FOPH_TXN_COMMIT) {
-				rc = op_sync_wait(fom);
-			}
-			if (m0_dtm0_in_ut() && m0_fom_phase(fom) == M0_FOPH_QUEUE_REPLY) {
-				m0_fom_phase_set(fom, M0_FOPH_TXN_COMMIT_WAIT);
-			}
+			/*
+			 * XXX: Find out if we need to wait for the tx to be logged.
+			 */
 		}
 	} else {
 		M0_ASSERT(m0_fom_phase(fom) == M0_FOPH_TYPE_SPECIFIC);
@@ -423,13 +404,22 @@ static int dtm0_fom_tick(struct m0_fom *fom)
 			if (m0_dtm0_is_a_volatile_dtm(fom->fo_service)) {
 				m0_be_dtm0_log_pmsg_post(svc->dos_log, fom->fo_fop);
 			} else {
-				m0_mutex_lock(&svc->dos_log->dl_lock);
+				/* m0_mutex_lock(&svc->dos_log->dl_lock); */
 				m0_dtm0_logrec_update(svc->dos_log, &fom->fo_tx.tx_betx, &req->dtr_txr, &buf);
-				m0_mutex_unlock(&svc->dos_log->dl_lock);
+				/* m0_mutex_unlock(&svc->dos_log->dl_lock); */
+				if (m0_dtm0_in_ut()) {
+					rc = m0_dtm0_tx_desc_copy(&req->dtr_txr, &rep->dr_txr);
+					if (rc != 0)
+						m0_fom_phase_move(fom, M0_ERR(rc), M0_FOPH_FAILURE);
+					rc = m0_dtm0_send_msg(fom, DTM_PERSISTENT,
+						&req->dtr_txr.dtd_id.dti_fid,
+						&req->dtr_txr);
+					if (rc != 0)
+					    m0_fom_phase_move(fom, M0_ERR(rc), M0_FOPH_FAILURE);
+				}
 			}
 		}
-
-		rep->dr_rc = 0;
+		rep->dr_rc = rc;
 		m0_fom_phase_set(fom, M0_FOPH_SUCCESS);
 		rc = M0_FSO_AGAIN;
 	}
--- a/dtm0/fop.c
+++ b/dtm0/fop.c
@@ -343,19 +343,61 @@ M0_INTERNAL int m0_dtm0_on_committed(struct m0_fom                *fom,
 	return M0_RC(rc);
 }
 
+/*
+ * Wait for the BE transaction to be persisted.
+ */
+static int op_sync_wait(struct m0_fom *fom)
+{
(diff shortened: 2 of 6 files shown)


dtm0/fop.c
```diff
@@ -307,55 +307,86 @@ M0_INTERNAL int m0_dtm0_logrec_update(struct m0_be_dtm0_log  *log,
...
-	/* TODO: Send P msgs to the rest of the participants. */
-	m0_dtm0_tx_desc_fini(&msg);
+	m0_mutex_lock(&log->dl_lock);
+	/* Get the latest state of the log record. */
+	rec = m0_be_dtm0_log_find(log, id);
+	m0_mutex_unlock(&log->dl_lock);
+	M0_ASSERT_INFO(rec != NULL, "Log record must be inserted into the log "
+		       "in cas_fom_tick().");
+	M0_ASSERT_INFO(m0_dtm0_tx_desc_state_eq(&rec->dlr_txd,
+		       M0_DTPS_PERSISTENT), "Log record not persistent");
```
ivan-alekhin, line 334: Some of the parts may not be persistent yet, and it is perfectly fine. I guess we do not need this assertion here.

dtm0/fop.c
```diff
@@ -307,55 +307,86 @@ M0_INTERNAL int m0_dtm0_logrec_update(struct m0_be_dtm0_log  *log,
...
+	M0_PRE(log->dl_is_persistent);
+
+	M0_ENTRY();
 
-	/* TODO: Send P msgs to the rest of the participants. */
-	m0_dtm0_tx_desc_fini(&msg);
+	m0_mutex_lock(&log->dl_lock);
+	/* Get the latest state of the log record. */
+	rec = m0_be_dtm0_log_find(log, id);
+	m0_mutex_unlock(&log->dl_lock);
```
ivan-alekhin, line 330: is it safe to release this lock here? If we have PPP here then this record might be removed by the log pruner. What if we get a copy of the record's txd, so that the lock can be safely released?

dtm0/fop.c
```diff
@@ -307,55 +307,86 @@ M0_INTERNAL int m0_dtm0_logrec_update(struct m0_be_dtm0_log  *log,
...
+	M0_ASSERT_INFO(rec != NULL, "Log record must be inserted into the log "
+		       "in cas_fom_tick().");
+	M0_ASSERT_INFO(m0_dtm0_tx_desc_state_eq(&rec->dlr_txd,
+		       M0_DTPS_PERSISTENT), "Log record not persistent");
+
+	/* Notify the originator */
+	rc = m0_dtm0_send_msg(fom, DTM_PERSISTENT, &rec->dlr_txd.dtd_id.dti_fid, &rec->dlr_txd);
```
ivan-alekhin, line 337: I guess we can go out right here, so that `rc2` won't be needed

dtm0/fop.c
```diff
@@ -307,55 +307,86 @@ M0_INTERNAL int m0_dtm0_logrec_update(struct m0_be_dtm0_log  *log,
...
+
+	/* Send P msgs to the rest of the participants. */
+	for (i = 0; i < rec->dlr_txd.dtd_ps.dtp_nr; ++i) {
+		if (m0_fid_eq(&rec->dlr_txd.dtd_ps.dtp_pa[i].p_fid,
+			      &dtms->dos_generic.rs_service_fid))
+			continue;
+		rc2 = m0_dtm0_send_msg(fom, DTM_PERSISTENT, &rec->dlr_txd.dtd_ps.dtp_pa->p_fid, &rec->dlr_txd);
```
ivan-alekhin, line 344: (style) it looks like we getting crossing the 80-char-per-line limit here

dtm0/fop.c
```diff
@@ -307,55 +307,86 @@ M0_INTERNAL int m0_dtm0_logrec_update(struct m0_be_dtm0_log  *log,
...
+		if (rc == 0 && rc2 != 0)
+			rc = M0_ERR_INFO(-rc2, "failed to send PERSISTENT msg " FID_F " -> " FID_F,
```
ivan-alekhin, line 346: If m0_dtm0_send_msg returns a negative error code then we should not make it positive.

cas/service.c
```diff
@@ -20,6 +20,7 @@
  */
 
 
+#include "dtm0/tx_desc.h"
```
ivan-alekhin, line 23: (style) usually, we add "include" directives after M0_TRACE_SUBSYSTEM definition. Let's add it to the group of the other dtm0/* directives. Also, please check this thing is needed: I guess dtm0/fop.h should include it.

dtm0/fop.c
```diff
@@ -369,28 +400,36 @@ static int dtm0_fom_tick(struct m0_fom *fom)
...
+			if (rc != 0)
+				m0_fom_phase_move(fom, M0_ERR(rc), M0_FOPH_FAILURE);
+		} else if (req->dtr_msg == DTM_PERSISTENT) {
+			if (m0_dtm0_is_a_volatile_dtm(fom->fo_service)) {
+				m0_be_dtm0_log_pmsg_post(svc->dos_log, fom->fo_fop);
+			} else {
+				m0_dtm0_logrec_update(svc->dos_log, &fom->fo_tx.tx_betx, &req->dtr_txr, &buf);
+				if (m0_dtm0_in_ut()) {
+					rc = m0_dtm0_tx_desc_copy(&req->dtr_txr, &rep->dr_txr);
+					if (rc != 0)
```
ivan-alekhin, line 418: I guess, if we are inside a UT then we can just use assertions here.

dtm0/fop.c
```diff
@@ -369,28 +400,36 @@ static int dtm0_fom_tick(struct m0_fom *fom)
...
-			m0_be_dtm0_log_pmsg_post(svc->dos_log, fom->fo_fop);
-		}
-
-		rep->dr_rc = 0;
+		rep->dr_rc = rc;
 		m0_fom_phase_set(fom, M0_FOPH_SUCCESS);
 		rc = M0_FSO_AGAIN;
 	}
-
+	M0_ADDB2_ADD(M0_AVI_FOM_TO_TX, phase_sm_id, rpc_sm_id);
```
ivan-alekhin, line 432: Is it a copy-paste error or is it some sophisticated mapping?

dtm0/ut/main.c
```diff
@@ -29,6 +29,7 @@
 #include "ut/ut.h"
 #include "cas/cas.h"
 #include "cas/cas_xc.h"
+#include "dtm0/service.h"             /* m0_dtm0_service_find */
```
ivan-alekhin, line 32: It looks like this thing has already been included

dtm0/ut/main.c
```diff
@@ -105,7 +112,41 @@ static void dtm0_ut_send_fops(struct m0_rpc_session *cl_rpc_session)
...
+		.dti_fid = cli_srv_fid
+	};
+	fop = m0_fop_alloc_at(cl_rpc_session,
+			      &dtm0_req_fop_fopt);
+	req = m0_fop_data(fop);
+	req->dtr_msg = DTM_PERSISTENT;
+	req->dtr_txr = txr;
+
+	m0_mutex_lock(&log->dl_lock);
+	rc = m0_be_dtm0_log_update(log, NULL, &txr, &(struct m0_buf){});
```
ivan-alekhin, line 137: Why do we need to update the log here? shouldn't it be able to work in the case where Pmsg was received before the corresponding request was executed?
Also, is it safe to use NULL instead of tx here? I guess BE will cause SIGSEGV or panic.

ivan-alekhin (commented):
I have posted some questions/suggestions

cas/service.c
```diff
@@ -1052,29 +1053,28 @@ static int cas_dtm0_logrec_credit_add(struct m0_fom *fom0)
...
+			      &dtms->dos_generic.rs_service_fid)) {
+			msg->dtd_ps.dtp_pa[i].p_state = state;
 			break;
 		}
 	}
 	rc = m0_xcode_obj_enc_to_buf(&M0_XCODE_OBJ(m0_cas_op_xc, cas_op(fom0)),
 				     &buf.b_addr, &buf.b_nob) ?:
-		m0_dtm0_logrec_update(dtms->dos_log, &fom0->fo_tx.tx_betx, msg,
-				      &buf);
+	     m0_dtm0_logrec_update(dtms->dos_log, &fom0->fo_tx.tx_betx, msg,
```
sergey-shilov, line 1076: Please check indent here (possibly a github's issue).

cas/service.c
```diff
@@ -1052,29 +1053,28 @@ static int cas_dtm0_logrec_credit_add(struct m0_fom *fom0)
...
-static int cas_dtm0_logrec_add(struct m0_fom *fom0, struct m0_dtm0_tx_desc *txd,
+static int cas_dtm0_logrec_add(struct m0_fom *fom0,
 			       enum m0_dtm0_tx_pa_state state)
 {
 	/* log the dtm0 logrec before completing the cas op */
 	struct m0_dtm0_service *dtms =
 		m0_dtm0_service_find(fom0->fo_service->rs_reqh);
 	struct m0_dtm0_tx_desc *msg = &cas_op(fom0)->cg_txd;
-	struct m0_buf	        buf = {};
+	struct m0_buf           buf = {};
```
sergey-shilov, line 1063: Please check spaces here (possibly a github's issue).

dtm0/fop.c
sergey-shilov: I guess we can check the rc here and log that we could not notify originator and continue with sending messages to participants.

dtm0/fop.c
sergey-shilov: Please use rc here instead of rc2.

dtm0/fop.c
sergey-shilov: Please remove rc2 here.

dtm0/fop.c
```diff
@@ -307,55 +307,86 @@ M0_INTERNAL int m0_dtm0_logrec_update(struct m0_be_dtm0_log  *log,
...
+				     const struct m0_dtm0_tid *id)
 {
 	struct m0_dtm0_service *dtms = m0_dtm0_service_find(
-		fom->fo_service->rs_reqh);
-	struct m0_dtm0_tx_desc  msg = {};
+					fom->fo_service->rs_reqh);
+	struct m0_be_dtm0_log  *log = dtms->dos_log;
+	struct m0_dtm0_log_rec *rec;
 	int                     rc;
+	int                     rc2;
```
sergey-shilov, line 318: Please remove rc2, it is not needed according to further comments.

dtm0/fop.c
```diff
@@ -307,55 +307,86 @@ M0_INTERNAL int m0_dtm0_logrec_update(struct m0_be_dtm0_log  *log,
...
 }
 
-
 static int dtm0_fom_tick(struct m0_fom *fom)
 {
-	int                     rc;
-	struct dtm0_req_fop    *req;
-	struct dtm0_rep_fop    *rep;
-	struct m0_dtm0_service *svc;
+	int                       rc = 0;
```
sergey-shilov, line 356: Please add a var "int result = M0_FSO_AGAIN;" and use it as a return value of this function and a result of generic FOM execution. rc should be used for result of non-FOM tick functions. Refer to CAS service as an example.

dtm0/fop.c
```diff
@@ -307,55 +307,86 @@ M0_INTERNAL int m0_dtm0_logrec_update(struct m0_be_dtm0_log  *log,
...
+	struct   dtm0_req_fop    *req = m0_fop_data(fom->fo_fop);
+	struct   m0_rpc_item     *item = &fom->fo_fop->f_item;
+	int                       phase = m0_fom_phase(fom);
+	uint64_t                  phase_sm_id = m0_sm_id_get(&fom->fo_sm_phase);
+	uint64_t                  rpc_sm_id   = m0_sm_id_get(&item->ri_sm);
+
+	M0_ENTRY("fom %p phase %d", fom, phase);
 
 	if (m0_fom_phase(fom) < M0_FOPH_NR) {
 		rc = m0_fom_tick_generic(fom);
```
sergey-shilov, line 369: Use result instead of rc here.

dtm0/fop.c
```diff
@@ -369,28 +400,36 @@ static int dtm0_fom_tick(struct m0_fom *fom)
...
-			/* TODO: Only the client side is here so far.
-			 * Remove the "is_volatile" part when plog is ready.
-			 */
-			m0_be_dtm0_log_pmsg_post(svc->dos_log, fom->fo_fop);
-		}
-
-		rep->dr_rc = 0;
+		rep->dr_rc = rc;
 		m0_fom_phase_set(fom, M0_FOPH_SUCCESS);
 		rc = M0_FSO_AGAIN;
```
sergey-shilov, line 430: Use result instead of rc here.

dtm0/fop.c
```diff
@@ -369,28 +400,36 @@ static int dtm0_fom_tick(struct m0_fom *fom)
...
 	}
-
+	M0_ADDB2_ADD(M0_AVI_FOM_TO_TX, phase_sm_id, rpc_sm_id);
 	return M0_RC(rc);
```
sergey-shilov, line 433: Use result instead of rc here.

dtm0/ut/main.c
```diff
@@ -105,7 +112,41 @@ static void dtm0_ut_send_fops(struct m0_rpc_session *cl_rpc_session)
 	M0_ASSERT(m0_dtm0_ts__invariant(&reply_data.dti_ts));
 
 	M0_UT_ASSERT(m0_dtm0_tid_cmp(&dcs, &txr.dtd_id, &reply_data) == M0_DTS_EQ);
+	m0_fop_put_lock(fop);
+
```
sergey-shilov, line 116: Too much blank lines, use a single one.

dtm0/fop.c
```diff
@@ -369,28 +400,36 @@ static int dtm0_fom_tick(struct m0_fom *fom)
...
 					      &req->dtr_txr.dtd_id.dti_fid,
 					      &req->dtr_txr);
-			M0_ASSERT(rc == 0);
+			if (rc != 0)
+				m0_fom_phase_move(fom, M0_ERR(rc), M0_FOPH_FAILURE);
+		} else if (req->dtr_msg == DTM_PERSISTENT) {
+			if (m0_dtm0_is_a_volatile_dtm(fom->fo_service)) {
+				m0_be_dtm0_log_pmsg_post(svc->dos_log, fom->fo_fop);
+			} else {
+				m0_dtm0_logrec_update(svc->dos_log, &fom->fo_tx.tx_betx, &req->dtr_txr, &buf);
```
sergey-shilov, line 415: m0_dtm0_logrec_update() is a non void function, please assign rc and check.

sergey-shilov (changes requested):

dtm0/fop.c
mehjoshi: OK.

dtm0/fop.c
mehjoshi: OK.

dtm0/ut/main.c
mehjoshi: Removed it.

dtm0/ut/main.c
mehjoshi: OK.

cas/service.c
mehjoshi: Moved it along with other dtm0 directives as suggested.
We still need to add dtm0/tx_desc.h separately because the coding guidelines say that we should explicitly include the headers that we need and not depend on it being included by others.

cas/service.c
mehjoshi: Not sure what should be the indentation here. Will leave it at the same level as before.

dtm0/fop.c
mehjoshi: Done. Corrected.

dtm0/fop.c
mehjoshi: Updated the code to not negate the error code.
I will leave the use of  rc2 here for now as per earlier comments.

dtm0/fop.c
ivan-alekhin: @mehjoshi, the idea is not about removal of all rc codes. We are just asking if it is possible to avoid the second variable. There are a lot of ways to do that: panic, log error, exit early, OR them and so on.

dtm0/fop.c
mehjoshi: OK. Will do.

dtm0/fop.c
mehjoshi: OK.

cas/service.c
mehjoshi: If are just using exported structures and functions, there isn't any more symbol isolation to be achieved.
I would agree, however, to keep the list of include directives short and I will remove this include.

PR #633, opened 2021-05-05
Upgrade Go bindings to compatible with Go version 1.16 and above.

kiwionly2:
Existing code no longer work with Go version 1.16 because of Go module.

Hence, update existing code to support module, so it could work with latest version of Go. ( Go version 1.17 above will only support module ).

-----
[View rendered bindings/go/README.md](https://github.com/kiwionly2/cortx-motr/blob/main/bindings/go/README.md)


cortx-admin:

PR #647, opened 2021-05-08
AUTHORS: drop duplicate names

andriytk:


huanghua78 (approved):

AUTHORS
```diff
@@ -54,7 +54,6 @@ Mandar Sawant
 Manish Honap
 Mateusz Kowalczyk
 Maxim Malezhin
-Maxim Medved
```
huanghua78, line 57: Where is the duplication?

andriytk: A few lines above: "Maksym Medvied".

PR #650, opened 2021-05-12
[DTM0] Document known issues for a2a

rajatpatil98:
Known issues added in all2all README.

-----
[View rendered dtm0/it/all2all/README.md](https://github.com/rajatpatil98/cortx-motr/blob/patch-1/dtm0/it/all2all/README.md)

### DIFF
--- a/dtm0/it/all2all/README.md
+++ b/dtm0/it/all2all/README.md
@@ -92,3 +92,21 @@ Now reload the daemon & stop the hare-consul-agent using following command
 [root@ssc-vm-c-1966 hare]# sudo systemctl daemon-reload && sleep 5 && service hare-consul-agent stop 
 ```
 Try running the test again. Refer [this](https://github.com/Seagate/cortx-motr/blob/dtm0-main/dtm0/it/all2all/README.md#command-to-run-this-test)
+
+## Known Issues
+### 1. all2all integration test performs all tests and at the end shows test status as FAIL because of m0d service failed.
+Example test log is as shown below:
+```sh
+ERR: Process m0d@0x7200000000000001:0xc.service failed
+ERR: Process m0d@0x7200000000000001:0x1a.service failed
+ERR: Process m0d@0x7200000000000001:0x28.service failed
+ERR: TEST STATUS: FAIL
+```
+In above case, in journalctl logs if we see the m0d services getting killed because of timed out like shown below, then it is not an issue with all2all test.
+```sh
+m0d@0x7200000000000001:0x28.service stop-sigterm timed out. Killing
+```
+### 2. If message shown below appears multiple times in journalctl log, then it is not an issue with all2all or dtm0 code.
+```sh
+motr[08732]: 4b30 ERROR [rpc/conn.c:1062:m0_rpc_conn_terminate] <! rc=-125
+```
--- a/dtm0/it/all2all/README.md
+++ b/dtm0/it/all2all/README.md
@@ -94,7 +94,9 @@ Now reload the daemon & stop the hare-consul-agent using following command
 Try running the test again. Refer [this](https://github.com/Seagate/cortx-motr/blob/dtm0-main/dtm0/it/all2all/README.md#command-to-run-this-test)
 
 ## Known Issues
-### 1. all2all integration test performs all tests and at the end shows test status as FAIL because of m0d service failed.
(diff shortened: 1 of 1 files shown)


cortx-admin:

vaibhavparatwar:
@huanghua78 could you please review?

dtm0/it/all2all/README.md
```diff
@@ -92,3 +92,21 @@ Now reload the daemon & stop the hare-consul-agent using following command
...
+Example test log is as shown below:
+```sh
+ERR: Process m0d@0x7200000000000001:0xc.service failed
+ERR: Process m0d@0x7200000000000001:0x1a.service failed
+ERR: Process m0d@0x7200000000000001:0x28.service failed
+ERR: TEST STATUS: FAIL
+```
+In above case, in journalctl logs if we see the m0d services getting killed because of timed out like shown below, then it is not an issue with all2all test.
+```sh
+m0d@0x7200000000000001:0x28.service stop-sigterm timed out. Killing
```
huanghua78, line 107: What is the real root cause of this killing which leads to dtm0 test failure?

If we got this issue, what is the solution?
Test it again?

dtm0/it/all2all/README.md
```diff
@@ -92,3 +92,21 @@ Now reload the daemon & stop the hare-consul-agent using following command
...
+```
+### 2. If message shown below appears multiple times in journalctl log, then it is not an issue with all2all or dtm0 code.
+```sh
+motr[08732]: 4b30 ERROR [rpc/conn.c:1062:m0_rpc_conn_terminate] <! rc=-125
```
huanghua78, line 111: What caused this error?
It is harmful?

dtm0/it/all2all/README.md
vaibhavparatwar: @ivan-alekhin  could you please comment?

dtm0/it/all2all/README.md
vaibhavparatwar: @ivan-alekhin  could you please comment?

dtm0/it/all2all/README.md
ivan-alekhin: > What is the real root cause of this killing which leads to dtm0 test failure?

The root cause may be related to https://github.com/Seagate/cortx-hare/pull/1606. Once the upstream gets a solution for this timeout issue, we will verify it to ensure it has fixed the all2all test too. @rajatpatil98, could you please add a link to the ticket (https://jts.seagate.com/browse/EOS-19683) here?

> If we got this issue, what is the solution? Test it again?

Yes -- either run the test again or look into the log to confirm that the process was killed by systemd due to the timeout. That does not sound good but that all we have at the moment.

dtm0/it/all2all/README.md
ivan-alekhin: > could you please comment?

I cannot confirm that "it is not an issue with all2all or dtm0 code". @rajatpatil98, could you please add more details here? did we get a panic? does it happen at the end of the test?

> What caused this error? It is harmful?

@mehjoshi  and @sergey-shilov are working right now on a similar issue  -- m0_panic observed when DTM0 service is about to be stopped. So that, I suspect that this -125 error could cause a panic here too. In this case it is quite harmful, and this entry has to be removed from this list (or at least we need to state that this issue has direct relation to DTM0). Let's wait until Rajat adds more details on that.

ivan-alekhin:
@rajatpatil98 , could you please add more details to the title? Here is the link where you can find some examples (https://github.com/Seagate/cortx-motr/pulls?q=is%3Apr+is%3Aclosed+%5BDTM0%5D).

For instance, you can call it "[DTM0] Document known issues for a2a".

dtm0/it/all2all/README.md
rajatpatil98: updated issue and added link to the ticket EOS-19683.

dtm0/it/all2all/README.md
rajatpatil98: Removed this issue from list, It is not reproducible so we should not put it into list of known issues.

ivan-alekhin (approved):
Overall the patch looks good to me.

sergey-shilov (approved):

Issue #584, opened 2021-04-15
sudo ./scripts/install-build-deps completes with failures=1


hessio:
When I try to build Motr on a single VM using CentOS 7.8.2003 when I run: `sudo ./scripts/install-build-deps` it completes but with failed=1 and then when I try to run `scripts/m0 rebuild` or `scripts/m0 make` this does not work which I presume is expected when the build deps command fails.

mukul-seagate11:
@huanghua78 @madhavemuri, please address the query as raised?

huanghua78:
Please attach detailed error messages here.

hessio:
`Here is the error which causes the failure for the` 

Here is the error when I run the command `sudo ./scripts/install-build-deps` that causes the failure:

```

fatal: [localhost]: FAILED! => {
   "changed": true,
   "cmd": [
       "rpmbuild",
       "-bb",
       "--without",
       "servers",
       "--without",
       "lustre-tests",
       "--define",
       "configure_args --disable-gss-keyring",
       "~/rpmbuild/SPECS/lustre.spec"
   ],
   "delta": "0:00:07.452584",
   "end": "2021-04-19 09:11:53.886878",
   "invocation": {
       "module_args": {
           "_raw_params": "rpmbuild -bb --without servers --without lustre-tests         --define 'configure_args --disable-gss-keyring'                  ~/rpmbuild/SPECS/lustre.spec\n",
           "_uses_shell": false,
           "argv": null,
           "chdir": null,
           "creates": "/root/rpmbuild/RPMS/x86_64/lustre-client-devel-*.rpm",
           "executable": null,
           "removes": null,
           "stdin": null,
           "stdin_add_newline": true,
           "strip_empty_ends": true,
           "warn": true
       }
   },
   "msg": "non-zero return code",
   "rc": 1,
   "start": "2021-04-19 09:11:46.434294",
[...]

PR #656, opened 2021-05-13
prov: fix install-build-deps on old CentOS versions

andriytk:
Old CentOS versions are not available at mirror.centos.org. Instead, vault.centos.org should be used for old versions. Otherwise, `install-build-deps` script would fail, like at https://github.com/Seagate/cortx-motr/issues/584.

Solution: update redhat_vault_repo_versions ansible variable with 7.7 and 7.8 CentOS versions, and don't skip vault-release tag at setup-node script (in order to use the vault repo versions).


andriytk:
cc @chumakd

chumakd (approved):

chumakd:
the only caveat is that enabling `vault-release` tag by default, for all versions of CentOS, might break yum repos of the current release, i.e. CentOS 7.9

its better to test the `install-build-deps` script on both 7.8 and 7.9 versions

andriytk:
Checked on freshly installed CentOS 7.8 and 7.9 versions - works fine.

hessio:
I have also tested this and it is working fine.

huanghua78:
> I have also tested this and it is working fine.

Thanks for this conformation.

madhavemuri (approved):

huanghua78 (approved):

PR #662, opened 2021-05-17
[DTM0] EOS-20249: Versioned pairs in btree and CAS.

ivan-alekhin:
The patch introduces tombstones and versions into btree.
Tombstones and versions may be used to enable the
"conflict-free" behavior: a set of key-value pairs
can always be merged with another set (for example,
a set of REDO records sent from another process).

Human-readable list of changes:
- Version field followed by the key-value data.
- The most significant bit of the version field
  used as the tombstone flag.
- A flag into btree was added. The flag helps
  to distinguish versioned trees from ordinary trees.
- btree_save and btree_delete are trying to AMB the key
  to get the versioned buffer, and they use the version
  specified there.
- btree cursor can be initialized in a way where it
  iterates over only alive or only dead pairs.
- New version-related UTs for btree.
- CAS using versioned btrees for ordinary catalogues.
- New version-related UTs for CAS.

Here is a list of changes of the on-disk format.

1. A new member of the btree - bb_flags.

2. Storage for version.

Right now the version+tombstone field is stored in the following way:
```
 |<- Buffer allocated for kv data     ->|
 |                                      |
 |<- bit  ->|<- u63 ->|                 |
 [ TOMBSTONE, VERSION ] [ KEY ] [ VALUE ]
                       / \      /\
                        |        |
                        |       btree_val
                       btree_key
```

### DIFF
--- a/be/btree.c
+++ b/be/btree.c
@@ -59,6 +59,104 @@ M0_INTERNAL const struct m0_fid_type m0_btree_fid_type = {
 	.ft_name = "btree fid",
 };
 
+struct btree_pair_ver {
+	uint64_t bpv_ver;
+};
+
+/*
+ * The key pointer is aligned with sizeof(void*). We add the version
+ * right before the key. This assertion helps us to ensure that this
+ * addition does not break this requirement for key alignment.
+ * See ::btree_save for details.
+ */
+M0_BASSERT(M0_IS_8ALIGNED(sizeof(struct btree_pair_ver)));
+
+enum {
+	/* Tombstone flag: marks a dead kv pair. We use the MSB here. */
+	BPV_TBS = 1L << (sizeof(uint64_t) * CHAR_BIT - 1),
+
+	/*
+	 * n_inf/p_inf marks non-allowed values of the
+	 * btree_pair_ver::bpv_ver.
+	 */
+	BPV_VER_PINF = UINT64_MAX & ~BPV_TBS,
+	BPV_VER_NINF = 0,
+
+	/* The biggest possible value of a version */
+	BPV_VER_MAX = BPV_VER_PINF - 1,
+	/* The lowest possible value of a version */
+	BPV_VER_MIN = BPV_VER_NINF + 1,
+};
+
+#define VER_DATUM_SIZE (sizeof(struct btree_pair_ver))
+M0_BASSERT(VER_DATUM_SIZE == sizeof(void *));
+
+static inline bool bpv_tbs_is_set(const struct btree_pair_ver *bpv)
+{
+	return bpv->bpv_ver & BPV_TBS;
+}
+
+static inline void bpv_tbs_set(struct btree_pair_ver *bpv)
+{
+	bpv->bpv_ver |= BPV_TBS;
+}
+
+static inline void bpv_tbs_clear(struct btree_pair_ver *bpv)
+{
+	bpv->bpv_ver |= ~BPV_TBS;
+}
+
+static inline uint64_t bpv_ver_get(const struct btree_pair_ver *bpv)
+{
+	return bpv->bpv_ver & ~BPV_TBS;
+}
+
+static bool bpv_invariant(const struct btree_pair_ver *bpv)
+{
+	return (_0C(bpv_ver_get(bpv) <= BPV_VER_MAX) &&
+		_0C(bpv_ver_get(bpv) >= BPV_VER_MIN));
+}
+
+static inline void bpv_ver_set(struct btree_pair_ver *bpv, uint64_t ver)
+{
+	M0_PRE(ver >= BPV_VER_MIN);
+	M0_PRE(ver <= BPV_VER_MAX);
+	bpv->bpv_ver = (bpv->bpv_ver & BPV_TBS) | ver;
+	M0_POST(bpv_invariant(bpv));
+	M0_POST(bpv_ver_get(bpv) == ver);
+}
+
+static inline bool bpv_gt(struct btree_pair_ver *left,
+			  struct btree_pair_ver *right)
+{
+	return bpv_ver_get(left) > bpv_ver_get(right);
+}
+
+/*
+ * KV pair version is a part of every kv pair. However, sometimes we
+ * do not want to consider versions. In this case BPV_NONE could be
+ * used to disable the version-aware logic.
+ */
+#define BPV_NONE ((struct btree_pair_ver) { .bpv_ver = BPV_VER_NINF })
+
+/*
+ * Converts a pointer to the start of the key-value pair data into a pointer
+ * to the corresponding version datum.
+ * TODO: consider replacing with a function if "const" is not in used at
+ * the callers side.
+ * TODO: consider removing the cast to uint8_t -- pointer arithmetic on
+ * void ptrs (gcc extension) might handle that (check if it is enabled).
+ */
+#define PAIR2VER(__pair) \
+	((struct btree_pair_ver *) (((uint8_t *) __pair) - VER_DATUM_SIZE))
+
+/*
+ * Converts a pointer to a version datum into a pointer to the start
+ * of the corresponding key-value pair data.
+ */
+#define VER2PAIR(__ver) \
+	((void *) (((uint8_t *) __ver) + VER_DATUM_SIZE))
+
 static struct m0_be_op__btree *op_tree(struct m0_be_op *op);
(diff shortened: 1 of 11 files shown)


be/btree.c
```diff
@@ -59,6 +59,104 @@ M0_INTERNAL const struct m0_fid_type m0_btree_fid_type = {
...
+	BPV_VER_NINF = 0,
+
+	/* The biggest possible value of a version */
+	BPV_VER_MAX = BPV_VER_PINF - 1,
+	/* The lowest possible value of a version */
+	BPV_VER_MIN = BPV_VER_NINF + 1,
+};
+
+#define VER_DATUM_SIZE (sizeof(struct btree_pair_ver))
+M0_BASSERT(VER_DATUM_SIZE == sizeof(void *));
```
just-now, line 92: this needs a comment

be/btree.c
```diff
@@ -59,6 +59,104 @@ M0_INTERNAL const struct m0_fid_type m0_btree_fid_type = {
...
+static inline void bpv_ver_set(struct btree_pair_ver *bpv, uint64_t ver)
+{
+	M0_PRE(ver >= BPV_VER_MIN);
+	M0_PRE(ver <= BPV_VER_MAX);
+	bpv->bpv_ver = (bpv->bpv_ver & BPV_TBS) | ver;
+	M0_POST(bpv_invariant(bpv));
+	M0_POST(bpv_ver_get(bpv) == ver);
+}
+
+static inline bool bpv_gt(struct btree_pair_ver *left,
```
just-now, line 129: *const* struct btree_pair_ver *left/right

be/btree.c
```diff
@@ -1983,6 +2144,34 @@ M0_INTERNAL void m0_be_btree_delete(struct m0_be_btree *tree,
...
+	M0_PRE(tree->bb_root != NULL && tree->bb_ops != NULL);
+	M0_PRE(key->b_nob == be_btree_ksize(tree, key->b_addr));
+	M0_PRE(ver != 0);
+
+	bpv_tbs_set(&bpv);
+	bpv_ver_set(&bpv, ver);
+	M0_ASSERT(bpv_invariant(&bpv));
+
+	/*
+	 * TODO: Is it fine to use BAP_NORMAL here?
```
just-now, line 2165: same zone as inside m0_be_btree_insert() should work OK

be/btree.c
```diff
@@ -59,6 +59,104 @@ M0_INTERNAL const struct m0_fid_type m0_btree_fid_type = {
...
+		_0C(bpv_ver_get(bpv) >= BPV_VER_MIN));
+}
+
+static inline void bpv_ver_set(struct btree_pair_ver *bpv, uint64_t ver)
+{
+	M0_PRE(ver >= BPV_VER_MIN);
+	M0_PRE(ver <= BPV_VER_MAX);
+	bpv->bpv_ver = (bpv->bpv_ver & BPV_TBS) | ver;
+	M0_POST(bpv_invariant(bpv));
+	M0_POST(bpv_ver_get(bpv) == ver);
```
just-now, line 126: Do these two look to be ambiguous?

be/btree.c
```diff
@@ -2138,37 +2327,53 @@ M0_INTERNAL void m0_be_btree_save_inplace(struct m0_be_btree        *tree,
...
+
 	M0_ENTRY("tree=%p zonemask=%"PRIx64, tree, zonemask);
 	M0_PRE(tree->bb_root != NULL && tree->bb_ops != NULL);
 	M0_PRE(key->b_nob == be_btree_ksize(tree, key->b_addr));
 
-	btree_save(tree, tx, op, key, NULL, anchor, overwrite ?
+	if (ver != 0) {
+		bpv_tbs_clear(&bpv);
+		bpv_ver_set(&bpv, ver);
+		M0_ASSERT(bpv_invariant(&bpv));
```
just-now, line 2344: why do we need to call this invariant unless we obtained the metadata from persistent storage?
bpv lives inside the stack memory.

cas/ctg_store.c
```diff
@@ -997,28 +1001,53 @@ static int ctg_op_exec(struct m0_ctg_op *ctg_op, int next_phase)
...
+	bool                       use_ver = false;
 
 	zones = M0_BITS(M0_BAP_NORMAL) |
 		((ctg_op->co_flags & COF_RESERVE) ? M0_BITS(M0_BAP_REPAIR) : 0);
 
+	/*
+	 * Use the versioned btree API for ordinary CAS operations while
+	 * working in DTM0 environment.
+	 */
+	use_ver = ENABLE_DTM0 && ctg_is_ordinary(ctg_op->co_ctg);
```
just-now, line 1018: tis condition can be rewritten w.r.t. our oral disussion.

cas/ctg_store.c
```diff
@@ -1028,27 +1057,57 @@ static int ctg_op_exec(struct m0_ctg_op *ctg_op, int next_phase)
...
+		if (use_ver) {
+			m0_be_btree_lookup_alive_inplace(btree, beop, key,
+							 anchor);
+			break;
+		}
+		/* XXX: sneaky fallthrough */
 	case CTG_OP_COMBINE(CO_GET, CT_META):
+		M0_ASSERT(!use_ver);
 		m0_be_btree_lookup_inplace(btree, beop, key, anchor);
 		break;
```
just-now, line 1069: Could you put a comment here saying that CT_BTREE-s will be versioned only.

just-now: it took ~30seconds to understand the context.

cas/ctg_store.c
```diff
@@ -1028,27 +1057,57 @@ static int ctg_op_exec(struct m0_ctg_op *ctg_op, int next_phase)
...
 	case CTG_OP_COMBINE(CO_DEL, CT_BTREE):
+		if (use_ver && ver != 0) {
+			m0_be_btree_kill(btree, tx, beop, key, ver);
+			break;
+		}
+		/* XXX: sneaky fallthrough */
 	case CTG_OP_COMBINE(CO_DEL, CT_META):
+		M0_ASSERT(!use_ver);
 		m0_be_btree_delete(btree, tx, beop, key);
 		break;
```
just-now, line 1091: same here.

just-now (changes requested):
1. Overall logic inside be/btree.c look OK.
2. Looks like CAS integration can be simpler w.r.t. putting version flag into btree structure and providing analysis there. If it's so, please remove CAS tests also.
3. btree tests can, as you have said can be more sophisticated, still let's skip this for now.

be/btree.c
ivan-alekhin: I have removed this line; only the alignment requirement is still there.

be/btree.c
ivan-alekhin: goold idea, added const.

be/btree.c
ivan-alekhin: Makes sense. I have removed the invariant.

cas/ctg_store.c
ivan-alekhin: yes, changed this thing as per our discussion

cas/ctg_store.c
ivan-alekhin: I added a bunch of pre-conditions instead.

nikitadanilov (commented):
A general question: why do we have to implement tombstones at btree level? Cannot this be done at cas level?

ivan-alekhin:
> A general question: why do we have to implement tombstones at btree level? Cannot this be done at cas level?

@nikitadanilov ,
We started out with btree because we thought that there may be some changes in the core functions. But as you see, from the patch, it appears that only `btree_save` requires modification: we are reusing the result of the `be_btree_search` call inside `btree_save` to compare the versions. But even this code can be shifted outside of btree: we could add a function to `m0_be_btree_kv_ops` that would compare the record version and the version associated with the key or value buffer provided by the user of `btree_save` (I'm not quite sure if it is really simple, but I believe it can be done without significant changes in btree code).

If you believe that it will be better to move this code to CAS, please let me know -- I'll be happy to put this into CAS. At first, it will probably help us to integrate this "versioned" logic with the new btree implementation, and at second it will reduce the amount of duplicated tests (I had to add tests into btree and into CAS; it would be better to have a single set of tests for versioned behavior).

just-now:
as it was decided in a talk with Nikita, please move tombstones logic out from the btree to CAS. cc @ivan-alekhin

ivan-alekhin:
Closing this PR in favor of https://github.com/Seagate/cortx-motr/pull/889

PR #577, opened 2021-04-14
EOS-19514 scripts: Verify lnet in config and prepare phases

atulsdeshmukh2312:
Signed-off-by: Atul Deshmukh <atul.deshmukh@seagate.com>

PR #675, opened 2021-05-21
EOS-19514: Add lnet check in config and in prepare just to verify (on tmain)

atulsdeshmukh2312:


atulsdeshmukh2312:
Since its same one as PR#577,  only thing is we used tmain branch for this PR.
Now we can use old PR https://github.com/Seagate/cortx-motr/pull/577
for merging to main.
Closing this PR.

PR #678, opened 2021-05-21
EOS-20816 fop/fom_generic: rename M0_FOPH_TXN_COMMIT_WAIT -> M0_FOPH_TXN_DONE_WAIT

max-seagate:
This fom phase waits for M0_BTS_DONE.


huanghua78:
Hello @max-seagate , please create a Jira to track this.
Or maybe, we can create a single Jira ticket to track all four PRs, if we combine these four PRs into one when we land them.

huanghua78 (approved):

max-seagate:
Created a Jira ticket for this PR and added it to the commit message.

> Or maybe, we can create a single Jira ticket to track all four PRs, if we combine these four PRs into one when we land them.

It would be against https://github.com/Seagate/cortx-motr/blob/main/doc/git-policy#L409, so I'm going to create 1 Jira ticket for each PR. But if you really want to skip following the git guidelines about organizing commits (separate commits refactoring/feature/bugfixes etc.), please let me know.

madhavemuri (commented):
I think txn_commit_wait is more generic and is followed mostly for transaction commit terminology, we don't need to change this name.

madhavemuri:
> I think txn_commit_wait is more generic and is followed mostly for transaction commit terminology, we don't need to change this name.

After reviewing https://github.com/Seagate/cortx-motr/pull/679 where we have introduced logged phase, it makes sense to rename commit_wait to txn_done.

madhavemuri (approved):

PR #679, opened 2021-05-21
EOS-20818 fop/fom_generic: add new M0_FOPH_TXN_LOGGED_WAIT phase

max-seagate:
It's added just before M0_FOPH_TXN_DONE_WAIT to be able to wait for BE
transaction to become logged.

Signed-off-by: Maksym Medvied <max.medved@seagate.com>

PR #680, opened 2021-05-21
EOS-20819 fop/fom_generic: post FDMI record when BE tx becomes LOGGED

max-seagate:
Previously it was done just before BE tx is closed. It might have led to
a situation when FDMI record is sent to FDMI plugin, but BE tx for this
FDMI record didn't become persistent.
Now FDMI record is posted only after BE tx becomes LOGGED.


max-seagate:
Please change base branch to main after #679 is landed.

fop/fom.h
```diff
@@ -875,6 +875,11 @@ M0_INTERNAL int m0_fom_addb2_init(struct m0_fom_type *type, uint64_t id);
  */
 M0_INTERNAL int m0_fom_fol_rec_add(struct m0_fom *fom);
 
+/*
```
huanghua78, line 878: Please use Doxygen style comments.
```
/**
 * some comments.
 */
```

fop/fom_generic.c
```diff
@@ -433,12 +452,30 @@ M0_INTERNAL int m0_fom_tx_logged_wait(struct m0_fom *fom)
...
+				/*
+				 * This reference is taken in fom_fol_rec_add().
+				 */
+				m0_be_tx_put(&fom->fo_tx.tx_betx);
+				/*
+				 * FDMI is reliable, so at this point there is
+				 * no way to handle the failure better than to
+				 * terminate the application.
+				 */
+				M0_ASSERT_INFO(rc == 0,
```
huanghua78, line 473: I am worrying about these assertions here.
What can cause this failure?
What can we do in this case?
Can we skip some FDMI records?

max-seagate: > What can cause this failure?

Already described in the comment above.

> What can we do in this case?

Already described in the comment above.

> Can we skip some FDMI records?

No, FDMI is reliable transport by design.

huanghua78: I don't think we must panic and stop the Motr services here.
FDMI is designed to be reliable, but if something bad happens, we need to process this failure, and restart the post again.

max-seagate: @huanghua78: please suggest a way to handle this error here.

fop/fom.h
max-seagate: Done.

fop/fom_generic.c
huanghua78: Motr itself should be "reliable" and "robust".
In FDMI source end, this post() is provide by plug-in from user.
If something wrong in a plug-in, we can report this error in IEM, and continue the service.

As an enterprise storage system, we should avoid panic the core software, especially in user provided code.

max-seagate: BE doesn't have rollback, and FDMI is reliable by design. If FDMI service is not able to encode a data structure into a preallocated buffer than we have a bug in the code. In such case we can't rollback the transaction and we can't skip sending FDMI record, so there is no other option except to crash the process. We have `M0_ASSERT()`s about assumptions all over the code, and I'm adding one more.

> In FDMI source end, this post() is provide by plug-in from user.

No, it's not. FDMI source sends FDMI record over the network to FDMI plugin, and this is what is provided by the user. Please take a look at documentation in https://github.com/Seagate/cortx-motr/blob/main/fdmi/fdmi.c.

> If something wrong in a plug-in, we can report this error in IEM, and continue the service.

The error condition here is on FDMI source side, not on FDMI plugin side. There is no way I could think of to handle error here with IEM. If you think otherwise please elaborate.

If you provide a way to handle the error here I would check if it conforms to the design and I would be happy to implement it. If there is no such way than `M0_ASSERT()` is the best from whatever's left.

huanghua78: >No, it's not. FDMI source sends FDMI record over the network to FDMI plugin, and this is what is provided by the user. Please take a look at documentation in https://github.com/Seagate/cortx-motr/blob/main/fdmi/fdmi.c.

As I asked before, what could cause the failure of this post()?
For example, if the BE tx may failure due to insufficient credit, then this is a bug.
If the network between the FDMI src and FDMI plug-in fails, is it possible to cause the failure?
So, in this case, this is not a bug. This is caused by some failure. We need to generate IEM alert messages, and so, this can be fixed.
If the FDMI plug-in side dies, is it possible to fail?


Or a more general question. If we are going to post a FDMI record from FDMI source to FDMI plug-in side, if there are some failures (network failures, or client error and not able to receive it), what should the Motr service do?
Panic and stop working?

max-seagate: The existing implementation always returns 0 and the caller couldn't handle failure, so I made #700 to make this function `void`.

> As I asked before, what could cause the failure of this post()?

Nothing currently.

> If the network between the FDMI src and FDMI plug-in fails, is it possible to cause the failure?
Not in this case, the current implementation simply posts an FDMI record to a queue.

> If the FDMI plug-in side dies, is it possible to fail?

No, handling of this error is supposed to happen asyncrhonously, far from this call.

> Or a more general question. If we are going to post a FDMI record from FDMI source to FDMI plug-in side, if there are some failures (network failures, or client error and not able to receive it), what should the Motr service do?
Panic and stop working?

FDMI service is supposed to resend indefinitely, until the receiving end tells that this message could be discarded. Otherwise FDMI wouldn't be a reliable transport.

fop/fom_generic.c
```diff
@@ -373,12 +373,12 @@ static int fom_failure(struct m0_fom *fom)
 	}
 	if (tx->tx_state == M0_DTX_INVALID)
 		/**
-		 * Fom phase is set to M0_FOPH_TXN_COMMIT here so that when
+		 * Fom phase is set to M0_FOPH_FOL_REC_ADD here so that when
```
madhavemuri, line 376: What is the purpose of  adding fol records even for failure cases ?

max-seagate: As comment says FOL record is not being added - `m0_fom_tick_generic()` would set the state to the next state and then it will transition to `M0_FOPH_QUEUE_REPLY`. I did this to state explicitly that there is no way to close BE tx and to not to add FOL record, because there is no state transition that allows that.

Let me add this to the comment so the next reader would know this.

max-seagate: Explained.

madhavemuri (approved):

PR #693, opened 2021-05-25
Eos 20316:Implementation of put key for btree

RadhaGulhane13:
-Handled scenario of key already exists while insertion


btree/btree.c
```diff
@@ -2359,7 +2360,7 @@ static struct m0_btree_oimpl *alloc(int height)
 /**
  * Checks if given segaddr is within segment boundaries.
 */
-static int address_is_valid(struct segaddr addr)
+static int address_in_segment(struct segaddr addr)
```
shashank-parulekar, line 2363: If we need to check if the address falls within segment boundaries then return bool with true or false indicating success or failure.

btree/btree.c
```diff
@@ -3733,6 +3755,9 @@ struct cb_data {
 
 static int btree_kv_put_cb(struct m0_btree_cb *cb, struct m0_btree_rec *rec)
 {
+	if (rec->r_flags == M0_BSC_KEY_EXISTS) {
+		return 0;
```
shashank-parulekar, line 3759: You need to return a non-zero value in case of error so that the caller can handle it.

btree/btree.c
RadhaGulhane13: okay

PR #700, opened 2021-05-26
EOS-20841 fdmi/source_dock: make M0_FDMI_SOURCE_POST_RECORD() void

max-seagate:
FDMI is a reliable transport by design, so there is no way for the
caller to handle M0_FDMI_SOURCE_POST_RECORD failure. The only
implementation that exists currently in fdmi/fol_fdmi_src always
returns 0. This patch shifts failure handing for
M0_FDMI_SOURCE_POST_RECORD() from the caller to FDMI source
implementations.


max-seagate:
Please change branch to main after #679 is landed.

huanghua78 (approved):

madhavemuri (approved):

PR #703, opened 2021-05-27
EOS-20316: Implement btree key-value operations kv-put

RadhaGulhane13:
Updated put function for btree:

- Handled scenario if key already exist while insertion
- added changes to support different values for ksize and vsize

### DIFF
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -862,8 +862,8 @@ struct node_type {
 	 * maximum key size */
 	int  (*nt_keysize)(const struct nd *node);
 
-	/** Returns size of the value of node. In case variable value size return
-	 * maximum value size. */
+	/** Returns size of the value of node. In case variable value size
+	 * return maximum value size. */
 	int  (*nt_valsize)(const struct nd *node);
 
 	/** Returns unique FID for this node */
@@ -2518,6 +2518,7 @@ static int m0_btree_put_alloc_phase(struct m0_btree_op *bop)
 	}
 }
 
+#if 0
 /**
  * This function is responsible to handle the overflow at node at particular
  * level. It will get called when given record is not able to fit in node. This
@@ -2712,6 +2713,8 @@ static int m0_btree_put_makespace_phase(struct m0_btree_op *bop)
 	}
 	return P_NEXTUP;
 }
+#endif
+
 
 /**
  * This function gets called when splitting is done at root node. This function
@@ -2720,7 +2723,7 @@ static int m0_btree_put_makespace_phase(struct m0_btree_op *bop)
  * @param bop it structure for btree operation which contains all required data
  * @return int it return state which needs to get executed next
  */
-static int m0_btree_put_root_split_handle(struct m0_btree_op *bop)
+static int m0_btree_put_root_split_handle(struct m0_btree_op *bop, struct m0_btree_rec *new_rec)
 {
 	struct td              *tree       = bop->bo_arbor->t_desc;
 	struct m0_btree_oimpl  *oi         = bop->bo_i;
@@ -2730,6 +2733,8 @@ static int m0_btree_put_root_split_handle(struct m0_btree_op *bop)
 	m0_bcount_t             vsize;
 	void                   *p_val;
 	struct m0_btree_rec     temp_rec;
+
+	bop->bo_rec   = *new_rec;
 
 	temp_rec.r_key.k_data = M0_BUFVEC_INIT_BUF(&p_key, &ksize);
 	temp_rec.r_val        = M0_BUFVEC_INIT_BUF(&p_val, &vsize);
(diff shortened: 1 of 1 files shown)


btree/btree.c
```diff
@@ -2503,12 +2558,13 @@ static int m0_btree_put_makespace_phase(struct m0_btree_op *bop)
 	 */
 
 	/* 1)Move some records from current node to new node */
-	node_move(curr_level->l_node, curr_level->l_alloc, D_LEFT, NR_EVEN,
-		  bop->bo_tx);
-
 	/* Update level for newly allocated */
+	curr_level->l_alloc->n_skip_inv = true;
```
NikhilBirgade, line 2562: Why `n_skip_inv` is needed ?

btree/btree.c
```diff
@@ -2619,7 +2677,9 @@ static int m0_btree_put_makespace_phase(struct m0_btree_op *bop)
 			       m0_vec_count(&l_slot.s_rec.r_key.k_data.ov_vec));
 		temp_rec.r_val.ov_buf[0] =  &(curr_level->l_alloc->n_addr);
 		m0_bufvec_copy(&bop->bo_rec.r_val,
-				&temp_rec.r_val, INTERNAL_VAL_SIZE);
+			       &temp_rec.r_val, INTERNAL_VAL_SIZE);
+		/* if we want to update vcount for slot */
```
NikhilBirgade, line 2681: Please remove the commented lines if not needed.

btree/btree.c
```diff
@@ -2647,6 +2707,8 @@ static int m0_btree_put_makespace_phase(struct m0_btree_op *bop)
 		temp_rec.r_val.ov_buf[0] =  &(curr_level->l_alloc->n_addr);
 		m0_bufvec_copy(&bop->bo_rec.r_val,
 				&temp_rec.r_val, INTERNAL_VAL_SIZE);
+		/* if we want to update vcount for slot */
```
NikhilBirgade, line 2710: Please remove the commented lines if not needed.

btree/btree.c
```diff
@@ -2809,19 +2883,26 @@ static int64_t btree_put_tick(struct m0_sm_op *smop)
 			}*/
 			oi->i_nop.no_node = NULL;
 
-			node_find(&node_slot, &bop->bo_rec.r_key);
+			key_exists = node_find(&node_slot, &bop->bo_rec.r_key);
```
NikhilBirgade, line 2886: you can use `oi->i_node_found` directly instead of `key_exists`.

btree/btree.c
```diff
@@ -3973,6 +4093,9 @@ struct cb_data {
 
 static int btree_kv_put_cb(struct m0_btree_cb *cb, struct m0_btree_rec *rec)
 {
+	if (rec->r_flags == M0_BSC_KEY_EXISTS) {
```
NikhilBirgade, line 4096: Move it below variable declaration.

btree/btree.c
```diff
@@ -3998,6 +4121,9 @@ static int btree_kv_put_cb(struct m0_btree_cb *cb, struct m0_btree_rec *rec)
 
 static int btree_kv_get_cb(struct m0_btree_cb *cb, struct m0_btree_rec *rec)
 {
+       if (rec->r_flags == M0_BSC_KEY_NOT_FOUND)
```
NikhilBirgade, line 4124: Move it below variable declaration.

btree/btree.c
```diff
@@ -4573,7 +4712,7 @@ struct m0_ut_suite btree_ut = {
 		{"basic_tree_op",         m0_btree_ut_basic_tree_oper},
 		{"basic_kv_ops",          m0_btree_ut_basic_kv_oper},
 		{"multi_stream_kv_op",    m0_btree_ut_multi_stream_kv_oper},
-		{"insert_rec",            m0_btree_ut_insert_record},
+		//{"insert_rec",            m0_btree_ut_insert_record},
```
NikhilBirgade, line 4715: commenting style.

btree/btree.c
RadhaGulhane13: with level >0, the invariant checks if level >0, then no. of records should be greater than 0, with newly allocated internal node, it will fail, and we can not set level of newly allocated node as 0 to make invariant successful, Moreover, when there is a situation where we want to move all the records from one internal node to other , the internal node will become empty and it will fail this check if level>0, rec>0, therefore in such scenario , we required to skip such invariant check
 as level is required for other invariant check , refer line no. 1970, ff_rec_is_valid

btree/btree.c
RadhaGulhane13: I have added comments intentionally, as in future , we might want to change the vcount for slot in case of variable key size , so we can refer this comments

btree/btree.c
RadhaGulhane13: I have added comments intentionally, as in future , we might want to change the vcount for slot in case of variable key size , so we can refer this comments

btree/btree.c
RadhaGulhane13: okay

btree/btree.c
RadhaGulhane13: okay

btree/btree.c
```diff
@@ -858,6 +858,14 @@ struct node_type {
 	/** Returns size of the node (as a shift value) */
 	int  (*nt_shift)(const struct nd *node);
 
+	/** Returns size of the key of node. In case of variable key size return
+	 * maximum key size */
+	int  (*nt_keysize)(const struct nd *node);
+
+	/** Returns size of the value of node. In case variable value size return
```
ydb242, line 865: Exceeding 80 characters.
Please check.

btree/btree.c
```diff
@@ -2870,6 +2951,18 @@ static int64_t btree_put_tick(struct m0_sm_op *smop)
 		/* Fall through to the next step i.e. P_MAKESPACE */
 	}
 	case P_MAKESPACE: {
+		if (oi->i_node_found) {
+			struct m0_btree_rec rec;
+			rec.r_flags = M0_BSC_KEY_EXISTS;
+			int rc = bop->bo_cb.c_act(&bop->bo_cb,
+							&rec);
```
ydb242, line 2958: Alignment

RadhaGulhane13: yes

shashank-parulekar (changes requested):

RadhaGulhane13:
Resolved all the issues related to put operation, tested code against all ut's:

- node_create_delete
- node_add_del_rec  
-  basic_tree_op
-  basic_kv_ops
- multi_stream_kv_op
- insert_rec

shashank-parulekar (changes requested):

btree/btree.c
```diff
@@ -858,6 +857,14 @@ struct node_type {
 	/** Returns size of the node (as a shift value) */
 	int  (*nt_shift)(const struct nd *node);
 
+	/** Returns size of the key of node. In case of variable key size return
+	 * maximum key size */
+	int  (*nt_keysize)(const struct nd *node);
```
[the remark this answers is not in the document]
RadhaGulhane13: okay

btree/btree.c
```diff
@@ -2231,7 +2275,10 @@ static void generic_move(struct nd *src, struct nd *tgt,
 #ifndef __KERNEL__
 static bool m0_btree_overflow_is_possible(const struct nd *node)
```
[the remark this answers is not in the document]
RadhaGulhane13: In case of variable key size, we need to consider max key, value size  and we can figured out in m0_btree_overflow_is_possible itself ,whether it is fixed format or variable format, For example, if it is fixed format we can get key size, val size from node, else we will consider max values for determine  overflow. If we take key size and value size as an parameter we need to write check every time if it is fixed format or not and then pass the key , value size accordingly.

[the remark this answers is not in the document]
shashank-parulekar: In fixed/variable format you know the ksize and vsize as provided to the put function. It is difficult to add a concept of max key/value-size for variable key/value since we never know what that value could be. So it is better to get the key size and value size as provided in the rec structure parameter to put function.

NikhilBirgade (approved):

btree/btree.c
```diff
@@ -2701,52 +2600,231 @@ static int m0_btree_put_root_split_handle(struct m0_btree_op *bop)
...
+					    curr_level->l_node, &new_rec,
+					    &tgt, bop->bo_tx);
+		tgt.s_rec = new_rec;
+		node_make(&tgt, bop->bo_tx);
+		tgt.s_rec = temp_rec_1;
+		node_rec(&tgt);
+		m0_bufvec_copy(&tgt.s_rec.r_key.k_data, &new_rec.r_key.k_data,
+			       m0_vec_count(&new_rec.r_key.k_data.ov_vec));
+		m0_bufvec_copy(&tgt.s_rec.r_val, &new_rec.r_val,
+				m0_vec_count(&new_rec.r_val.ov_vec));
```
ydb242, line 2811: Extra space in alignment.

btree/btree.c
```diff
@@ -2809,11 +2888,16 @@ static int64_t btree_put_tick(struct m0_sm_op *smop)
 			}*/
 			oi->i_nop.no_node = NULL;
 
-			node_find(&node_slot, &bop->bo_rec.r_key);
+			oi->i_key_found = node_find(&node_slot,
+						     &bop->bo_rec.r_key);
```
ydb242, line 2892: Extra space in alignment.

btree/btree.c
```diff
@@ -2902,41 +2986,32 @@ static int64_t btree_put_tick(struct m0_sm_op *smop)
...
 			.s_idx  = curr_level->l_idx
 		};
-		node_slot.s_rec.r_key.k_data.ov_vec.v_count = &ksize;
-		node_slot.s_rec.r_key.k_data.ov_buf = &p_key;
-
-		node_slot.s_rec.r_val.ov_vec.v_count = &vsize;
-		node_slot.s_rec.r_val.ov_buf = &p_val;
+		rec = &node_slot.s_rec;
+		rec->r_key.k_data =  M0_BUFVEC_INIT_BUF(&p_key, &ksize);
+		rec->r_val        =  M0_BUFVEC_INIT_BUF(&p_val,&vsize);
```
ydb242, line 2991: space after comma

btree/btree.c
```diff
@@ -2902,41 +2986,32 @@ static int64_t btree_put_tick(struct m0_sm_op *smop)
...
+		 * revert back the changes made on btree. Detailed
+		 * explination is provided at P_MAKESPACE stage.
+		 */
+		//rec = &node_slot.s_rec;
+		rec->r_flags = M0_BSC_SUCCESS;
+		int rc = bop->bo_cb.c_act(&bop->bo_cb, rec);
+		if (rc) {
+			/* handle if callback fail i.e undo make */
+			node_del(node_slot.s_node, node_slot.s_idx,
+					bop->bo_tx);
```
ydb242, line 3010: alignment

btree/btree.c
```diff
@@ -3275,13 +3351,17 @@ static int64_t btree_get_tick(struct m0_sm_op *smop)
 			lev->l_node = oi->i_nop.no_node;
 			node_slot.s_node = oi->i_nop.no_node;
 			lev->l_seq = lev->l_node->n_seq;
-			oi->i_node_found = node_find(&node_slot, 
+			oi->i_key_found = node_find(&node_slot,
 						     &bop->bo_rec.r_key);
```
ydb242, line 3355: extra space in alignment

btree/btree.c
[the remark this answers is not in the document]
RadhaGulhane13: This will work for leaf node, but for internal nodes, if there is splitting , key can be different from the key provided in rec structure, so in this case we need to consider maximum key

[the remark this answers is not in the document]
shashank-parulekar: Are we sure the key in the leaf node will be different than the key in internal node ? I thought the value would be different in both the node types.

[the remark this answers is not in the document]
RadhaGulhane13: Yes. Lets take an example, once we reach leaf node, and we find that there is an overflow based on given ksize, vsize from record structure provided by user, once we split the leaf node into left and right leaf node, the key which needs to get inserted at internal node will be the 1st key from right leaf node, so here the key which needs to get inserted at the parent node can be different from key provided by user.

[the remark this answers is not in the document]
shashank-parulekar: ok in that case when you are doing a split at the level below you can find the max key from all the available keys in the child node since one of them would be the key that would be placed in the parent node.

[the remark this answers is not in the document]
RadhaGulhane13: Yes we can do that, Currently , I haven't added any logic for variable key, value size nodes, as some api's may required and depending on implementation of variable key value format, logic may changed, For example, to find max key from node, we may need api which gives size of key, in fixed format , as there is no such things we do not want any such api, and this api is not implemented yet. So, this things needs to be discussed. I have changed m0_btree_overflow_is_possible routine and now it is taking key value size as a parameter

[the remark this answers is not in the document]
shashank-parulekar: I think that the code will become difficult since you need to go through every key and get its size. Let us see if there is another quick way to do this or for now you can fallback to your max KEY size. We will revisit this one again later, but you need to add a TODO comment and mention about getting the correct max key size at runtime.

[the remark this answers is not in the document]
RadhaGulhane13: okay I will add comments at line . no. 2925 where we are initializing key value size for variable node format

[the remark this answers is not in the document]
RadhaGulhane13: Added comments at line no. 2925

shashank-parulekar (approved):

ydb242 (approved):

PR #715, opened 2021-06-01
EOS-19810: avoidance of internal dix layout index lookups

imvenkip:
Added idx_pool_version_get() internal API to get index pool version
and layout type. It will be called when m0_entity_create() API is
triggered from consumers like s3, m0kv. It is applicable for create
index operation.

Added required changes to support pver and layout type which are
passed by S3. It is applicable for PUT, GET, NEXT and DEL KV operations
in md-path.

Signed-off-by: Venkateswarlu Payidimarry <venkateswarlu.payidimarry@seagate.com>

### DIFF
--- a/motr/client.c
+++ b/motr/client.c
@@ -264,6 +264,14 @@ m0__obj_instance(const struct m0_obj *obj)
 	return m0__entity_instance(&obj->ob_entity);
 }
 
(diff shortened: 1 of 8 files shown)


huanghua78 (approved):
Looks fine to me.

madhavemuri (approved):

PR #718, opened 2021-06-01
Transaction support in node functions

ydb242:
This is a draft pull request.
Please suggest additional changes or modifications.
Signed-off-by: Yash Bhamare <yash.bhamare@seagate.com>

### DIFF
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -1651,7 +1651,7 @@ static const struct node_type fixed_format;
 static int64_t mem_tree_get(struct node_op *op, struct segaddr *addr, int nxt)
 {
 	struct td *tree = NULL;
-	int        i     = 0;
+	int        i    = 0;
 	uint32_t   offset;
 
 	m0_rwlock_write_lock(&trees_lock);
@@ -2018,7 +2018,12 @@ static void ff_init(const struct nd *node, int shift, int ksize, int vsize,
 	h->ff_shift = shift;
 	h->ff_ksize = ksize;
 	h->ff_vsize = vsize;
-	//m0_be_tx_capture();
+	/**
+	 * ToDo: We need to capture the changes occuring in the header using
+	 * m0_be_tx_capture().
+	 * Capture only those fields where there is any updation instead of the
+	 * whole header.
+	 */
 }
 
 static void ff_fini(const struct nd *node)
@@ -2129,9 +2134,16 @@ static void ff_make(struct slot *slot, struct m0_be_tx *tx)
 	M0_PRE(ff_rec_is_valid(slot));
 	M0_PRE(ff_isfit(slot));
 	memmove(start + rsize, start, rsize * (h->ff_used - slot->s_idx));
-	//m0_be_tx_capture();
+	/**
+	 * ToDo: We need to capture the changes occuring in the memory whose 
+	 * address starts from "start + rsize" and has its respective size using
+	 * m0_be_tx_capture().
+	 */
 	h->ff_used++;
-	//m0_be_tx_capture();
+	/**
+	 * ToDo: We need to capture the changes occuring in the header's ff_used
+	 * field using m0_be_tx_capture().
+	 */
 }
 
 static bool ff_find(struct slot *slot, const struct m0_btree_key *find_key)
@@ -2196,9 +2208,16 @@ static void ff_del(const struct nd *node, int idx, struct m0_be_tx *tx)
 	M0_PRE(idx < h->ff_used);
 	M0_PRE(h->ff_used > 0);
 	memmove(start, start + rsize, rsize * (h->ff_used - idx - 1));
-	//m0_be_tx_capture();
+	/**
+	 * ToDo: We need to capture the changes occuring in the memory whose 
+	 * address starts from "start" and has its respective size using
+	 * m0_be_tx_capture().
+	 */
 	h->ff_used--;
-	//m0_be_tx_capture();
+	/**
+	 * ToDo: We need to capture the changes occuring in the header's ff_used
+	 * field using m0_be_tx_capture().
+	 */
 }
 
 static void ff_set_level(const struct nd *node, uint8_t new_level,
@@ -2207,7 +2226,10 @@ static void ff_set_level(const struct nd *node, uint8_t new_level,
 	struct ff_head *h = ff_data(node);
 
 	h->ff_level = new_level;
-	//m0_be_tx_capture();
+	/**
+	 * ToDo: We need to capture the changes occuring in the node-header's 
+	 * ff_level field using m0_be_tx_capture().
+	 */
 }
 
 static void generic_move(struct nd *src, struct nd *tgt,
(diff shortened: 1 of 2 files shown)


btree/btree.c
```diff
@@ -1072,7 +1073,8 @@ static void node_del  (const struct nd *node, int idx, struct m0_be_tx *tx);
 static void node_set_level  (const struct nd *node, uint8_t new_level);
 static void node_move (struct nd *src, struct nd *tgt,
 		       enum dir dir, int nr, struct m0_be_tx *tx);
-static void mem_update(void);
+static void mem_update(struct node_op *op, struct m0_be_tx *tx, void *ptr,
```
shashank-parulekar, line 1076: I see you had introduced this function. I am not sure why we need this one. Also you have added this function amongst the node_* functions. If this is specific to mem_* operations then this function needs to be moved to the proper location. Can you also please explain the intention behind the existence of this function ?

btree/btree.c
```diff
@@ -1143,12 +1145,15 @@ static uint32_t         trees_loaded = 0;
...
-static void node_init(struct node_op *n_op, int ksize, int vsize)
+static void node_init(struct node_op *n_op, int ksize, int vsize,
+		      struct m0_be_tx *tx)
 {
 	const struct node_type *n_type = n_op->no_node->n_type;
 
 	n_type->nt_init(n_op->no_node, segaddr_shift(&n_op->no_node->n_addr),
 			ksize, vsize);
+	if (tx != NULL)
+		mem_update(n_op, tx, n_op->no_node, sizeof *n_op->no_node);
```
shashank-parulekar, line 1156: Again it is unwise to call the mem_* functions directly from node functions. You can use the structure of function pointers to reach the correct functions.

btree/btree.c
ydb242: The reason I have added this mem_update() is that it will call our m0_be_tx_capture() function to write data to segment.

btree/btree.c
ydb242: mem_update is not among the mem_* functions.
It is just a generic name I gave.

ydb242: The function will be responsible to call the m0_be_tx_capture() from our new btree

btree/btree.c
```diff
@@ -3198,17 +3205,17 @@ int calc_shift(int value)
...
-					  BNT_FIXED_KEYSIZE_VARIABLE_VALUESIZE ?
-					  data->bt->ksize : -1);
-	int 			 v_size = data->nt->nt_id == BNT_FIXED_FORMAT ?
-					  data->bt->vsize :(data->nt->nt_id ==
-					  BNT_VARIABLE_KEYSIZE_FIXED_VALUESIZE ?
-					  data->bt->vsize : -1);
+	struct m0_btree_op    *bop = M0_AMB(bop, smop, bo_op);
+	struct m0_btree_oimpl *oi = bop->bo_i;
+	struct m0_btree_idata *data = &bop->b_data;
+	int		       k_size = data->nt->nt_id == BNT_FIXED_FORMAT ?
```
shashank-parulekar, line 3211: Tabs can only be present at the start of the line. Once you have non-tab character in a line then the rest of the indentation for that line uses only spaces.

btree/btree.c
```diff
@@ -3529,19 +3546,28 @@ void m0_btree_create(void *addr, int nob, const struct m0_btree_type *bt,
 		     const struct node_type *nt, struct m0_be_tx *tx,
 		     struct m0_btree_op *bop)
 {
-	bop->b_data.addr	= addr;
-	bop->b_data.num_bytes	= nob;
-	bop->b_data.bt		= bt;
-	bop->b_data.nt		= nt;
+	bop->b_data.addr      = addr;
```
shashank-parulekar, line 3549: Same comment as above for not using tabs.

btree/btree.c
shashank-parulekar: If the node functions are the ones which change the on-disk structures then they are the ones which need the tx pointer and also it is their responsibility to capture their modifications  in the tx. Apart from the node functions (including tree_create and tree_destroy) if any other functions make changes to on-disk structures then we need to discuss about them to see if we need to have some more node_* functions too. Currently I do not feel mem_update is necessary but we can talk about this and discuss.

ydb242: Removed mem_update and added commented line where m0_be_tx_capture() is expected to be called.

btree/btree.c
```diff
@@ -2028,7 +2018,7 @@ static void ff_init(const struct nd *node, int shift, int ksize, int vsize,
 	h->ff_shift = shift;
 	h->ff_ksize = ksize;
 	h->ff_vsize = vsize;
-	mem_update(node, tx, h, sizeof *h);
+	//m0_be_tx_capture();
```
shashank-parulekar, line 2021: Revert to C-style comments. Also write a TBD with a few words indicating that tx_capture needs to happen at this location. Do the same for all the other places where you think the transaction capture should happen.

btree/btree.c
```diff
@@ -2139,9 +2129,9 @@ static void ff_make(struct slot *slot, struct m0_be_tx *tx)
 	M0_PRE(ff_rec_is_valid(slot));
 	M0_PRE(ff_isfit(slot));
 	memmove(start + rsize, start, rsize * (h->ff_used - slot->s_idx));
-	mem_update(node, tx, start + rsize, sizeof *(start + rsize));
+	//m0_be_tx_capture();
```
shashank-parulekar, line 2132: In the comment mention that this capture should be for capturing the memory where the records have been modified.

btree/btree.c
```diff
@@ -2139,9 +2129,9 @@ static void ff_make(struct slot *slot, struct m0_be_tx *tx)
...
 	h->ff_used++;
-	mem_update(node, tx, h, sizeof *h);
+	//m0_be_tx_capture();
```
shashank-parulekar, line 2134: Mention here that this capture should note the node-header contents which have been modified.

btree/btree.c
```diff
@@ -2298,8 +2288,8 @@ static void generic_move(struct nd *src, struct nd *tgt,
 	}
 	node_fix(src, tx);
 	node_fix(tgt, tx);
-	mem_update(src, tx, src, sizeof *src);
-	mem_update(tgt, tx, tgt, sizeof *tgt);
+	//m0_be_tx_capture();
```
shashank-parulekar, line 2291: These captures need to be a bit smarter where only the modified memory from both the nodes should be captured. This might mean a few more local variables need to be added to identify the locations in the node which have changed.

btree/btree.c
```diff
@@ -4229,19 +4217,20 @@ static int btree_kv_del_cb(struct m0_btree_cb *cb, struct m0_btree_rec *rec)
...
+	struct m0_btree_op      b_op        = {};
+	struct m0_btree        *tree;
+	void                   *temp_node;
+	int                     i;
+	time_t                  curr_time;
+	struct m0_btree_cb      ut_cb;
+	uint64_t                first_key;
+	bool                    first_key_initialized = false;
+	struct m0_btree_op      kv_op                 = {};
+	const struct node_type *nt                   = &fixed_format;
```
shashank-parulekar: Align the '=' with the previous line.

shashank-parulekar (changes requested):

btree/btree.c
ydb242: Updated in the latest commit.

btree/btree.c
ydb242: Updated in the latest commit.

btree/btree.c
ydb242: Updated in the latest commit.

btree/btree.c
ydb242: Yes.

btree/btree.c
ydb242: Updated in the latest commit.

shashank-parulekar (approved):

RadhaGulhane13 (approved):

PR #723, opened 2021-06-02
EOS-20955: Motr panic during write operation.

gshipra:
EOS-20955: Motr panic during write operation.
Problem:
There is a situation in which S3 code can send
NULL data to m0_client_calculate_pi. There are checks and
asserts in motr code which fails as of today.
Solution:
Update the checks in motr code and allow
callers to pass NULL buffer vec in m0_client_calculate_pi.

Testing
Code compiled and newly written UT passed.
Tried sending NULL buf vec and it worked.

### DIFF
--- a/motr/client.c
+++ b/motr/client.c
@@ -859,9 +859,9 @@ M0_INTERNAL int m0_calculate_md5_inc_context(
 
 	M0_PRE(pi != NULL);
 	M0_PRE(curr_context != NULL);
-	M0_PRE(bvec != NULL);
-	M0_PRE(bvec->ov_vec.v_count != NULL);
-	M0_PRE(bvec->ov_buf != NULL);
+	M0_PRE(ergo(bvec != NULL && bvec->ov_vec.v_nr != 0,
+				bvec->ov_vec.v_count != NULL &&
+				bvec->ov_buf != NULL));
 
 	/* This call is for first data unit, need to initialize prev_context */
 	if (flag & M0_PI_CALC_UNIT_ZERO) {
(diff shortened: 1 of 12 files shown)


gshipra:
@sg-shankar @nileshgovande - please review the code.

huanghua78 (approved):

motr/client.h
```diff
@@ -917,6 +917,8 @@ enum
 
 enum m0_pi_calc_flag {
```
knrajnambiar76, line 918: consider a case where in user passes flag by ORing
ie flag = M0_PI_CALC_UNIT_ZERO | M0_PI_SKIP_CALC_FINAL
here the value of flag is 3 (enum doesn't have value 3)

motr/client.c
```diff
@@ -847,9 +847,9 @@ M0_INTERNAL int m0_calculate_md5_inc_context(
 
 	M0_PRE(pi != NULL);
 	M0_PRE(curr_context != NULL);
-	M0_PRE(bvec != NULL);
-	M0_PRE(bvec->ov_vec.v_count != NULL);
-	M0_PRE(bvec->ov_buf != NULL);
+	M0_PRE(ergo(bvec != NULL && bvec->ov_vec.v_nr != 0,
```
knrajnambiar76, line 850: if this means we dont assert when bvec->ov_vec.v_nr is 0 or bvec->ov_vec.v_count is 0 or bvec is 0 -- seems to be fine

motr/client.h
gshipra: flag = M0_PI_CALC_UNIT_ZERO | M0_PI_SKIP_CALC_FINAL
flag value = 3 or 11 in binary

M0_PI_CALC_UNIT_ZERO - 01
M0_PI_SKIP_CALC_FINAL - 10

11 & 01 - true
11 & 10 - true

We do not need to add another flag for 3.

motr/client.c
gshipra: This means if bvec is NULL or bvec->ov_vec.v_nr == 0, in that case bvec->ov_vec.v_count can be NULL, bvec->ov_buf can be NULL.

madhavemuri:
@yeshpal-jain-seagate : Is this PR still valid one?

r-wambui:
Closing this as per recommendations from @chandradharraval via Motr Support Teams channel.

PR #728, opened 2021-06-03
Handle corner cases in put operation and generic_move 

RadhaGulhane13:
Changes in put operation and generic_move function
-Handling some corner cases

### DIFF
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -2242,7 +2242,8 @@ static void generic_move(struct nd *src, struct nd *tgt,
 	while (true) {
 		if (nr == 0 || (nr == NR_EVEN &&
 			       (node_space(tgt) <= node_space(src))) ||
-			       (nr == NR_MAX && srcidx == -1))
+			       (nr == NR_MAX && (srcidx == -1 ||
+			       node_count_rec(src) == 0)))
 			break;
 
 		/** Get the record at src index in rec. */
@@ -2424,21 +2425,24 @@ static struct m0_btree_oimpl *level_alloc(int height)
 	return oi;
 }
 
-static void level_cleanup(struct m0_btree_oimpl *oi)
+static void level_cleanup(struct m0_btree_oimpl *oi,
+			  enum m0_btree_opcode bo_opc)
 {
 	int i;
-	for (i = 0; i < oi->i_used; ++i) {
+	for (i = 0; i <= oi->i_used; ++i) {
 		if (oi->i_level[i].l_node != NULL) {
 			node_put(oi->i_level[i].l_node);
 			oi->i_level[i].l_node = NULL;
 		}
-		if (oi->i_level[i].l_alloc != NULL) {
-			node_put(oi->i_level[i].l_alloc);
-			oi->i_level[i].l_alloc = NULL;
-		}
 	}
-	if (oi->i_extra_node != NULL) {
-		node_put(oi->i_extra_node);
+	if (bo_opc == M0_BO_PUT) {
+		for (i = 0; i <= oi->i_used; ++i) {
+			if (oi->i_level[i].l_alloc != NULL) {
+				oi->i_level[i].l_alloc = NULL;
+			}
+		}
+		if (oi->i_extra_node != NULL)
+			oi->i_extra_node = NULL;
 	}
 	m0_free(oi->i_level);
 	m0_free(oi);
@@ -2502,7 +2506,7 @@ static int64_t m0_btree_put_alloc_phase(struct m0_btree_op *bop)
 				return P_ALLOC;
 			} else {
 				node_op_fini(&oi->i_nop);
-				level_cleanup(oi);
+				level_cleanup(oi, bop->bo_opc);
 				return P_INIT;
 			}
 		}
@@ -2527,7 +2531,7 @@ static int64_t m0_btree_put_alloc_phase(struct m0_btree_op *bop)
 			return P_ALLOC;
 		} else {
 			node_op_fini(&oi->i_nop);
-			level_cleanup(oi);
+			level_cleanup(oi, bop->bo_opc);
 			return P_INIT;
 		}
 	}
@@ -2689,6 +2693,24 @@ static void m0_btree_put_split_and_find(struct nd *l_alloc , struct nd *l_node,
 
 	diff = m0_bufvec_cursor_cmp(&cur_1, &cur_2);
 	tgt->s_node = diff < 0 ? l_slot.s_node : r_slot.s_node;
+
+	/**
+	 * Corner case: If given record needs to be inseted at internal left
+	 * node and if the key of given record is greater than key at last index
+	 * of left record, initialised tgt->s_idx explicitly, as node_find will
+	 * not compare key with last indexed key.
+	 */
+	if (node_level(tgt->s_node) > 0 && tgt->s_node == l_slot.s_node) {
+		l_slot.s_idx = node_count(l_slot.s_node);
+		l_slot.s_rec = temp_rec;
+		node_key(&l_slot);
+		m0_bufvec_cursor_init(&cur_2, &r_slot.s_rec.r_key.k_data);
+		diff = m0_bufvec_cursor_cmp(&cur_1, &cur_2);
+		if (diff > 0) {
+			tgt->s_idx = node_count(l_slot.s_node) + 1;
+			return;
+		}
+	}
 
 	node_find(tgt, &rec->r_key);
 }
@@ -3030,10 +3052,12 @@ static int64_t btree_put_tick(struct m0_sm_op *smop)
 		return P_CLEANUP;
 		//return m0_sm_op_sub(&bop->bo_op, P_CLEANUP, P_DONE);
 	}
-	case P_CLEANUP:
-		level_cleanup(oi);
+	case P_CLEANUP: {
+		oi->i_used = bop->bo_arbor->t_height - 1;
+		level_cleanup(oi, bop->bo_opc);
 		return P_DONE;
 		//return m0_sm_op_ret(&bop->bo_op);
+	}
 	default:
 		M0_IMPOSSIBLE("Wrong state: %i", bop->bo_op.o_sm.sm_state);
 	};
@@ -3461,7 +3485,7 @@ static int64_t btree_get_tick(struct m0_sm_op *smop)
 		//return m0_sm_op_sub(&bop->bo_op, P_CLEANUP, P_DONE);
 	}
 	case P_CLEANUP:
-		level_cleanup(oi);
+		level_cleanup(oi, bop->bo_opc);
 		return P_DONE;
 		//return m0_sm_op_ret(&bop->bo_op);
 	default:
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -862,8 +862,8 @@ struct node_type {
 	 * maximum key size */
 	int  (*nt_keysize)(const struct nd *node);
 
-	/** Returns size of the value of node. In case variable value size return
-	 * maximum value size. */
+	/** Returns size of the value of node. In case variable value size
+	 * return maximum value size. */
 	int  (*nt_valsize)(const struct nd *node);
 
 	/** Returns unique FID for this node */
@@ -2518,6 +2518,7 @@ static int m0_btree_put_alloc_phase(struct m0_btree_op *bop)
 	}
 }
 
+#if 0
 /**
  * This function is responsible to handle the overflow at node at particular
  * level. It will get called when given record is not able to fit in node. This
@@ -2712,6 +2713,8 @@ static int m0_btree_put_makespace_phase(struct m0_btree_op *bop)
 	}
 	return P_NEXTUP;
 }
+#endif
+
 
 /**
  * This function gets called when splitting is done at root node. This function
@@ -2720,7 +2723,7 @@ static int m0_btree_put_makespace_phase(struct m0_btree_op *bop)
  * @param bop it structure for btree operation which contains all required data
  * @return int it return state which needs to get executed next
  */
-static int m0_btree_put_root_split_handle(struct m0_btree_op *bop)
+static int m0_btree_put_root_split_handle(struct m0_btree_op *bop, struct m0_btree_rec *new_rec)
 {
 	struct td              *tree       = bop->bo_arbor->t_desc;
 	struct m0_btree_oimpl  *oi         = bop->bo_i;
@@ -2730,6 +2733,8 @@ static int m0_btree_put_root_split_handle(struct m0_btree_op *bop)
 	m0_bcount_t             vsize;
 	void                   *p_val;
 	struct m0_btree_rec     temp_rec;
+
+	bop->bo_rec   = *new_rec;
 
 	temp_rec.r_key.k_data = M0_BUFVEC_INIT_BUF(&p_key, &ksize);
 	temp_rec.r_val        = M0_BUFVEC_INIT_BUF(&p_val, &vsize);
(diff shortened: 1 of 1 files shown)


btree/btree.c
```diff
@@ -2233,7 +2233,8 @@ static void generic_move(struct nd *src, struct nd *tgt,
 	while (true) {
 		if (nr == 0 || (nr == NR_EVEN &&
 			       (node_space(tgt) <= node_space(src))) ||
-			       (nr == NR_MAX && srcidx == -1))
+			       (nr == NR_MAX && (srcidx == -1 ||
+			       node_count_rec(src) == 0)))
```
RadhaGulhane13, line 2237: node_count_rec(src) == 0 is needed as If dir = D_RIGHT , and nr=NR_MAX and  if we move all records, so src will contain zero records and we need come out of the loop. Earlier, conditions were not satisfying this case

btree/btree.c
```diff
@@ -2681,6 +2685,23 @@ static void m0_btree_put_split_and_find(struct nd *l_alloc , struct nd *l_node,
...
+		l_slot.s_idx = node_count(l_slot.s_node);
+		l_slot.s_rec = temp_rec;
+		node_key(&l_slot);
+		m0_bufvec_cursor_init(&cur_2, &r_slot.s_rec.r_key.k_data);
+		diff = m0_bufvec_cursor_cmp(&cur_1, &cur_2);
+		if (diff > 0) {
+			tgt->s_idx = node_count(l_slot.s_node) + 1;
+			return;
+		}
+	}
```
RadhaGulhane13, line 2704: We need to compare the given record key with last indexed key from left node as node_find will not consider last indexed key for comparison in case of internal node . But in this case we are moving records from right node to left node, we might want to compare given record key with last indexed key from left node, as if it is greater than last indexed key from left node, we might want to add given record at the end of left node.
example, [10,20,30,40]node keys in internal  left node after moving records form right to left. our key is 50, we want index 4 as result but node_find will give 3 as result as it will not consider 40 for comparison as it is internal node

btree/btree.c
```diff
@@ -2681,6 +2685,23 @@ static void m0_btree_put_split_and_find(struct nd *l_alloc , struct nd *l_node,
...
+	 * Corner case: If given record needs to be inseted at internal left
+	 * node and if the key of given record is greater than key at last index
+	 * of left record, initialised tgt->s_idx explicitly, as node_find will
+	 * not compare key with last indexed key.
+	 */
+	if (node_level(tgt->s_node) > 0 && tgt->s_node == l_slot.s_node) {
+		l_slot.s_idx = node_count(l_slot.s_node);
+		l_slot.s_rec = temp_rec;
+		node_key(&l_slot);
+		m0_bufvec_cursor_init(&cur_2, &r_slot.s_rec.r_key.k_data);
```
shashank-parulekar, line 2698: The second parameter should be &l_slot.s_rec.r_key.k_data and not &r_slot.s_rec.r_key.k_data.

shashank-parulekar (changes requested):

btree/btree.c
RadhaGulhane13: yes right

btree/btree.c
```diff
@@ -2415,21 +2416,24 @@ static struct m0_btree_oimpl *level_alloc(int height)
...
-		if (oi->i_level[i].l_alloc != NULL) {
-			node_put(oi->i_level[i].l_alloc);
-			oi->i_level[i].l_alloc = NULL;
-		}
 	}
-	if (oi->i_extra_node != NULL) {
-		node_put(oi->i_extra_node);
+	if (bo_opc == M0_BO_PUT) {
+		for (i = 0; i <= oi->i_used; ++i) {
+			if (oi->i_level[i].l_alloc != NULL) {
```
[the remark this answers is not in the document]
RadhaGulhane13: okay I will update this

btree/btree.c
```diff
@@ -2415,21 +2433,25 @@ static struct m0_btree_oimpl *level_alloc(int height)
...
+				if(oi->i_level[i].l_alloc_flag)
```
shashank-parulekar, line 2449: This flag l_alloc_flag is set values NODE_USED or NODE_FREE which are both non-zero so this test will always be TRUE. So this check does not make sense.
Also there is no need to add a flag such as l_alloc_flag. Instead the l_alloc pointer is normall set to NULL by default. When any possible sibling nodes are allocated then this pointer will have a non-zero value. Once the assigned value of the node is connected to the tree then the l_alloc pointer can be reset to NULL (after releasing the descriptor) so that if this variable l_alloc has a non-zero value in level_cleanup we are assured that the allocated node was not used and hence need to be freed in the level_cleanup routine.

btree/btree.c
```diff
@@ -2900,27 +2941,47 @@ static int64_t btree_put_tick(struct m0_sm_op *smop)
...
+				ksize = node_keysize(lev->l_node);
+				vsize = node_valsize(lev->l_node);
+			} else {
+				/**
+				 * @todo : ksize, vsize for variable key:
+				 * get the correct max key size at runtime
+				 * */
+				ksize = MAX_KEY_SIZE;
+				vsize = MAX_VAL_SIZE;
+			}
```
shashank-parulekar, line 2967: Why don't we have 
ksize = m0_vec_count(&bop->bo_rec.r_key.k_data.ov_vec)
vsize = m0_vec_count(&bop->bo_rec.r_val.ov_vec)

btree/btree.c
```diff
@@ -2942,14 +3003,15 @@ static int64_t btree_put_tick(struct m0_sm_op *smop)
 				else
 					bop->bo_flags |= BOF_LOCKALL;
 			}
-			if (bop->bo_arbor->t_height <= tree->t_height) {
+			if (bop->bo_arbor->t_height < tree->t_height) {
```
shashank-parulekar, line 3006: If the height changes (increases or decreases) you can free all the allocations and start over. The assumption is that the height change is a rare occurrence.

shashank-parulekar (changes requested):

btree/btree.c
RadhaGulhane13: It will be work for fixed size format not for variable format , Example, for leaf node we can use m0_vec_count(&bop->bo_rec.r_key.k_data.ov_vec), m0_vec_count(&bop->bo_rec.r_val.ov_vec) to determine overflow condition.but for internal node key size will be different

btree/btree.c
RadhaGulhane13: Okay I will make changes

btree/btree.c
```diff
@@ -2415,21 +2433,25 @@ static struct m0_btree_oimpl *level_alloc(int height)
 	return oi;
 }
 
-static void level_cleanup(struct m0_btree_oimpl *oi)
+static void level_cleanup(struct m0_btree_oimpl *oi,
+			  enum m0_btree_opcode bo_opc)
```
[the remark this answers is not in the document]
RadhaGulhane13: Alignment is correct, git is not reflecting alignment correctly here

btree/btree.c
```diff
@@ -2792,23 +2791,28 @@ static int64_t m0_btree_put_makespace_phase(struct m0_btree_op *bop)
...
 	node_slot.s_rec = temp_rec;
 	node_key(&node_slot);
 	new_rec.r_key = node_slot.s_rec.r_key;
 
 	newvalue      = INTERNAL_NODE_VALUE_SIZE;
-	newv_ptr      = &(curr_level->l_alloc->n_addr);
+	newv_ptr      = &(lev->l_alloc->n_addr);
 	new_rec.r_val = M0_BUFVEC_INIT_BUF(&newv_ptr, &newvsize);
 
+
```
shashank-parulekar, line 2804: Remove extra blank lines.

shashank-parulekar (approved):

btree/btree.c
```diff
@@ -2621,21 +2637,27 @@ static int64_t m0_btree_put_root_split_handle(struct m0_btree_op *bop,
...
 	node_done(&node_slot, bop->bo_tx, true);
-	node_fix(curr_level->l_node, bop->bo_tx);
+	node_fix(lev->l_node, bop->bo_tx);
 
 	/* Increase height by one */
 	tree->t_height++;
 
+
+	node_put(lev->l_alloc);
+	lev->l_alloc = NULL;
```
NikhilBirgade, line 2647: Do we need to set l_alloc to NULL here?

btree/btree.c
```diff
@@ -2752,23 +2791,28 @@ static int64_t m0_btree_put_makespace_phase(struct m0_btree_op *bop)
...
 	temp_rec_1.r_val          = M0_BUFVEC_INIT_BUF(&p_val_1, &vsize_1);
 
 	for (i = oi->i_used - 1; i >= 0; i--) {
-		curr_level = &oi->i_level[i];
-		node_slot.s_node = curr_level->l_node;
-		node_slot.s_idx  = curr_level->l_idx;
+		node_put(lev->l_alloc);
+		lev->l_alloc = NULL;
+
+		lev = &oi->i_level[i];
```
NikhilBirgade, line 2813: Should `lev = &oi->i_level[i]` be at line 2810 ?

btree/btree.c
```diff
@@ -2902,29 +2947,47 @@ static int64_t btree_put_tick(struct m0_sm_op *smop)
...
+							   ksize, vsize))
+				break;
+			if (oi->i_used == 0) {
+				if (lev->l_alloc == NULL ||
+				    oi->i_extra_node == NULL)
+					alloc = true;
+				break;
+			} else if (lev->l_alloc == NULL) {
+				alloc = true;
+				break;
```
NikhilBirgade, line 2983: Do we really need a do-while loop. Here we are starting the loop from leaf level. If overflow is possible and l_alloc will always be NULL at the beginning so it will always exit from leaf node. In what condition this loop will traverse upwards ?

btree/btree.c
```diff
@@ -2944,14 +3007,14 @@ static int64_t btree_put_tick(struct m0_sm_op *smop)
...
+			if (bop->bo_arbor->t_height != tree->t_height) {
+				/* If height decreased or increased */
 				lock_op_unlock(tree);
-				return P_CLEANUP;
+				return m0_sm_op_sub(&bop->bo_op, P_CLEANUP,
+					            P_INIT);
 			} else {
-				/* If height decreased */
+				/* If height same */
+				bop->bo_arbor->t_height = tree->t_height;
```
NikhilBirgade, line 3017: If  height is same then no need of re-initializing t_height.

btree/btree.c
```diff
@@ -3012,21 +3076,23 @@ static int64_t btree_put_tick(struct m0_sm_op *smop)
...
+		node_fix(lev->l_node, bop->bo_tx);
 
 		lock_op_unlock(tree);
 		return P_CLEANUP;
 		//return m0_sm_op_sub(&bop->bo_op, P_CLEANUP, P_DONE);
 	}
-	case P_CLEANUP:
-		level_cleanup(oi);
+	case P_CLEANUP: {
+		oi->i_used = bop->bo_arbor->t_height - 1;
```
NikhilBirgade, line 3091: why this initialization?

btree/btree.c
```diff
@@ -3080,18 +3146,13 @@ static struct m0_sm_state_descr btree_states[P_NR] = {
 	[P_CHECK] = {
 		.sd_flags   = 0,
 		.sd_name    = "P_CHECK",
-		.sd_allowed = M0_BITS(P_MAKESPACE, P_CLEANUP, P_DOWN, P_ACT,
-				      P_INIT, P_NEXTUP),
+		.sd_allowed = M0_BITS(P_MAKESPACE, P_CLEANUP, P_DONE, P_DOWN,
```
NikhilBirgade, line 3149: I don't see a place where we are jumping from P_CHECK to P_DONE.

btree/btree.c
```diff
@@ -3080,18 +3146,13 @@ static struct m0_sm_state_descr btree_states[P_NR] = {
...
 	[P_MAKESPACE] = {
 		.sd_flags   = 0,
 		.sd_name    = "P_MAKESPACE",
-		.sd_allowed = M0_BITS(P_ACT, P_NEXTUP),
-	},
-	[P_NEXTUP] = {
-		.sd_flags   = 0,
-		.sd_name    = "P_NEXTUP",
-		.sd_allowed = M0_BITS(P_MAKESPACE, P_CLEANUP),
+		.sd_allowed = M0_BITS(P_CLEANUP, P_DONE, P_ACT),
```
NikhilBirgade, line 3155: Not jumping to P_DONE from P_MAKESPACE, this is not allowed state.

btree/btree.c
```diff
@@ -3137,14 +3198,15 @@ static struct m0_sm_trans_descr btree_trans[256] = {
...
+	{ "put-check-init", P_CHECK, P_INIT },
+	{ "put-makespace-cleanup", P_MAKESPACE, P_CLEANUP },
+	{ "put-makespace-done", P_MAKESPACE, P_DONE },
 	{ "put-makespace", P_MAKESPACE, P_ACT },
-	{ "put-makespace-resolve", P_MAKESPACE, P_NEXTUP },
-	{ "put-nextup", P_NEXTUP, P_MAKESPACE },
-	{ "put-nextup-done", P_NEXTUP, P_CLEANUP },
 	{ "put/get-act", P_ACT, P_CLEANUP },
 	{ "put/get-done", P_CLEANUP, P_DONE },
+	{ "put", P_CLEANUP, P_INIT },
```
NikhilBirgade, line 3209: Put some comment like put restart or something.

btree/btree.c
RadhaGulhane13: Yes , else it will get freed in level_cleanup

btree/btree.c
RadhaGulhane13: lev at line no. 2810, indicate the level for l_alloc , lev at line no.:2813 indicate it's parents level, once we done splitting, etc, and done with using l_alloc , we want to do node_put and make it as NULL and then we will go level about it and perform operation at it s parent

btree/btree.c
RadhaGulhane13: Yes

btree/btree.c
RadhaGulhane13: In some cases , we might be calling P_CLEANUP when oi->i_used is not bop->bo_arbor->t_height - 1 and we are at some internal node, But I can make some changes as well to avoid this initialization

btree/btree.c
```diff
@@ -2436,25 +2426,29 @@ static struct m0_btree_oimpl *level_alloc(int height)
...
-	if (bo_opc == M0_BO_PUT) {
-		for (i = 0; i <= oi->i_used; ++i) {
-			if (oi->i_level[i].l_alloc != NULL) {
-				if(oi->i_level[i].l_alloc_flag)
-					oi->i_level[i].l_alloc = NULL;
-			}
+		if (oi->i_level[i].l_alloc != NULL) {
+			node_put(oi->i_level[i].l_alloc);
+			oi->i_nop.no_opc = NOP_FREE;
+			node_free(&oi->i_nop, oi->i_level[i].l_alloc, tx, 0);
```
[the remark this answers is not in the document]
RadhaGulhane13: okay

btree/btree.c
RadhaGulhane13: I added it because if we use return m0_sm_op_sub(&bop->bo_op, P_CLEANUP, P_DONE); instead of P_CLEANUP, so I thought transition P_MAKESPACE to P_DONE will be needed.

btree/btree.c
RadhaGulhane13: do-while loop will help, when check failed but height of tree is same, and we start from P_DOWN , so here we can first start checking from leaf , first check if overflow possible, then if allocation is already done i.e. l_alloc != NULL, so, l_alloc will not be always null, it will be null when it is the first traversal but not in every traversal

NikhilBirgade (approved):

PR #747, opened 2021-06-09
EOS-20305: Implement code for m0_btree_open()

ydb242:
The following changes include the implementation for m0_btree_open().

### DIFF
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -3437,20 +3437,6 @@ int64_t btree_destroy_tick(struct m0_sm_op *smop)
 	}
 }
 
-/* Based on the flag get the next/previous sibling index. */
-static int sibling_index_get(int index, uint64_t flags, bool key_exists)
-{
-	if (flags & BOF_NEXT)
-		return key_exists ? ++index : index;
-	return --index;
-}
-
-/* Checks if the index is in the range of valid key range for node. */
-static bool index_is_valid(struct level *lev)
-{
-	return (lev->l_idx >= 0) && (lev->l_idx < node_count(lev->l_node));
-}
-
 /**
  * btree_open_tick function is used to traverse through different states to
  * facilitate the working of m0_btree_open().
@@ -3496,6 +3482,20 @@ int64_t btree_open_tick(struct m0_sm_op *smop)
 	default:
 		M0_IMPOSSIBLE("Wrong state: %i", bop->bo_op.o_sm.sm_state);
 	}
+}
+
+/* Based on the flag get the next/previous sibling index. */
+static int sibling_index_get(int index, uint64_t flags, bool key_exists)
+{
+	if (flags & BOF_NEXT)
+		return key_exists ? ++index : index;
+	return --index;
+}
+
+/* Checks if the index is in the range of valid key range for node. */
+static bool index_is_valid(struct level *lev)
+{
+	return (lev->l_idx >= 0) && (lev->l_idx < node_count(lev->l_node));
 }
 
 /**
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -3242,9 +3242,9 @@ int calc_shift(int value)
 
 int64_t btree_create_tick(struct m0_sm_op *smop)
 {
-	struct m0_btree_op     *bop   = M0_AMB(bop, smop, bo_op);
-	struct m0_btree_oimpl  *oi    = bop->bo_i;
-	struct m0_btree_idata  *data  = &bop->b_data;
+	struct m0_btree_op    *bop    = M0_AMB(bop, smop, bo_op);
+	struct m0_btree_oimpl *oi     = bop->bo_i;
+	struct m0_btree_idata *data   = &bop->b_data;
 	int                    k_size = data->bt->ksize == -1 ? MAX_KEY_SIZE :
 					data->bt->ksize;
 	int                    v_size = data->bt->vsize == -1 ? MAX_VAL_SIZE :
@@ -3322,6 +3322,13 @@ int64_t btree_destroy_tick(struct m0_sm_op *smop)
 	}
 }
 
+/**
+ * btree_open_tick function is used to traverse through different states to
+ * facilitate the working of m0_btree_open().
+ *
+ * @param smop Represents the state machine operation
+ * @return int64_t It returns the next state to be executed.
+ */
 int64_t btree_open_tick(struct m0_sm_op *smop)
 {
 	struct m0_btree_op    *bop  = M0_AMB(bop, smop, bo_op);
@@ -3337,6 +3344,7 @@ int64_t btree_open_tick(struct m0_sm_op *smop)
 			if (bop->bo_i == NULL)
 				return M0_ERR(-ENOMEM);
 			oi = bop->bo_i;
+
 			oi->i_nop.no_addr = segaddr_build(bop->b_data.addr, 
 					    calc_shift(bop->b_data.num_bytes));
 
@@ -3614,6 +3622,17 @@ int64_t btree_del_tick(struct m0_sm_op *smop)
 	}
 }
 
+/**
+ * m0_btree_open is the API which is used by motr to fill the m0_btree structure
+ * with the tree descriptor and other detials.
+ *
+ * @param addr It is the address of root node allocated by the caller of this
+ * function
+ * @param nob It is the size of root node.
+ * @param out Pointer of m0_btree which is to be filled with data
+ * @param bop It represents the structure containing all the relevant details
+ * for carrying out btree operations.
+ */
 int  m0_btree_open(void *addr, int nob, struct m0_btree **out, 
 		   struct m0_btree_op *bop)
 {
@@ -4141,16 +4160,17 @@ static void m0_btree_ut_basic_tree_oper(void)
 						 nt, tx, &b_op),
 				 &b_op.bo_sm_group, &b_op.bo_op_exec);
 
-	//m0_btree_close(b_op.bo_arbor);
+	/** m0_btree_close(b_op.bo_arbor); */
 
-	// M0_BTREE_OP_SYNC_WITH_RC(&b_op.bo_op,
-	// 			 m0_btree_open(temp_node, 1024, &btree, &b_op),
-	// 			 &b_op.bo_sm_group, &b_op.bo_op_exec);
+	/** 
+	 * M0_BTREE_OP_SYNC_WITH_RC(&b_op.bo_op,
+	 *			 m0_btree_open(temp_node, 1024, &btree, &b_op),
+	 * 			 &b_op.bo_sm_group, &b_op.bo_op_exec);
+	 */
 
-	//m0_btree_close(btree);
+	/** m0_btree_close(btree); */
 
-	//btree = b_op.bo_arbor;
-
+	/** btree = b_op.bo_arbor; */
 	M0_BTREE_OP_SYNC_WITH_RC(&b_op.bo_op, m0_btree_destroy(b_op.bo_arbor,
 				 &b_op), &b_op.bo_sm_group, &b_op.bo_op_exec);
 
@@ -4163,10 +4183,15 @@ static void m0_btree_ut_basic_tree_oper(void)
 	/** Now run some invalid cases */
 
 	/** Open a non-existent btree */
-	//m0_btree_open(invalid_addr, 1024, &btree, &b_op);
+	/** M0_BTREE_OP_SYNC_WITH_RC(&b_op.bo_op,
+	 *			     m0_btree_open(invalid_addr, 1024, &btree,
+	 *					   &b_op), &b_op.bo_sm_group,
+	 *			     &b_op.bo_op_exec);
+	 */
+
 
 	/** Close a non-existent btree */
-	//m0_btree_close(btree);
+	/** m0_btree_close(btree); */
 
 	/** Destroy a non-existent btree */
 	/**
@@ -4184,23 +4209,32 @@ static void m0_btree_ut_basic_tree_oper(void)
 				 &b_op.bo_sm_group, &b_op.bo_op_exec);
 
 	/** Close it */
-	//m0_btree_close(b_op.bo_arbor);
+	/** m0_btree_close(b_op.bo_arbor); */
 
 	/** Try closing again */
-	//m0_btree_close(b_op.bo_arbor);
+	/** m0_btree_close(b_op.bo_arbor); */
 
 	/** Re-open it */
-	//m0_btree_open(invalid_addr, 1024, &btree, &b_op);
+	/** M0_BTREE_OP_SYNC_WITH_RC(&b_op.bo_op,
+	 *			     m0_btree_open(invalid_addr, 1024, &btree,
+	 *					   &b_op), &b_op.bo_sm_group,
+	 *			     &b_op.bo_op_exec);
+	 */
+
 
 	/** Open it again */
-	//m0_btree_open(invalid_addr, 1024, &btree, &b_op);
+	/** m0_btree_open(invalid_addr, 1024, &btree, &b_op); */
 
 	/** Destory it */
 	M0_BTREE_OP_SYNC_WITH_RC(&b_op.bo_op, m0_btree_destroy(b_op.bo_arbor,
 				 &b_op), &b_op.bo_sm_group, &b_op.bo_op_exec);
 
 	/** Attempt to reopen the destroyed tree */
-	//m0_btree_open(invalid_addr, 1024, &btree, &b_op);
+	/** M0_BTREE_OP_SYNC_WITH_RC(&b_op.bo_op,
+	 *			     m0_btree_open(invalid_addr, 1024, &btree,
+	 *					   &b_op), &b_op.bo_sm_group,
+	 *			     &b_op.bo_op_exec);
+	 */
 	btree_ut_fini();
 
 }
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -3164,7 +3164,7 @@ static struct m0_sm_state_descr btree_states[P_NR] = {
 
 static struct m0_sm_trans_descr btree_trans[256] = {
 	{ "create-init", P_INIT,  P_ACT  },
-	{ "create-act",  P_ACT,   P_DOWN },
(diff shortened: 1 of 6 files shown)


ydb242:
Currently the call for m0_btree_open has been commented out in the basic_tree_op UT.
This is because m0_btree_close() has not been implemented yet due to which next m0_btree_create() might cause some issues.

btree/btree.c
```diff
@@ -3328,40 +3328,43 @@ int calc_shift(int value)
...
-			bop->bo_arbor = m0_alloc(sizeof *bop->bo_arbor);
-			if (bop->bo_arbor == NULL)
-				return M0_ERR(-ENOMEM);
-
-			oi->i_nop.no_addr = segaddr_build(data->addr,
-							  calc_shift(
-							  data->num_bytes));
-			return tree_get(&oi->i_nop, &oi->i_nop.no_addr, P_ACT);
-
+			if (!addr_is_aligned(data->addr)) {
```
shashank-parulekar, line 3347: This is a good check but I am worried that we might not have the luxury of having an aligned address passed to us. Let us keep it for now but we might be forced to remove this check in the future. Add some comment here indicating such a possibility.

btree/btree.c
```diff
@@ -3443,6 +3447,92 @@ static bool index_is_valid(struct level *lev)
...
+	struct m0_btree_op    *bop  = M0_AMB(bop, smop, bo_op);
+	struct td             *tree = NULL;
+	int                    i    = 0;
+	struct m0_btree_oimpl *oi   = bop->bo_i;
+	struct segaddr         curr_addr;
+
+	switch(bop->bo_op.o_sm.sm_state)
+	{
+	case P_INIT:
+		if (!addr_is_aligned(bop->b_data.addr)) {
```
shashank-parulekar, line 3468: This check is okay during create_tree but not needed here anymore.

btree/btree.c
```diff
@@ -3443,6 +3447,92 @@ static bool index_is_valid(struct level *lev)
...
+							m0_rwlock_write_unlock(
+							&tree->t_lock);
+							m0_rwlock_write_unlock(
+							&trees_lock);
+							return P_DONE;
+						}
+					}
+					m0_rwlock_write_unlock(&tree->t_lock);
+				}
+			}
```
shashank-parulekar, line 3514: Why are you doing this? This is a part of tree_get. If you need to make changes then enhance that function.

btree/btree.c
```diff
@@ -3443,6 +3447,92 @@ static bool index_is_valid(struct level *lev)
...
+
+			bop->bo_i = m0_alloc(sizeof *bop->bo_i);
```
shashank-parulekar, line 3516: You have not even assigned a tree descriptor if an existing tree descriptor pointing to the same root node is not found.

btree/btree.c
```diff
@@ -4395,7 +4505,7 @@ static void m0_btree_ut_basic_tree_oper(void)
 	struct m0_btree_op      b_op = {};
 	void                   *temp_node;
 	const struct node_type *nt = &fixed_format;
-
+	int                     rc;
 	/** Prepare transaction to capture tree operations. */
```
shashank-parulekar, line 4509: Add a blank line after variable declarations.

sm/sm.c
```diff
@@ -413,7 +413,8 @@ static void state_set(struct m0_sm *mach, int state, int32_t rc)
 {
 	const struct m0_sm_state_descr *sd;
 
-	mach->sm_rc = rc;
+	if (!(strcmp(mach->sm_conf->scf_name, "btree-conf") == 0))
+		mach->sm_rc = rc;
```
shashank-parulekar, line 417: No this is not correct. The state machine implementation should not be partial to any module (such as btree here).

shashank-parulekar (changes requested):

sm/sm.c
ydb242: Ok.
Then I am not sure how do we get the return code.
Because till this point, mach->sm_rc has our error code. After the assignment on line 417, our error code vanishes as rc = 0.

btree/btree.c
ydb242: Yes.
This is because since we wont be having any tree_descriptor, then node_get function has the functionality to get a free node descriptor when there isnt any tree descriptor.
We rely on that mechanism for getting a free descriptor.

btree/btree.c
ydb242: We need this part of tree_get so that we can identify if there is any existing tree_descriptor pointing to the node.
We dont need to make changes to tree_get. However, enhancement can be done to create a independent function containing this code which then can be used by both tree_get and this open_tick function.

btree/btree.c
ydb242: Ok.
I added it because we have few cases in UT for the same. Since, create and open kind of have similar functionality, I think it will be prudent to keep it in both places.

btree/btree.c
ydb242: Ok. I will update it.

btree/btree.c
ydb242: Ok. I will update it.

btree/btree.c
shashank-parulekar: Why don't you just use the tree_get function. I see it does what you need from it.

btree/btree.c
shashank-parulekar: As I said this handling is not correct. You need to use tree_get as it does what you need.

btree/btree.c
```diff
@@ -3328,40 +3328,43 @@ int calc_shift(int value)
...
-				return M0_ERR(-ENOMEM);
-
-			oi->i_nop.no_addr = segaddr_build(data->addr,
-							  calc_shift(
-							  data->num_bytes));
-			return tree_get(&oi->i_nop, &oi->i_nop.no_addr, P_ACT);
-
+			if (!addr_is_aligned(data->addr)) {
+				return M0_ERR(-EFAULT);
+			} else {
```
NikhilBirgade, line 3349: No need of `else`, you are already returning if address is not aligned.

btree/btree.c
NikhilBirgade, line 3347: curly braces not required.

btree/btree.c
```diff
@@ -3328,40 +3328,43 @@ int calc_shift(int value)
...
+				bop->bo_i = m0_alloc(sizeof *bop->bo_i);
+				if (bop->bo_i == NULL)
+					return M0_ERR(-ENOMEM);
+				oi = bop->bo_i;
+				bop->bo_arbor = m0_alloc(sizeof *bop->bo_arbor);
+				if (bop->bo_arbor == NULL)
+					return M0_ERR(-ENOMEM);
```
NikhilBirgade, line 3356: Free up the memory for `bop->bo_i` as well for error condition.

btree/btree.c
```diff
@@ -3425,7 +3428,8 @@ int64_t btree_destroy_tick(struct m0_sm_op *smop)
 			return P_DONE;
 
 		default:
-			return 0;
+			M0_IMPOSSIBLE("Wrong state: %i",
+				      bop->bo_op.o_sm.sm_state);;
```
NikhilBirgade, line 3432: Two semicolons.

btree/btree.c
```diff
@@ -3328,40 +3328,43 @@ int calc_shift(int value)
...
+	struct m0_btree_oimpl *oi     = bop->bo_i;
+	struct m0_btree_idata *data   = &bop->b_data;
 	int                    k_size = data->bt->ksize == -1 ? MAX_KEY_SIZE :
 					data->bt->ksize;
 	int                    v_size = data->bt->vsize == -1 ? MAX_VAL_SIZE :
 					data->bt->vsize;
 
-
 	switch(bop->bo_op.o_sm.sm_state)
 	{
```
NikhilBirgade, line 3345: curly bracket should be on line 3344. Align rest of the code accordingly after the change.

btree/btree.c
```diff
@@ -3443,6 +3447,92 @@ static bool index_is_valid(struct level *lev)
...
+int64_t btree_open_tick(struct m0_sm_op *smop)
+{
+	struct m0_btree_op    *bop  = M0_AMB(bop, smop, bo_op);
+	struct td             *tree = NULL;
+	int                    i    = 0;
+	struct m0_btree_oimpl *oi   = bop->bo_i;
+	struct segaddr         curr_addr;
+
+	switch(bop->bo_op.o_sm.sm_state)
+	{
```
NikhilBirgade, line 3466: curly bracket should be on line 3465. Align rest of the code accordingly after the change.

btree/btree.c
```diff
@@ -3443,6 +3447,92 @@ static bool index_is_valid(struct level *lev)
...
+	case P_INIT:
+		if (!addr_is_aligned(bop->b_data.addr)) {
+			bop->bo_op.o_sm.sm_rc = M0_ERR(-EFAULT);
+			return P_DONE;
+		} else {
```
NikhilBirgade, line 3471: No need of else statement since you are returning on error condition on line 3471. This will also help in better alignment in rest of the code.

btree/btree.c
```diff
@@ -3328,40 +3328,43 @@ int calc_shift(int value)
...
-
-			oi->i_nop.no_addr = segaddr_build(data->addr,
-							  calc_shift(
-							  data->num_bytes));
-			return tree_get(&oi->i_nop, &oi->i_nop.no_addr, P_ACT);
-
+			if (!addr_is_aligned(data->addr)) {
+				return M0_ERR(-EFAULT);
+			} else {
+				bop->bo_i = m0_alloc(sizeof *bop->bo_i);
```
NikhilBirgade, line 3350: You have already initialised `oi` to `bop->bo_i`, you can use `oi` in place of `bop->bo_i` and remove line 3353.

btree/btree.c
```diff
@@ -3443,6 +3447,92 @@ static bool index_is_valid(struct level *lev)
...
+
+			m0_rwlock_write_lock(&trees_lock);
+			M0_ASSERT(trees_loaded <= ARRAY_SIZE(trees));
+
+			/**
+			 *  If existing allocated tree is found then return it
+			 *  after increasing the reference count and capturing
+			 *  the btree data.
+			 */
+			if (trees_loaded) {
```
NikhilBirgade, line 3485: Ideally `trees_loaded` should be non zero at this point. Either you can add an assert to check `trees_loaded`  is non-zero at this point and remove the `if (trees_loaded)` check. Or if you are expecting that `trees_loaded` can be zero at this point then add the `else` part handling.

btree/btree.c
NikhilBirgade: I concur `tree_get()` function is doing the same.

btree/btree.c
```diff
@@ -3443,6 +3447,92 @@ static bool index_is_valid(struct level *lev)
...
+						}
+					}
+					m0_rwlock_write_unlock(&tree->t_lock);
+				}
+			}
+
+			bop->bo_i = m0_alloc(sizeof *bop->bo_i);
+			if (bop->bo_i == NULL)
+				return M0_ERR(-ENOMEM);
+			oi = bop->bo_i;
```
NikhilBirgade, line 3519: Same comment as line 3350. Please be consistent either use `bop->bo_i` or oi.

btree/btree.c
```diff
@@ -3443,6 +3447,92 @@ static bool index_is_valid(struct level *lev)
...
+							= tree->t_type;
+
+							m0_rwlock_write_unlock(
+							&tree->t_lock);
+							m0_rwlock_write_unlock(
+							&trees_lock);
+							return P_DONE;
+						}
+					}
+					m0_rwlock_write_unlock(&tree->t_lock);
```
NikhilBirgade, line 3512: If you are not going inside the line 3489 `if check`, then you are not releasing `trees_lock` lock.

btree/btree.c
```diff
@@ -4411,21 +4521,23 @@ static void m0_btree_ut_basic_tree_oper(void)
...
-	m0_btree_open(temp_node, 1024, &btree);
-
 	m0_btree_close(btree);
 
-	btree = b_op.bo_arbor;
-
-	M0_BTREE_OP_SYNC_WITH_RC(&b_op.bo_op, m0_btree_destroy(btree, &b_op),
-				 &b_op.bo_sm_group, &b_op.bo_op_exec);
+	M0_BTREE_OP_SYNC_WITH_RC(&b_op.bo_op, m0_btree_destroy(b_op.bo_arbor,
+				 &b_op), &b_op.bo_sm_group, &b_op.bo_op_exec);
```
NikhilBirgade, line 4540: Alignment.

ydb242: Ok.
I will update it.

btree/btree.c
ydb242: Ok.

btree/btree.c
ydb242: Ok

btree/btree.c
ydb242: The error condition is for memory allocation failure. So there wont be any space allocated to bop->bo_arbor. Therefore, I think we dont need to free it.

btree/btree.c
ydb242: Ok.
I will update it.

btree/btree.c
ydb242: Ok.
I will update it.

btree/btree.c
ydb242: Ok.
I will make modifications to use tree_get()

btree/btree.c
ydb242: Ok.

btree/btree.c
ydb242: Ok.

btree/btree.c
ydb242: Oh Yes.
I will handle it.

btree/btree.c
ydb242: Ok.
I will update it.

btree/btree.c
ydb242: Ok.

btree/btree.c
ydb242: With tree_get(), this part wont be needed.
So it will be removed in the upcoming commit.

btree/btree.c
ydb242: With tree_get(), this part wont be needed.
So it will be removed in the upcoming commit.

Also, when the said tree_descriptor is freed using `tree_put()` it is the one which will release the trees_lock.

btree/btree.c
ydb242: Ok.

btree/btree.c
ydb242: Yes

NikhilBirgade (approved):

btree/btree.c
```diff
@@ -3328,104 +3328,112 @@ int calc_shift(int value)
...
-			if (bop->bo_arbor == NULL)
-				return M0_ERR(-ENOMEM);
-
-			oi->i_nop.no_addr = segaddr_build(data->addr,
-							  calc_shift(
-							  data->num_bytes));
-			return tree_get(&oi->i_nop, &oi->i_nop.no_addr, P_ACT);
+		oi = m0_alloc(sizeof *bop->bo_i);
+		if (oi == NULL)
+			return M0_ERR(-ENOMEM);
```
RadhaGulhane13, line 3357: instead of returning M0_ERR(-ENOMEM) you can set bop->bo_op.o_sm.sm_rc = M0_ERR(-ENOMEM), and return P_DONE. as *tick function should return next state which needs to get executed.
Similar changes for tree_destoy, open

btree/btree.c
```diff
@@ -3443,6 +3451,52 @@ static bool index_is_valid(struct level *lev)
 	return (lev->l_idx >= 0) && (lev->l_idx < node_count(lev->l_node));
 }
 
+/**
```
RadhaGulhane13, line 3454: shift line 3440 to 3453 after btree_open_tick as these functions are part of btree_get_tick

btree/btree.c
ydb242: No, We wont be doing that.
There are changes to the error code return mechanism.
So now all we need to do is return the error.
But I would suggest that you dont make any changes in your PR till these changes are tested thoroughly.

RadhaGulhane13: okay

btree/btree.c
ydb242: Updated in latest commit.

RadhaGulhane13 (approved):

btree/btree.c
shashank-parulekar: You might want to add a check for valid address in the node which could be based on node_is_valid or other things but not alignment of the node.

sm/sm.c
```diff
@@ -476,6 +476,12 @@ void m0_sm_state_set(struct m0_sm *mach, int state)
 	M0_PRE(m0_sm_invariant(mach));
 	state_set(mach, state, 0);
 }
+
+void m0_sm_state_and_rc_set(struct m0_sm *mach, int state, int32_t rc)
+{
+	M0_PRE(m0_sm_invariant(mach));
+	state_set(mach, state, rc);
+}
 M0_EXPORTED(m0_sm_state_set);
```
shashank-parulekar, line 485: You want to add a similar M0_EXPORTED line for the function m0_sm_state_and_rc_set

shashank-parulekar (changes requested):

btree/btree.c
ydb242: Ok
The issue that will arise is that if we remove this check of bad address for m0_btree_open(), then around 4 cases need to be commented out which were checking the invalid address case in m0_btree_ut_basic_tree_oper(). 
Also node_is_valid() check has not been implemented yet.

sm/sm.c
shashank-parulekar: As discussed use the patch and the change which I have requested in the PR for adding M0_EXPORTED attribute to the new sm function and that should be enough.

sm/sm.c
ydb242: Ok. I will update it.

shashank-parulekar (approved):

PR #755, opened 2021-06-13
Eos 20314: Add ut to test KV operations over multiple threads on multiple trees.

shashank-parulekar:
The code will run tree operations in parallel in multiple threads. Also the ut code will run KV operations in parallel on multiple threads over multiple trees.
In addition the support for slant in the ut is also added.

### DIFF
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -4535,9 +4535,12 @@ static int btree_kv_get_cb(struct m0_btree_cb *cb, struct m0_btree_rec *rec)
 	m0_bcount_t              vsize;
 	struct cb_data          *datum = cb->c_datum;
 
+	/** The caller can look at these flags if he needs to. */
+	datum->flags = rec->r_flags;
+
 	if (rec->r_flags == M0_BSC_KEY_NOT_FOUND ||
 	    rec->r_flags == M0_BSC_KEY_BTREE_BOUNDARY)
-     		return rec->r_flags;
+		return rec->r_flags;
 
 	ksize = m0_vec_count(&datum->key->k_data.ov_vec);
 	M0_PRE(m0_vec_count(&rec->r_key.k_data.ov_vec) <= ksize);
@@ -5055,6 +5058,8 @@ static void btree_ut_thread_kv_oper(struct btree_ut_thread_info *ti)
 		} else
 			key_end = ti->ti_key_last;
 
+		ut_cb.c_act        = btree_kv_put_cb;
+
 		/** PUT keys and their corresponding values in the tree. */
 		while (key_start <= key_end) {
 			/**
@@ -5119,7 +5124,7 @@ static void btree_ut_thread_kv_oper(struct btree_ut_thread_info *ti)
 					 &kv_op.bo_op_exec);
 		keys_found_count++;
 
-		while (get_data.flags != M0_BSC_KEY_NOT_FOUND) {
+		while (1) {
 			M0_BTREE_OP_SYNC_WITH_RC(&kv_op.bo_op,
 						 m0_btree_iter(tree,
 							       &rec.r_key,
@@ -5128,11 +5133,14 @@ static void btree_ut_thread_kv_oper(struct btree_ut_thread_info *ti)
 							       &kv_op),
 						 &kv_op.bo_sm_group,
 						 &kv_op.bo_op_exec);
+			if (get_data.flags == M0_BSC_KEY_BTREE_BOUNDARY)
+				break;
+
 			keys_found_count++;
 
 			/** Copy over the gotten key for the next search. */
-			for (i = 1; i < ARRAY_SIZE(key); i++)
-				key[i] = get_key[0];
+			for (i = 0; i < ARRAY_SIZE(key); i++)
+				key[i] = get_key[i];
 		}
 
 		M0_ASSERT(keys_found_count == keys_put_count);
@@ -5146,6 +5154,7 @@ static void btree_ut_thread_kv_oper(struct btree_ut_thread_info *ti)
 
 		key_start = key_iter;
 		del_key = (r % 2 == 0) ? key_start : key_end;
+		ut_cb.c_act        = btree_kv_del_cb;
 		while (keys_found_count) {
 			key[0] = (del_key << (sizeof(ti->ti_thread_id) * 8)) +
 				 ti->ti_thread_id;
@@ -5225,7 +5234,6 @@ static void btree_ut_num_threads_num_trees_kv_oper(uint32_t thread_count,
 {
 	int                          rc;
 	struct btree_ut_thread_info *ti;
-	int                          threads_to_run;
 	int                          i;
 	struct btree_ut_tree_info   *ut_trees;
 	uint16_t                     cpu;
@@ -5258,14 +5266,12 @@ static void btree_ut_num_threads_num_trees_kv_oper(uint32_t thread_count,
 	online_cpu_id_get(&cpuid_ptr, &cpu_count);
 
 	if (thread_count == 0)
-		threads_to_run = cpu_count -1; /** Skip Core-0 */
-	else
-		threads_to_run = thread_count;
+		thread_count = cpu_count -1; /** Skip Core-0 */
 
 	if (tree_count == 0)
-		tree_count = threads_to_run;
+		tree_count = thread_count;
 
-	M0_ASSERT(threads_to_run >= tree_count);
+	M0_ASSERT(thread_count >= tree_count);
 
 	thread_start = false;
 
@@ -5287,13 +5293,13 @@ static void btree_ut_num_threads_num_trees_kv_oper(uint32_t thread_count,
 		ut_trees[i].tree = b_op.bo_arbor;
 	}
 
-	M0_ALLOC_ARR(ti, threads_to_run);
+	M0_ALLOC_ARR(ti, thread_count);
 	M0_ASSERT(ti != NULL);
 
 	cpu_max = m0_processor_nr_max();
 
 	cpu = 1; /** We skip Core-0 for Linux kernel and other processes. */
-	for (i = 0; i < threads_to_run; i++) {
+	for (i = 0; i < thread_count; i++) {
 		rc = m0_bitmap_init(&ti[i].ti_cpu_map, cpu_max);
 		m0_bitmap_set(&ti[i].ti_cpu_map, cpuid_ptr[cpu], true);
 		cpu++;
@@ -5312,7 +5318,7 @@ static void btree_ut_num_threads_num_trees_kv_oper(uint32_t thread_count,
 		ti[i].ti_value_size = btree_type.vsize;
 	}
 
-	for (i = 0; i < threads_to_run; i++) {
+	for (i = 0; i < thread_count; i++) {
 		rc = M0_THREAD_INIT(&ti[i].ti_q, struct btree_ut_thread_info *,
 				    btree_ut_thread_init,
 				    &btree_ut_thread_kv_oper, &ti[i],
@@ -5323,7 +5329,7 @@ static void btree_ut_num_threads_num_trees_kv_oper(uint32_t thread_count,
 	/** Initialized all the threads by now. Let's get rolling ... */
 	thread_start = true;
 
-	for (i = 0; i < threads_to_run;i++) {
+	for (i = 0; i < thread_count;i++) {
 		m0_thread_join(&ti[i].ti_q);
 		m0_thread_fini(&ti[i].ti_q);
 	}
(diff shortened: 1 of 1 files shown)


btree/btree.c
```diff
@@ -5128,24 +5160,75 @@ static void btree_ut_thread_kv_oper(struct btree_ut_thread_info *ti)
...
+					  (sizeof(ti->ti_thread_id) * 8)) +
+					 ti->ti_thread_id;
+				key[0] = m0_byteorder_cpu_to_be64(key[0]);
+				for (i = 1; i < ARRAY_SIZE(key); i++)
+					key[i] = key[0];
+
+				M0_BTREE_OP_SYNC_WITH_RC(&kv_op.bo_op,
+							 m0_btree_get(tree,
+								      &rec.r_key,
+								      &ut_get_cb,
```
RadhaGulhane13, line 5201: 80 char limit exceed for line 5200, 5201. Please check

btree/btree.c
```diff
@@ -5258,14 +5343,12 @@ static void btree_ut_num_threads_num_trees_kv_oper(uint32_t thread_count,
 	online_cpu_id_get(&cpuid_ptr, &cpu_count);
 
 	if (thread_count == 0)
-		threads_to_run = cpu_count -1; /** Skip Core-0 */
-	else
-		threads_to_run = thread_count;
+		thread_count = cpu_count -1; /** Skip Core-0 */
```
RadhaGulhane13, line 5346: space before 1

btree/btree.c
```diff
@@ -5082,19 +5110,19 @@ static void btree_ut_thread_kv_oper(struct btree_ut_thread_info *ti)
 			M0_ASSERT(data.flags == 0);
```
RadhaGulhane13, line 5110: you might want to use M0_BSC_SUCCESS instead of 0 as you have used M0_BSC_SUCCESS in some other places as well and it gives more readability.

btree/btree.c
```diff
@@ -5082,19 +5110,19 @@ static void btree_ut_thread_kv_oper(struct btree_ut_thread_info *ti)
...
 
-		key_start = key_iter;
+		key_first = key_iter_start;
 		if (r % 2) {
 			/** Iterate forward. */
 			iter_dir = BOF_NEXT;
-			key[0] = (key_start <<
+			key[0] = (key_first <<
 				  (sizeof(ti->ti_thread_id) * 8)) +
 				 ti->ti_thread_id;
```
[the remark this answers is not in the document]
shashank-parulekar: The spacing is correct because the expression on line 5126 is used to '<<' key_first while the variable on line 5127 is added to the result.

btree/btree.c
```diff
@@ -5103,23 +5131,27 @@ static void btree_ut_thread_kv_oper(struct btree_ut_thread_info *ti)
 		} else {
 			/** Iterate backward. */
 			iter_dir = BOF_PREV;
-			key[0] = (key_end <<
+			key[0] = (key_last <<
 				  (sizeof(ti->ti_thread_id) * 8)) +
 				 ti->ti_thread_id;
```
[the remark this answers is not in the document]
shashank-parulekar: The spacing is correct because the expression on line 5126 is used to '<<' key_first while the variable on line 5127 is added to the result.

btree/btree.c
shashank-parulekar: Good catch.

btree/btree.c
shashank-parulekar: This is difficult for me to do because everything is shifted to the right. I was hoping no one would see it. Anyways let me try some changes.

btree/btree.c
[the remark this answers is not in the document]
RadhaGulhane13: okay

btree/btree.c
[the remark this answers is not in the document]
RadhaGulhane13: okay

btree/btree.c
```diff
@@ -5358,6 +5470,215 @@ static void m0_btree_ut_mt_mt_kv_oper(void)
...
+					      .ksize = sizeof(key),
+					      .vsize = sizeof(value),
+					     };
+	const struct node_type *nt         = &fixed_format;
+	struct m0_be_tx        *tx         = NULL;
+
+	random_r(&ti->ti_random_buf, &loop_count);
+	loop_count %= (MAX_TREE_LOOPS - MIN_TREE_LOOPS);
+	loop_count += MIN_TREE_LOOPS;
+
```
NikhilBirgade, line 5522: It will be good to add transaction apis (like you have added for KV oper handler), so that we do not miss it later.

btree/btree.c
```diff
@@ -5358,6 +5470,215 @@ static void m0_btree_ut_mt_mt_kv_oper(void)
...
+	for (i = 0; i < thread_count; i++) {
+		rc = M0_THREAD_INIT(&ti[i].ti_q, struct btree_ut_thread_info *,
+				    btree_ut_thread_init,
+				    &btree_ut_tree_oper_thread_handler, &ti[i],
+				    "Thread-%d", i);
+		M0_ASSERT(rc == 0);
+	}
+
+	/** Initialized all the threads. Now start the chaos ... */
+	thread_start = true;
```
NikhilBirgade, line 5661: Do you want to wait in the handler till all the threads are started? If yes, please add while check in handler or else you can remove this flag.

btree/btree.c
```diff
@@ -5214,6 +5308,8 @@ struct btree_ut_tree_info {
 	struct m0_btree *tree;
 };
 
+#define MAX_RECS_PER_THREAD  100000 /** Records per thread will work with */
```
NikhilBirgade, line 5311: Please define `MAX_RECS_PER_THREAD` and `btree_ut_tree_info` at one place where other UT related macros/functions are declared.

btree/btree.c
```diff
@@ -5358,6 +5470,215 @@ static void m0_btree_ut_mt_mt_kv_oper(void)
...
+		tree = b_op.bo_arbor;
+
+		random_r(&ti->ti_random_buf, &rec_count);
+		rec_count %= MAX_RECS_FOR_TREE_TEST;
+		rec_count = rec_count ? : (MAX_RECS_FOR_TREE_TEST / 2);
+
+		for (i = 1; i <= rec_count; i++) {
+			key   = i;
+			value = key;
+
```
NikhilBirgade, line 5560: can be replaced with `value = key = i`

btree/btree.c
```diff
@@ -5358,6 +5470,215 @@ static void m0_btree_ut_mt_mt_kv_oper(void)
...
+		m0_btree_open(temp_node, 1024, &tree);
+
+		for (i = 1; i <= rec_count; i++) {
+			key   = i;
+			value = key;
+
+			M0_BTREE_OP_SYNC_WITH_RC(&kv_op.bo_op,
+						 m0_btree_get(tree, &rec.r_key,
+							      &ut_cb, 0,
```
NikhilBirgade, line 5580: replace 0 with `BOF_EQUAL`

btree/btree.c
```diff
@@ -5358,6 +5470,215 @@ static void m0_btree_ut_mt_mt_kv_oper(void)
...
+							      &rec, &ut_cb,
+							      0, &kv_op),
+						 &kv_op.bo_sm_group,
+						 &kv_op.bo_op_exec);
+			M0_ASSERT(data.flags == M0_BSC_SUCCESS);
+		}
+
+		m0_btree_close(tree);
+
+		m0_btree_open(temp_node, 1024, &tree);
```
NikhilBirgade, line 5572: We can add return code check (ASSERT) for tree/open close.

btree/btree.c
```diff
@@ -5576,10 +5897,16 @@ struct m0_ut_suite btree_ut = {
...
+		{"single_thread_single_tree_kv_op",
+					m0_btree_ut_st_st_kv_oper},
 		{"multi_thread_single_tree_kv_op",
 					m0_btree_ut_mt_st_kv_oper},
 		{"multi_thread_multi_tree_kv_op",
 					m0_btree_ut_mt_mt_kv_oper},
+		{"single_thread_tree_op",
+					m0_btree_ut_st_tree_oper},
+		{"multi_thread_tree_op",
+					m0_btree_ut_mt_tree_oper},
```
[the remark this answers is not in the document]
shashank-parulekar: Will make changes.

btree/btree.c
shashank-parulekar: I am planning to add that as a separate task since that will require some more changes. I had just started them with no definite plan.

btree/btree.c
shashank-parulekar: There is no need to wait because when we come out of M0_THREAD_INIT the thread has already started running and it is waiting on this variable. If there are multiple threads (like we have in the loop) then all the threads are started and are waiting on this variable to change to 'true'.

btree/btree.c
shashank-parulekar: I have kept them here since they are used only by the functions defined below this line. If I move them up then there might be some confusion if we were to look just at the define's or structures.

btree/btree.c
shashank-parulekar: will do.

btree/btree.c
shashank-parulekar: Good catch.

btree/btree.c
shashank-parulekar: I will add for btree_open but cannot add for btree_close since it returns a void.

btree/btree.c
```diff
@@ -5906,16 +6013,252 @@ static void btree_ut_num_threads_num_trees_kv_oper(uint32_t thread_count,
...
+					      &kv_op.bo_op_exec);
+		M0_ASSERT(rc == 0);
+
+		for (i = 1; i <= rec_count; i++) {
+			value = key = i;
+
+			rc = M0_BTREE_OP_SYNC_WITH_RC(&kv_op.bo_op,
+						      m0_btree_get(tree,
+								   &rec.r_key,
+								   &ut_cb,
```
RadhaGulhane13, line 6149: you are using btree_kv_put_cb for callback you might want to use btree_kv_get_cb

btree/btree.c
```diff
@@ -5906,16 +6013,252 @@ static void btree_ut_num_threads_num_trees_kv_oper(uint32_t thread_count,
...
+
+		tree = b_op.bo_arbor;
+
+		random_r(&ti->ti_random_buf, &rec_count);
+		rec_count %= MAX_RECS_FOR_TREE_TEST;
+		rec_count = rec_count ? : (MAX_RECS_FOR_TREE_TEST / 2);
+
+		for (i = 1; i <= rec_count; i++) {
+			value = key = i;
+
```
RadhaGulhane13, line 6123: cpu to 64 bit conversion for key value

btree/btree.c
```diff
@@ -5906,16 +6013,252 @@ static void btree_ut_num_threads_num_trees_kv_oper(uint32_t thread_count,
...
+					      m0_btree_open(temp_node,
+							    1024, &tree,
+							    &kv_op),
+					      &kv_op.bo_sm_group,
+					      &kv_op.bo_op_exec);
+		M0_ASSERT(rc == 0);
+
+		for (i = 1; i <= rec_count; i++) {
+			value = key = i;
+
```
RadhaGulhane13, line 6145: cpu to 64 bit conversion for key value

btree/btree.c
```diff
@@ -5906,16 +6013,252 @@ static void btree_ut_num_threads_num_trees_kv_oper(uint32_t thread_count,
...
```
RadhaGulhane13, line 6169: cpu to 64 bit conversion for key value

btree/btree.c
```diff
@@ -5906,16 +6013,252 @@ static void btree_ut_num_threads_num_trees_kv_oper(uint32_t thread_count,
...
+			ut_cb.c_act = btree_kv_del_cb;
+
```
RadhaGulhane13, line 6171: if you rest callback for here, you might want to reset it at line no. 6123, as put,get,del operations are in loop, after resetting callback for del you have to reset it again for put opearation

btree/btree.c
shashank-parulekar: I have moved the defines in an enum and removed the struct btree_ut_tree_info as it is not needed.

btree/btree.c
```diff
@@ -5906,16 +6019,250 @@ static void btree_ut_num_threads_num_trees_kv_oper(uint32_t thread_count,
...
+					     };
+	const struct node_type *nt         = &fixed_format;
+	struct m0_be_tx        *tx         = NULL;
+	int                     rc;
+
+	random_r(&ti->ti_random_buf, &loop_count);
+	loop_count %= (MAX_TREE_LOOPS - MIN_TREE_LOOPS);
+	loop_count += MIN_TREE_LOOPS;
+
+	while(!thread_start)
```
NikhilBirgade, line 6085: space between while and (

btree/btree.c
RadhaGulhane13: okay

RadhaGulhane13 (approved):

NikhilBirgade (approved):

ydb242 (approved):
Looks good

PR #768, opened 2021-06-17
client: fix panic at nw_xfer_tioreq_get()

andriytk:
In case when target_ioreq_init() returns an error, the
target_ioreq is freed, but after this we still reference
to it, which might lead to panic.

Solution: return the error code immediately from
nw_xfer_tioreq_get() after freeing the target_ioreq.

### DIFF
--- a/motr/io_nw_xfer.c
+++ b/motr/io_nw_xfer.c
@@ -1169,11 +1169,12 @@ static int nw_xfer_tioreq_get(struct nw_xfer_request *xfer,
 		rc = target_ioreq_init(ti, xfer, fid, ta_obj, session, size);
 		if (rc == 0) {
 			tioreqht_htable_add(&xfer->nxr_tioreqs_hash, ti);
-			M0_LOG(M0_INFO, "New target_ioreq added for "FID_F,
-			       FID_P(fid));
-		}
-		else
+			M0_LOG(M0_INFO, "New target_ioreq %p added for "FID_F,
+			                ti, FID_P(fid));
+		} else {
 			m0_free(ti);
+			return M0_ERR_INFO(rc, "target_ioreq_init() failed");
+		}
 	}
 
 	if (ti->ti_dgvec == NULL && M0_IN(ioreq_sm_state(ioo),


motr/io_nw_xfer.c
```diff
@@ -1169,11 +1169,12 @@ static int nw_xfer_tioreq_get(struct nw_xfer_request *xfer,
...
 			tioreqht_htable_add(&xfer->nxr_tioreqs_hash, ti);
-			M0_LOG(M0_INFO, "New target_ioreq added for "FID_F,
-			       FID_P(fid));
-		}
-		else
+			M0_LOG(M0_INFO, "New target_ioreq %p added for "FID_F,
+			                ti, FID_P(fid));
+		} else {
 			m0_free(ti);
+			return M0_ERR_INFO(rc, "target_ioreq_init() failed");
```
madhavemuri, line 1176: It is not clear how this change is fixing the panic, as the error vlaue in `rc` returns from line 1186 as well. Can you elaborate more about it?

andriytk: @madhavemuri, see the few lines just before the return - https://github.com/Seagate/cortx-motr/blob/b64b1bba78571693a3123b462402736719632b1d/motr/io_nw_xfer.c#L1179-L1181

madhavemuri: You meant,  rc = dgmode_rwvec_alloc_init() is being called even if target_ioreq_init() fails, right ?

madhavemuri: Got it, ti should not be referenced at line:1179.

madhavemuri (approved):

motr/io_nw_xfer.c
andriytk: Yes, and `dgmode_rwvec_alloc_init(ti)` should not be called with it.

huanghua78 (approved):

huanghua78:
http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/Motr/525/

huanghua78:
> http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/Motr/525/

The result is not shown automatically here.
Please view the status from above link.
```
	Test Result (2 failures / ±0)
02motr-single-node.23m0d-signal-test.test
02motr-single-node.51kem.test
```
They are known issues.

mehjoshi (approved):

PR #770, opened 2021-06-17
EOS-21226: Fixed ut/sm.c

shashank-parulekar:
Changed the size of permissive_trans array from 256 elements to 225 elements to
make the ut run without crashing.

### DIFF
--- a/sm/ut/sm.c
+++ b/sm/ut/sm.c
@@ -1,6 +1,6 @@
 /* -*- C -*- */
 /*
- * Copyright (c) 2011-2020 Seagate Technology LLC and/or its Affiliates
+ * Copyright (c) 2011-2021 Seagate Technology LLC and/or its Affiliates
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


sm/ut/sm.c
```diff
@@ -628,9 +628,9 @@ static void ast_wait(void)
 	m0_mutex_fini(&wait_guard);
 }
 
-static struct m0_sm_state_descr permissive_states[16] = {};
+static struct m0_sm_state_descr permissive_states[15] = {};
```
NikhilBirgade, line 631: It will better to have a #define.

sm/ut/sm.c
```diff
@@ -628,9 +628,9 @@ static void ast_wait(void)
...
 
-static struct m0_sm_trans_descr permissive_trans[256] = {};
+static struct m0_sm_trans_descr permissive_trans[225] = {};
```
NikhilBirgade, line 633: It will better to have a #define.

ydb242 (approved):
Barring nikhil's comment, the PR looks good!

sm/ut/sm.c
shashank-parulekar: sure. let me use a define instead

sm/ut/sm.c
shashank-parulekar: sure. let me use a define instead

NikhilBirgade (approved):

RadhaGulhane13 (approved):

PR #779, opened 2021-06-21
go/mio: use new m0_layout_find_by_objsz() API

andriytk:
m0_layout_find_by_objsz() was introduced recently at commit 56d4b36, so use it directly instead of custom getOptimalUnitSz().


huanghua78 (approved):
Looks fine.

mehjoshi (approved):
Looks good to me.

PR #782, opened 2021-06-22
Eos 21219 : Changes related to LOCKS

RadhaGulhane13:
Changes related to memory free for struct m0_btree_oimpl *oi during each operation :
         -added one more phase : P_FINI

Changes related to locks:
            -added flag in oi as, oi structure will exist during entire operation
              due to above changes.
            -the flag will determine if thread has acquired the lock or not.

fixed bug in P_CHECK.

### DIFF
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -1178,6 +1178,9 @@ struct m0_btree_oimpl {
 
 	/** Node descriptor for cookie if it is going to be used. **/
 	struct nd      *i_cookie_node;
+
+	/** Flag for indicating if operation is aquiring lock or not. **/
+	bool            i_lock_aquired;
 };
 
 static struct td        trees[M0_TREE_COUNT];
@@ -2531,24 +2534,21 @@ static bool sibling_node_check(struct m0_btree_oimpl *oi)
 	return true;
 }
 
-static bool locked(struct td *tree)
-{
-	return false;
-}
-
 static int64_t lock_op_init(struct m0_sm_op *bo_op, struct node_op  *i_nop,
-			    struct td       *tree, int nxt)
+			    struct td *tree, struct m0_btree_oimpl *oi, int nxt)
 {
 	/** parameters which has passed but not will be used while state machine
 	 *  implementation for  locks
 	 */
 	m0_rwlock_write_lock(&tree->t_lock);
+	oi->i_lock_aquired = true;
 	return nxt;
 }
 
-static void lock_op_unlock(struct td *tree)
+static void lock_op_unlock(struct td *tree, struct m0_btree_oimpl *oi)
 {
 	m0_rwlock_write_unlock(&tree->t_lock);
+	oi->i_lock_aquired = false;
 }
 
 static void level_alloc(struct m0_btree_oimpl *oi, int height)
@@ -2792,7 +2792,7 @@ static int64_t btree_put_root_split_handle(struct m0_btree_op *bop,
 	node_put(oi->i_extra_node);
 	oi->i_extra_node = NULL;
 
-	lock_op_unlock(tree);
+	lock_op_unlock(tree, oi);
 	return P_CLEANUP;
 }
 
@@ -2927,7 +2927,7 @@ static int64_t btree_put_makespace_phase(struct m0_btree_op *bop)
 		node_fix(lev->l_node, bop->bo_tx);
 		node_move(lev->l_alloc, lev->l_node, D_RIGHT,
 		          NR_MAX, bop->bo_tx);
-		lock_op_unlock(bop->bo_arbor->t_desc);
+		lock_op_unlock(bop->bo_arbor->t_desc, oi);
 		return fail(bop, rc);
 	}
 	node_done(&tgt, bop->bo_tx, true);
@@ -2970,7 +2970,7 @@ static int64_t btree_put_makespace_phase(struct m0_btree_op *bop)
 			node_done(&node_slot, bop->bo_tx, true);
 			node_fix(lev->l_node, bop->bo_tx);
 
-			lock_op_unlock(bop->bo_arbor->t_desc);
+			lock_op_unlock(bop->bo_arbor->t_desc, oi);
 			return P_CLEANUP;
 		}
 
@@ -3009,7 +3009,7 @@ static int64_t btree_put_tick(struct m0_sm_op *smop)
 	struct level          *lev;
 
 	switch (bop->bo_op.o_sm.sm_state) {
-	case P_INIT: {
+	case P_INIT:
 		if (oi == NULL) {
 			bop->bo_i = m0_alloc(sizeof *oi);
 			if (bop->bo_i == NULL)
@@ -3020,7 +3020,6 @@ static int64_t btree_put_tick(struct m0_sm_op *smop)
 			return P_COOKIE;
 		else
 			return P_SETUP;
-	}
 	case P_COOKIE:
 		if (cookie_is_valid(tree, &bop->bo_rec.r_key.k_cookie) &&
 		    !node_isoverflow(oi->i_cookie_node))
@@ -3036,10 +3035,8 @@ static int64_t btree_put_tick(struct m0_sm_op *smop)
 		return P_LOCKALL;
 	}
 	case P_LOCKALL:
-		if (bop->bo_flags & BOF_LOCKALL) {
-			/* return m0_sm_op_sub(&bop->bo_op, P_LOCK, P_DOWN); */
-			m0_rwlock_write_lock(&tree->t_lock);
-		}
+		if (bop->bo_flags & BOF_LOCKALL)
+			return m0_sm_op_sub(&bop->bo_op, P_LOCK, P_DOWN);
 		/* Fall through to the next stage */
 	case P_DOWN:
 		oi->i_used = 0;
@@ -3119,9 +3116,9 @@ static int64_t btree_put_tick(struct m0_sm_op *smop)
 		return P_LOCK;
 	}
 	case P_LOCK:
-		if (!locked(tree))
+		if (!oi->i_lock_aquired)
 			return lock_op_init(&bop->bo_op, &bop->bo_i->i_nop,
-					    bop->bo_arbor->t_desc, P_CHECK);
+					    bop->bo_arbor->t_desc, oi, P_CHECK);
 		else
 			return P_CHECK;
 	case P_CHECK: {
@@ -3135,12 +3132,12 @@ static int64_t btree_put_tick(struct m0_sm_op *smop)
 			}
 			if (bop->bo_arbor->t_height != tree->t_height) {
 				/* If height decreased or increased */
-				lock_op_unlock(tree);
+				lock_op_unlock(tree, oi);
 				return m0_sm_op_sub(&bop->bo_op, P_CLEANUP,
 					            P_INIT);
 			} else {
 				/* If height same */
-				lock_op_unlock(tree);
+				lock_op_unlock(tree, oi);
 				return P_DOWN;
 			}
 		}
@@ -3152,10 +3149,10 @@ static int64_t btree_put_tick(struct m0_sm_op *smop)
 			rec.r_flags = M0_BSC_KEY_EXISTS;
 			int rc = bop->bo_cb.c_act(&bop->bo_cb, &rec);
 			if (rc) {
-				lock_op_unlock(tree);
+				lock_op_unlock(tree, oi);
 				return fail(bop, rc);
 			}
-			lock_op_unlock(tree);
+			lock_op_unlock(tree, oi);
 			return P_CLEANUP;
 		}
 
@@ -3205,13 +3202,13 @@ static int64_t btree_put_tick(struct m0_sm_op *smop)
 			node_del(node_slot.s_node, node_slot.s_idx, bop->bo_tx);
 			node_done(&node_slot, bop->bo_tx, true);
 			node_fix(lev->l_node, bop->bo_tx);
-			lock_op_unlock(tree);
+			lock_op_unlock(tree, oi);
 			return fail(bop, rc);
 		}
 		node_done(&node_slot, bop->bo_tx, true);
 		node_fix(lev->l_node, bop->bo_tx);
 
-		lock_op_unlock(tree);
+		lock_op_unlock(tree, oi);
 		return P_CLEANUP;
 	}
 	case P_CLEANUP: {
@@ -3219,11 +3216,10 @@ static int64_t btree_put_tick(struct m0_sm_op *smop)
 		return P_FINI;
 		/* return m0_sm_op_ret(&bop->bo_op); */
 	}
-	case P_FINI : {
+	case P_FINI :
 		if(oi)
 			m0_free(oi);
 		return P_DONE;
-	}
 	default:
 		M0_IMPOSSIBLE("Wrong state: %i", bop->bo_op.o_sm.sm_state);
 	};
@@ -3636,7 +3632,7 @@ static int64_t btree_get_tick(struct m0_sm_op *smop)
 	struct level          *lev;
 
 	switch (bop->bo_op.o_sm.sm_state) {
-	case P_INIT: {
+	case P_INIT:
 		if (oi == NULL) {
 			bop->bo_i = m0_alloc(sizeof *oi);
 			if (bop->bo_i == NULL)
@@ -3647,7 +3643,6 @@ static int64_t btree_get_tick(struct m0_sm_op *smop)
 			return P_COOKIE;
 		else
 			return P_SETUP;
-	}
 	case P_COOKIE:
 		if (cookie_is_valid(tree, &bop->bo_rec.r_key.k_cookie))
 			return P_LOCK;
@@ -3661,8 +3656,7 @@ static int64_t btree_get_tick(struct m0_sm_op *smop)
 		return P_LOCKALL;
 	case P_LOCKALL:
 		if (bop->bo_flags & BOF_LOCKALL)
-			/* return m0_sm_op_sub(&bop->bo_op, P_LOCK, P_DOWN); */
-			return P_LOCK;
+			return m0_sm_op_sub(&bop->bo_op, P_LOCK, P_DOWN);
 		/** Fall through if LOCKALL flag is not set. */
 	case P_DOWN:
 		oi->i_used = 0;
@@ -3701,9 +3695,9 @@ static int64_t btree_get_tick(struct m0_sm_op *smop)
 			return fail(bop, oi->i_nop.no_op.o_sm.sm_rc);
 		}
 	case P_LOCK:
-		if (!locked(tree))
+		if (!oi->i_lock_aquired)
 			return lock_op_init(&bop->bo_op, &bop->bo_i->i_nop,
-					    bop->bo_arbor->t_desc, P_CHECK);
+					    bop->bo_arbor->t_desc, oi, P_CHECK);
 		else
 			return P_CHECK;
 	case P_CHECK:
@@ -3717,12 +3711,12 @@ static int64_t btree_get_tick(struct m0_sm_op *smop)
 			}
 			if (bop->bo_arbor->t_height != tree->t_height) {
 				/* If height increased */
-				lock_op_unlock(tree);
+				lock_op_unlock(tree, oi);
 				return m0_sm_op_sub(&bop->bo_op, P_CLEANUP,
 				                    P_INIT);
 			} else {
 				/* If height decreased */
-				lock_op_unlock(tree);
+				lock_op_unlock(tree, oi);
 				return P_DOWN;
 			}
 		}
@@ -3771,7 +3765,7 @@ static int64_t btree_get_tick(struct m0_sm_op *smop)
 
 		bop->bo_cb.c_act(&bop->bo_cb, &s.s_rec);
 
-		lock_op_unlock(tree);
+		lock_op_unlock(tree, oi);
 		return P_CLEANUP;
 	}
 	case P_CLEANUP: {
@@ -3779,11 +3773,10 @@ static int64_t btree_get_tick(struct m0_sm_op *smop)
 		return P_FINI;
 		/* return m0_sm_op_ret(&bop->bo_op); */
 	}
-	case P_FINI : {
+	case P_FINI :
 		if(oi)
 			m0_free(oi);
 		return P_DONE;
-	}
 	default:
 		M0_IMPOSSIBLE("Wrong state: %i", bop->bo_op.o_sm.sm_state);
 	};
@@ -3962,9 +3955,9 @@ int64_t btree_iter_tick(struct m0_sm_op *smop)
 		}
 
 	case P_LOCK:
-		if (!locked(tree))
+		if (!oi->i_lock_aquired)
 			return lock_op_init(&bop->bo_op, &bop->bo_i->i_nop,
-					    bop->bo_arbor->t_desc, P_CHECK);
+					    bop->bo_arbor->t_desc, oi, P_CHECK);
 		else
 			return P_CHECK;
 	case P_CHECK:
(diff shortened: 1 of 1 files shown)


btree/btree.c
```diff
@@ -3995,11 +4012,15 @@ int64_t btree_iter_tick(struct m0_sm_op *smop)
...
 		bop->bo_cb.c_act(&bop->bo_cb, &s.s_rec);
-		lock_op_unlock(tree);
+		lock_op_unlock(tree, oi);
 		return m0_sm_op_sub(&bop->bo_op, P_CLEANUP, P_DONE);
 	}
 	case P_CLEANUP:
 		level_cleanup(oi, bop->bo_tx);
+		return P_FINI;
+	case P_FINI:
+		if(oi)
```
NikhilBirgade, line 4022: Remove if check. The code should make sure that we do not reach this state if memory is not allocated. If you want to catch bugs for this case then add M0_ASSERT.

btree/btree.c
```diff
@@ -3995,11 +4012,15 @@ int64_t btree_iter_tick(struct m0_sm_op *smop)
 			node_rec(&s);
 		}
 		bop->bo_cb.c_act(&bop->bo_cb, &s.s_rec);
-		lock_op_unlock(tree);
+		lock_op_unlock(tree, oi);
 		return m0_sm_op_sub(&bop->bo_op, P_CLEANUP, P_DONE);
 	}
 	case P_CLEANUP:
 		level_cleanup(oi, bop->bo_tx);
+		return P_FINI;
```
NikhilBirgade, line 4020: Keep the old code `return m0_sm_op_ret(&bop->bo_op);`

btree/btree.c
```diff
@@ -3995,11 +4012,15 @@ int64_t btree_iter_tick(struct m0_sm_op *smop)
 			node_rec(&s);
 		}
 		bop->bo_cb.c_act(&bop->bo_cb, &s.s_rec);
-		lock_op_unlock(tree);
+		lock_op_unlock(tree, oi);
 		return m0_sm_op_sub(&bop->bo_op, P_CLEANUP, P_DONE);
```
NikhilBirgade, line 4016: `return m0_sm_op_sub(&bop->bo_op, P_CLEANUP, P_FINI);`

btree/btree.c
```diff
@@ -3780,6 +3793,11 @@ int64_t btree_iter_tick(struct m0_sm_op *smop)
 
 	switch (bop->bo_op.o_sm.sm_state) {
 	case P_INIT:
+		if (oi == NULL) {
```
NikhilBirgade, line 3796: Remove if check. The code makes sure that this state is reached only once at the beginning.

btree/btree.c
```diff
@@ -2461,8 +2465,27 @@ static bool cookie_is_valid(struct td *tree, struct m0_bcookie *k_cookie)
...
+	oi->i_lock_aquired = false;
+}
+
 static int fail(struct m0_btree_op *bop, int rc)
 {
+	if (bop->bo_i->i_lock_aquired)
+		lock_op_unlock(bop->bo_arbor->t_desc, bop->bo_i);
 	bop->bo_op.o_sm.sm_rc = rc;
 	return P_CLEANUP;
 	/* return m0_sm_op_sub(&bop->bo_op, P_CLEANUP, P_DONE); */
```
NikhilBirgade, line 2491: When was this changed? Till the iteration operation implementation `return m0_sm_op_sub(&bop->bo_op, P_CLEANUP, P_DONE);`  was uncommented.

btree/btree.c
```diff
@@ -3209,20 +3203,24 @@ static int64_t btree_put_tick(struct m0_sm_op *smop)
...
 		return P_CLEANUP;
 	}
 	case P_CLEANUP: {
 		level_cleanup(oi, bop->bo_tx);
-		return P_DONE;
+		return P_FINI;
 		/* return m0_sm_op_ret(&bop->bo_op); */
 	}
+	case P_FINI :
+		if(oi)
```
ydb242, line 3221: space after if.

btree/btree.c
```diff
@@ -3758,13 +3766,18 @@ static int64_t btree_get_tick(struct m0_sm_op *smop)
...
 		return P_CLEANUP;
 	}
-	case P_CLEANUP:
+	case P_CLEANUP: {
 		level_cleanup(oi, bop->bo_tx);
+		return P_FINI;
+		/* return m0_sm_op_ret(&bop->bo_op); */
+	}
+	case P_FINI :
+		if(oi)
```
ydb242, line 3778: space after if.

btree/btree.c
```diff
@@ -4407,9 +4430,13 @@ static int64_t btree_del_tick(struct m0_sm_op *smop)
 	}
 	case P_CLEANUP : {
 		level_cleanup(oi, bop->bo_tx);
-		return P_DONE;
+		return P_FINI;
 		//return m0_sm_op_ret(&bop->bo_op);
 	}
+	case P_FINI :
+		if(oi)
```
ydb242, line 4437: space after if.

btree/btree.c
ydb242, line 4022: space after if.

btree/btree.c
NikhilBirgade, line 3778: Remove `if(oi`) check.

btree/btree.c
```diff
@@ -3758,13 +3766,18 @@ static int64_t btree_get_tick(struct m0_sm_op *smop)
...
 		bop->bo_cb.c_act(&bop->bo_cb, &s.s_rec);
 
-		lock_op_unlock(tree);
+		lock_op_unlock(tree, oi);
 		return P_CLEANUP;
 	}
-	case P_CLEANUP:
+	case P_CLEANUP: {
 		level_cleanup(oi, bop->bo_tx);
+		return P_FINI;
```
NikhilBirgade, line 3774: it should be `return m0_sm_op_ret(&bop->bo_op);`

btree/btree.c
```diff
@@ -3758,13 +3766,18 @@ static int64_t btree_get_tick(struct m0_sm_op *smop)
 
 		bop->bo_cb.c_act(&bop->bo_cb, &s.s_rec);
 
-		lock_op_unlock(tree);
+		lock_op_unlock(tree, oi);
 		return P_CLEANUP;
```
NikhilBirgade, line 3770: it should be `return m0_sm_op_sub(&bop->bo_op, P_CLEANUP, P_FINI);`

btree/btree.c
```diff
@@ -3688,28 +3696,28 @@ static int64_t btree_get_tick(struct m0_sm_op *smop)
...
+			} else {
+				oi->i_trial++;
+				lock_op_unlock(tree, oi);
 			}
 			if (bop->bo_arbor->t_height != tree->t_height) {
-				/* If height increased */
-				lock_op_unlock(tree);
+				/* If height has changed. */
 				return m0_sm_op_sub(&bop->bo_op, P_CLEANUP,
 				                    P_INIT);
```
NikhilBirgade, line 3718: change `P_INIT to P_SETUP`

btree/btree.c
```diff
@@ -3630,6 +3634,11 @@ static int64_t btree_get_tick(struct m0_sm_op *smop)
 
 	switch (bop->bo_op.o_sm.sm_state) {
 	case P_INIT:
+		if (oi == NULL) {
```
NikhilBirgade, line 3637: Remove `if` check.

btree/btree.c
RadhaGulhane13: we want to go to P_FINI, I think return m0_sm_op_ret(&bop->bo_op); should be in P_FINI not in P_CLEANUP

btree/btree.c
RadhaGulhane13: if (oi==NULL ) is required, because there are various scenarios where we start traversing again from P_INIT. For example if check fails and height has changed , so here we do not want to allocate memory again for oi.

btree/btree.c
RadhaGulhane13: I don't have idea, but I can change it to return m0_sm_op_sub(&bop->bo_op, P_CLEANUP, P_FINI)

btree/btree.c
RadhaGulhane13: you want to add return m0_sm_op_ret(&bop->bo_op); at P_CLENUP phase of P_FINI phase

btree/btree.c
RadhaGulhane13: then we won't be able to check cookies while traversing again

btree/btree.c
RadhaGulhane13: not required due to code changes.

btree/btree.c
RadhaGulhane13: not required due to code changes.

btree/btree.c
RadhaGulhane13: not required due to code changes.

btree/btree.c

btree/btree.c

btree/btree.c

btree/btree.c

btree/btree.c

btree/btree.c

btree/btree.c

btree/btree.c

btree/btree.c

btree/btree.c
```diff
@@ -2571,11 +2558,11 @@ static void level_cleanup(struct m0_btree_oimpl *oi,
 	int i;
 	for (i = 0; i <= oi->i_used; ++i) {
 		if (oi->i_level[i].l_node != NULL) {
-			node_put(oi->i_level[i].l_node);
+			node_put(oi->i_level[i].l_node, false);
```
shashank-parulekar, line 2561: level_cleanup is called from multiple places. It seems better if level_cleanup takes an extra parameter lock_acquired and passes it to the node_put routines. Also the callers of level_cleanup can pass the current state of the lock to level_cleanup properly.

btree/btree.c
```diff
@@ -2664,8 +2650,10 @@ static int64_t btree_put_alloc_phase(struct m0_btree_op *bop)
 			} else {
 				node_op_fini(&oi->i_nop);
 				oi->i_used = bop->bo_arbor->t_height - 1;
+				if (bop->bo_flags & BOF_LOCKALL)
+					lock_op_unlock(tree);
```
shashank-parulekar, line 2654: If this unlocking is done just for calling level_cleanup then as I mentioned above make sure to just pass the current lock state to level_cleanup instead of unlocking here.

btree/btree.c
```diff
@@ -2690,8 +2678,10 @@ static int64_t btree_put_alloc_phase(struct m0_btree_op *bop)
 		} else {
 			node_op_fini(&oi->i_nop);
 			oi->i_used = bop->bo_arbor->t_height - 1;
+			if (bop->bo_flags & BOF_LOCKALL)
+				lock_op_unlock(tree);
```
shashank-parulekar, line 2682: same comment as above.

shashank-parulekar (changes requested):

btree/btree.c
```diff
@@ -1795,7 +1798,7 @@ static int64_t mem_tree_get(struct node_op *op, struct segaddr *addr, int nxt)
 	tree->t_ref++;
 
 	if (addr) {
-		mem_node_get(op, tree, addr, nxt);
+		mem_node_get(op, tree, addr, true, nxt);
```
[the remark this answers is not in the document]
RadhaGulhane13: Because at line no. 1797 : m0_rwlock_write_lock(&tree->t_lock); we are already taking lock

btree/btree.c
RadhaGulhane13: We can do that, But current implementation ensures that level_cleanup will be called after unlocking . So we even if we passed parameter for lock_acquired, value for it is going to be false. Because level_cleanup will get called at P_CLEANUP phase, and there can be only two possibility i.e. we might be in LOCKALL  mode or non-LOCKALL, 
If we are not in LOCKALL mode, i.e. either we have completed our operation and we are in unlock mode, and even if we have not completed operation, we have to traversed again, but we will be unlock the lock and go for P_CLEANUP phase. i.e. we will come at P_CLEANUP when we are in unlock state.

Akso, if we are want to go for LOCKALL mode, we are traversing again after unlocking and then calling P_CLENUP and then P_SETUP or P_LOCKALL state (refer P_CHECK phase line no. :3128)

Moreover, 
level_cleanup(oi, bop->bo_tx);
return P_SETUP;
above steps can be return as m0_sm_op_sub(&bop->bo_op, P_CLEANUP, P_SETUP); as it provides more readability and suggested by NIkhil as well as he is also using the m0_sm_op_sub instead of calling level_cleanup() explicitly.

btree/btree.c
RadhaGulhane13: Need to see if we really want to add lock_acquired parameter for level_cleanup

btree/btree.c
RadhaGulhane13: Need to see if we really want to add lock_acquired parameter for level_cleanup

btree/btree.c
```diff
@@ -1547,8 +1549,9 @@ struct seg_ops {
 				     struct m0_be_tx *tx, int nxt);
 	void       (*so_tree_put)(struct td *tree);
 	int64_t    (*so_node_get)(struct node_op *op, struct td *tree,
-			          struct segaddr *addr, int nxt);
-	void       (*so_node_put)(struct nd *node);
+			          struct segaddr *addr, bool lock_aquired,
+				  int nxt);
+	void       (*so_node_put)(struct nd *node, bool lock_aquired);
```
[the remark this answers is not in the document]

btree/btree.c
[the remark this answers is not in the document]
RadhaGulhane13: need to pass false value for lock_aquired parameter according to latest changes in PR https://github.com/Seagate/cortx-motr/pull/776.

I will change this while merging

btree/btree.c
shashank-parulekar: I see that you are unlocking just before calling level_cleanup at some places if this unlocking is done just to make sure that level_cleanup is called in unlocked mode then that is not right as in future if we have a code which mistakenly does not unlock before calling level_cleanup then it will fail.

btree/btree.c
shashank-parulekar: Please discuss this with Yash and Nikhil before arriving at any conclusion.

shashank-parulekar (changes requested):

btree/btree.c
```diff
@@ -1043,8 +1044,8 @@ static int64_t tree_delete(struct node_op *op, struct td *tree,
 static void    tree_put   (struct td *tree);
 #ifndef __KERNEL__
 static int64_t    node_get  (struct node_op *op, struct td *tree,
-			     struct segaddr *addr, int nxt);
-static void       node_put  (struct nd *node);
+			     struct segaddr *addr, bool lock_aquired, int nxt);
+static void       node_put  (struct nd *node, bool lock_aquired);
```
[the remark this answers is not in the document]

btree/btree.c
RadhaGulhane13: The unlocking is not done just to make sure that level_cleanup is called in unlocked mode, it is done because there is no need to lock anymore.

Also, I have discussed this with Nikhil and Yash, they are fine with either way as   get and iter code,not calling level_cleanup in lock mode and same for put and delete

btree/btree.c
RadhaGulhane13: The purpose of unlocking at this stage is not to unlock because we are calling level_calling after that. But to make sure unlock the lock before new traversal. because thread will request for lock in LOCKALL

btree/btree.c
RadhaGulhane13: please refer above comment

btree/btree.c
```diff
@@ -3642,21 +3654,24 @@ static int64_t btree_get_tick(struct m0_sm_op *smop)
...
 	case P_LOCKALL:
 		if (bop->bo_flags & BOF_LOCKALL)
-			/* return m0_sm_op_sub(&bop->bo_op, P_LOCK, P_DOWN); */
-			return P_LOCK;
+			return lock_op_init(&bop->bo_op, &bop->bo_i->i_nop,
+				            bop->bo_arbor->t_desc, P_DOWN);
 		/** Fall through if LOCKALL flag is not set. */
-	case P_DOWN:
+	case P_DOWN: {
+		bool lock_acquired = bop->bo_flags & BOF_LOCKALL;
```
NikhilBirgade, line 3667: You don't have to assign lock_acquired at each level. Just define at the top of the function. Initially it will be zero. If it is updated in P_LOCKALL , it will be returned from state machine and then you will have updated lock_acquired next time onwards. Likewise in other state machines.

btree/btree.c
```diff
@@ -3679,38 +3694,38 @@ static int64_t btree_get_tick(struct m0_sm_op *smop)
...
 			node_op_fini(&oi->i_nop);
 			return fail(bop, oi->i_nop.no_op.o_sm.sm_rc);
 		}
 	case P_LOCK:
-		if (!locked(tree))
-			return lock_op_init(&bop->bo_op, &bop->bo_i->i_nop,
-					    bop->bo_arbor->t_desc, P_CHECK);
-		else
-			return P_CHECK;
+		return lock_op_init(&bop->bo_op, &bop->bo_i->i_nop,
```
NikhilBirgade, line 3707: If the lock is already acquired during P_LOCKALL, this will give you error. Likewise in other state machines.

btree/btree.c
```diff
@@ -3679,38 +3694,38 @@ static int64_t btree_get_tick(struct m0_sm_op *smop)
...
+			if (oi->i_trial >= MAX_TRIALS) {
+				if (bop->bo_flags & BOF_LOCKALL) {
+					lock_op_unlock(bop->bo_arbor->t_desc);
 					return fail(bop, -ETOOMANYREFS);
-				else
+				} else
 					bop->bo_flags |= BOF_LOCKALL;
-			}
+			} else
+				oi->i_trial++;
```
NikhilBirgade, line 3718: No need of else statement, you can add `oi->i_trial++ just after path_check on line 3711. Likewise in other statemachines.

btree/btree.c
```diff
@@ -2530,13 +2532,8 @@ static bool sibling_node_check(struct m0_btree_oimpl *oi)
...
-static bool locked(struct td *tree)
-{
-	return false;
-}
-
 static int64_t lock_op_init(struct m0_sm_op *bo_op, struct node_op  *i_nop,
-			    struct td       *tree, int nxt)
+			    struct td *tree, int nxt)
 {
 	/** parameters which has passed but not will be used while state machine
```
ydb242, line 2538: Is it necessary to return state while calling lock_op_init()?
As I see through the code, we are anyways falling through to the next state in every occasion.

btree/btree.c
RadhaGulhane13: we can have lock_acquired at top of function but we can not initialized with zero . because it is local for function ,even if we  update flag LOCKALL , it will be again zero for NEXTDOWN state. rather we can initialized flag with bop->bo_flags & BOF_LOCKALL

btree/btree.c
RadhaGulhane13: Yes Thanks for catching this. I will fix this

btree/btree.c
RadhaGulhane13: that can be done too.

btree/btree.c
RadhaGulhane13: In future if we implement separate  state machine for locks the return statement will be needed as we might get nxt as WAIT.

btree/btree.c
RadhaGulhane13: made respective changes

btree/btree.c

btree/btree.c

btree/btree.c
shashank-parulekar: Even if that is the case there needs to be a comment in level_cleanup to indicate that this function assumes the thread is unlocked when level_cleanup runs.

btree/btree.c
shashank-parulekar: I see that the check and unlock was added in this PR so it makes me assume that this unlock is happening to create the right environment before calling level_cleanup

btree/btree.c
RadhaGulhane13: okay I will add that

btree/btree.c
RadhaGulhane13: okay but this is not the case we are unlocking whenever we want to start traversing again or we are done with operation . and in both the scenarios we are calling  level_cleanup. thats why we are unlocking so indirectly level_cleanup gets called in unlock mode.

btree/btree.c
RadhaGulhane13: comment added

btree/btree.c
```diff
@@ -3263,33 +3272,33 @@ static int64_t btree_put_kv_tick(struct m0_sm_op *smop)
 		return P_LOCK;
 	}
 	case P_LOCK:
-		if (!locked(tree))
+		if (!lock_acquired)
 			return lock_op_init(&bop->bo_op, &bop->bo_i->i_nop,
 					    bop->bo_arbor->t_desc, P_CHECK);
 		else
 			return P_CHECK;
```
NikhilBirgade, line 3279: We can just fall down in instead of `else P_CHECK`.

btree/btree.c
```diff
@@ -1686,13 +1689,13 @@ static const struct node_type fixed_format;
...
 static int64_t node_get(struct node_op *op, struct td *tree,
-			struct segaddr *addr, int nxt)
+			struct segaddr *addr, bool lock_acquired, int nxt)
 {
 	int                     nxt_state;
 	const struct node_type *nt;
 	struct nd              *node;
 
-	nxt_state =  segops->so_node_get(op, tree, addr, nxt);
+	nxt_state =  segops->so_node_get(op, tree, addr, lock_acquired, nxt);
```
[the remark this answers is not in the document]

btree/btree.c
```diff
@@ -1882,7 +1886,7 @@ static int64_t mem_tree_get(struct node_op *op, struct segaddr *addr, int nxt)
 
 	if (addr) {
 		m0_rwlock_write_unlock(&tree->t_lock);
-		node_get(op, tree, addr, nxt);
+		node_get(op, tree, addr, true, nxt);
```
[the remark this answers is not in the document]
RadhaGulhane13: right

btree/btree.c
```diff
@@ -2690,32 +2688,30 @@ static void lock_op_unlock(struct td *tree)
 	m0_rwlock_write_unlock(&tree->t_lock);
 }
 
-static struct m0_btree_oimpl *level_alloc(int height)
+static void level_alloc(struct m0_btree_oimpl *oi, int height)
```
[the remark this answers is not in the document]
RadhaGulhane13: Even if we pass struct level it won't help, because then we anyway have to return struct level*,
Here I am passing oi as parameter to avoid returning level pointer, If we do not want to pass struct m0_btree_oimpl *oi,, then we have to return level pointer from function.

btree/btree.c
```diff
@@ -2804,8 +2800,10 @@ static int64_t btree_put_alloc_phase(struct m0_btree_op *bop)
 			} else {
 				node_op_fini(&oi->i_nop);
 				oi->i_used = bop->bo_arbor->t_height - 1;
+				if (bop->bo_flags & BOF_LOCKALL)
```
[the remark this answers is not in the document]

btree/btree.c
```diff
@@ -2830,8 +2829,10 @@ static int64_t btree_put_alloc_phase(struct m0_btree_op *bop)
 		} else {
 			node_op_fini(&oi->i_nop);
 			oi->i_used = bop->bo_arbor->t_height - 1;
+			if (bop->bo_flags & BOF_LOCKALL)
```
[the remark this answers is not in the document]

btree/btree.c
RadhaGulhane13: made changes, Also made respective changes in btree-conf. Please check.

NikhilBirgade (approved):

btree/btree.c
```diff
@@ -2704,32 +2701,30 @@ static void lock_op_unlock(struct td *tree)
...
 static void level_cleanup(struct m0_btree_oimpl *oi,
 			  struct m0_be_tx *tx)
 {
+	/**
+	 * This function assumes the thread is unlocked when level_cleanup runs.
+	 * As with current implementation, there is no such need to call
+	 * level_cleanup in lock mode. In future, if there is such need,
+	 * flag needs to be passed as parameter to this function indicating lock
+	 * is aquired or not by thread as this flag will be needed by node_put
+	 * operation to determine if lock is already aquired by thread.
```
shashank-parulekar, line 2718: Replace these lines with:
If ever there arises a need to call level_cleanup() with the lock owned by the calling thread then this routine will need some changes such as accepting a parameter which would tell us if the lock is already taken by this thread.

shashank-parulekar (approved):

btree/btree.c

ydb242 (approved):

PR #785, opened 2021-06-23
Eos 21213 : Implement sequence counter

RadhaGulhane13:


btree/btree.c
```diff
@@ -921,6 +921,12 @@ struct node_type {
 	/** Returns index of the record containing the key in the node */
 	bool (*nt_find) (struct slot *slot, const struct m0_btree_key *key);
 
+	/**
+	 * Updates the sequence counter to the current timestamp. This function
+	 * can be called whenever there is change in node.
+	 */
+	void (*nt_seq_cnt_update)  (struct nd *node);
+
```
shashank-parulekar, line 929: We do not need format specific code for sequence counter which is hosted within node-descriptor and can be updated directly by the node_*() functions directly.

shashank-parulekar (changes requested):

btree/btree.c
RadhaGulhane13: okay I will make related changes.

shashank-parulekar (approved):

NikhilBirgade (approved):

PR #790, opened 2021-06-24
EOS-21218: Node descriptor management

NikhilBirgade:

### DIFF
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -1829,8 +1829,10 @@ static int64_t node_get(struct node_op *op, struct td *tree,
 			 */
 
 			m0_rwlock_write_lock(&lru_lock);
-			/* M0_ASSERT_EX(ndlist_tlist_contains(&btree_lru_nds,
-			 *                                    op->no_node)); */
+			/**
+			 * M0_ASSERT_EX(ndlist_tlist_contains(&btree_lru_nds,
+			 *                                    op->no_node));
+			 */
 			ndlist_tlist_del(op->no_node);
 			m0_rwlock_write_unlock(&lru_lock);
 
@@ -1847,16 +1849,29 @@ static int64_t node_get(struct node_op *op, struct td *tree,
 		m0_rwlock_write_unlock(&op->no_node->n_lock);
 	} else {
 		/**
-		 * If node descriptor is not present allocate a new one
-		 * and assign to node.
-		 */
-		/**
-		 * TODO: Adding lru_lock to protect from multiple threads 
-		 * allocating more than one node descriptors for the same node. 
-		 * Replace it with a different global lock once hash 
+		 * TODO: Adding lru_lock to protect from multiple threads
+		 * allocating more than one node descriptors for the same node.
+		 * Replace it with a different global lock once hash
 		 * functionality is implemented.
 		 */
 		m0_rwlock_write_lock(&lru_lock);
+		/**
+		 * If node descriptor is already allocated for the node, no need
+		 * to allocate node descriptor again.
+		 */
+		op->no_node = nt->nt_opaque_get(addr);
+		if (op->no_node != NULL &&
+		    op->no_node->n_addr.as_core == addr->as_core) {
+			m0_rwlock_write_lock(&op->no_node->n_lock);
+			node_refcnt_update(op->no_node, true);
+			m0_rwlock_write_unlock(&op->no_node->n_lock);
+			m0_rwlock_write_unlock(&lru_lock);
+			return nxt_state;
+		}
+		/**
+		 * If node descriptor is not present allocate a new one
+		 * and assign to node.
+		 */
 		node = m0_alloc(sizeof *node);
 		/**
 		 * TODO: If Node-alloc fails, free up any node descriptor from
@@ -1888,8 +1903,10 @@ static int64_t node_get(struct node_op *op, struct td *tree,
  * the reference count reaches '0' then the node descriptor is moved to LRU
  * list.
  *
+ * @param op load operation to perform.
  * @param node node descriptor.
- * @param tree pointer to tree.
+ * @param lock_acquired true if lock is already taken else false.
+ * @param tx changes will be captured in this transaction.
  *
  */
 static void node_put(struct node_op *op, struct nd *node, bool lock_acquired,
@@ -2165,7 +2182,6 @@ static void mem_node_put(struct nd *node, bool lock_acquired)
 		 * active list and add to lru list
 		 */
 		tree_lock(node->n_tree, lock_acquired);
-		/* M0_ASSERT_EX(ndlist_tlist_contains(&node->n_tree->t_active_nds, node)); */
 		ndlist_tlist_del(node);
 		tree_unlock(node->n_tree, lock_acquired);
 		node->n_seq = 0;
@@ -3997,7 +4013,7 @@ int64_t btree_close_tree_tick(struct m0_sm_op *smop)
 		 * to be handled in a better way.
 		 */
 		if (ndlist_tlist_length(&td_curr->t_active_nds) > 1) {
-			if (m0_time_seconds(m0_time_now() - 
+			if (m0_time_seconds(m0_time_now() -
 					    td_curr->t_starttime) > 5) {
 				td_curr->t_starttime = 0;
 				return M0_ERR(-ETIMEDOUT);
@@ -5521,7 +5537,7 @@ static void ut_basic_tree_oper(void)
 							     &btree_type, nt,
 							     &b_op, tx));
 	M0_ASSERT(rc == 0);
-	
+
 	rc = M0_BTREE_OP_SYNC_WITH_RC(&b_op, m0_btree_close(b_op.bo_arbor,
 							    &b_op));
 	M0_ASSERT(rc == 0);
@@ -6733,7 +6749,7 @@ static void btree_ut_tree_oper_thread_handler(struct btree_ut_thread_info *ti)
 		rc = M0_BTREE_OP_SYNC_WITH_RC(&b_op,
 					      m0_btree_close(tree, &b_op));
 		M0_ASSERT(rc == 0);
-		
+
 		if (b_op.bo_arbor->t_desc->t_ref > 0) {
 			rc = M0_BTREE_OP_SYNC_WITH_RC(&b_op,
 						      m0_btree_destroy(tree,
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -1802,6 +1802,12 @@ static int64_t node_get(struct node_op *op, struct td *tree,
 		 * If node descriptor is not present allocate a new one
 		 * and assign to node.
 		 */
+		/**
+		 * TODO: Adding lru_lock to protect from two threads allocating
+		 * two node descriptors for the same node. Replace it with a
+		 * different global lock once hash functionality is implemented.
+		 */
+		m0_rwlock_write_lock(&lru_lock);
 		node = m0_alloc(sizeof *node);
 		/**
 		 * TODO: If Node-alloc fails, free up any node descriptor from
@@ -1809,15 +1815,14 @@ static int64_t node_get(struct node_op *op, struct td *tree,
 		 * segment. Take up with BE segment task.
 		 */
 		M0_ASSERT(node != NULL);
-		m0_rwlock_init(&node->n_lock);
-		m0_rwlock_write_lock(&node->n_lock);
 		node->n_addr = *addr;
 		node->n_tree = tree;
 		node->n_type = nt;
 		node->n_seq  = m0_time_now();
 		node->n_ref  = 1;
+		m0_rwlock_init(&node->n_lock);
 		op->no_node = node;
-		m0_rwlock_write_unlock(&node->n_lock);
+		m0_rwlock_write_unlock(&lru_lock);
 
 		nt->nt_opaque_set(addr, node);
 
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -711,7 +711,7 @@ static int get_tick(struct m0_btree_op *bop)
 
 		for (i = 0; i < oi->i_used; ++i) {
 			if (oi->i_level[i].l_node != NULL) {
-				node_put(oi->i_level[i].l_node);
+				node_put(oi->i_level[i].l_node, tree);
 				oi->i_level[i].l_node = NULL;
 			}
 		}
@@ -1074,7 +1074,7 @@ static void    tree_put   (struct td *tree);
 static int64_t    node_get  (struct node_op *op, struct td *tree,
 			     struct segaddr *addr, int nxt);
 #ifndef __KERNEL__
-static void       node_put  (struct nd *node);
+static void       node_put  (struct nd *node, struct td *tree);
 #endif
 
 
@@ -1085,9 +1085,9 @@ static struct nd *node_try  (struct td *tree, struct segaddr *addr);
 static int64_t    node_alloc(struct node_op *op, struct td *tree, int size,
 			     const struct node_type *nt, int ksize, int vsize,
 			     struct m0_be_tx *tx, int nxt);
-#ifndef __KERNEL__
 static int64_t    node_free(struct node_op *op, struct nd *node,
 			    struct m0_be_tx *tx, int nxt);
+#ifndef __KERNEL__
 static void node_op_fini(struct node_op *op);
 #endif
 #ifndef __KERNEL__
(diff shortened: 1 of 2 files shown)


btree/btree.c
```diff
@@ -1977,15 +2014,30 @@ static int64_t mem_node_get(struct node_op *op, struct td *tree,
...
-static void mem_node_put(struct nd *node)
+static void mem_node_put(struct nd *node, struct td *tree)
 {
-	/**
-	 * This implementation does not perform any action, but the final one
-	 * should decrement the reference count of the node and if this
-	 * reference count goes to '0' then detach this node descriptor from the
-	 * existing node and make this node descriptor available for 'node_get'.
-	 */
 	node->n_ref--;
```
RadhaGulhane13, line 2019: don't we want to have implementation of nd::n_lock in this PR, because you are dealing with adding and deleting nd from ndlist /tree list , there can be scenario where one thread see node->n_ref ==0 and meanwhile other thread increase n_ref, So, to avoid such scenarios we should  increment or decrement n_ref using n_lock and should perform addition /deletion of nd from ndlist/treelist

btree/btree.c
```diff
@@ -3645,6 +3684,12 @@ int64_t btree_destroy_tick(struct m0_sm_op *smop)
 		if (bop->bo_i == NULL)
 			return M0_ERR(-ENOMEM);
 
+		/**
+		 * TODO: Currently putting it here, call it inside
+		 * tree_*() function once desctroy tick is implemented
```
RadhaGulhane13, line 3689: desctroy->destroy

btree/btree.c
```diff
@@ -4055,7 +4102,10 @@ int64_t btree_iter_kv_tick(struct m0_sm_op *smop)
 			struct slot    s = {};
 			struct segaddr child;
 
+			lev = &oi->i_level[oi->i_pivot];
```
RadhaGulhane13, line 4105: you need not have to store sibling node at each level because we are only interested in leaf node sibling . If you are storing just to do node_put() as I see you are doing node_put() siblings present at each level in level_cleanup, you can do node_put at this state only, if it is internal node, you get the child index load child and then perform node_put for the parent .

btree/btree.c
```diff
@@ -3746,11 +3791,12 @@ int  btree_sibling_first_key_get(struct m0_btree_oimpl *oi, struct td *tree,
 				node_child(s, &child);
 				if (!address_in_segment(child))
 					return M0_ERR(-EFAULT);
+				i++;
 				node_get(&oi->i_nop, tree,
 					 &child, P_CLEANUP);
 				s->s_idx = 0;
 				s->s_node = oi->i_nop.no_node;
-				i++;
+				oi->i_level[i].l_sibling = oi->i_nop.no_node;
```
RadhaGulhane13, line 3799: Please refer the same comment below .

btree/btree.c
```diff
@@ -1708,13 +1711,40 @@ static int64_t node_get(struct node_op *op, struct td *tree,
...
+			m0_rwlock_write_lock(&lru_lock);
+			M0_ASSERT_EX(ndlist_tlist_contains(&btree_lru_nds,
+							   op->no_node));
+			ndlist_tlist_del(op->no_node);
+			m0_rwlock_write_unlock(&lru_lock);
+
+			m0_rwlock_write_lock(&tree->t_lock);
+			ndlist_tlist_add(&tree->t_active_nds, op->no_node);
+			m0_rwlock_write_unlock(&tree->t_lock);
+			/**
```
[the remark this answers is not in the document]
NikhilBirgade: Thanks, Yes this is valid scenario. Will make changes.

btree/btree.c
NikhilBirgade: I wanted to but since the n_lock will be used for other fields of nd also (n_seq) etc, I thought of taking it as a separate task. I will do one thing, will add n_ref in this PR.  Rest of the n_lock related changes can be taken up in separate task.

btree/btree.c
NikhilBirgade: Done.

btree/btree.c
NikhilBirgade: I am planning to modify the `get_tick` SLANT code also, to work in unlock mode. For that I will need sibling node. Just to have coherent code, I am storing l_sibling and clearing up it in level_cleanup(). We anyways had allocated memory for `l_sibling`, so assigning a node to this will not cause any performance degradation.

btree/btree.c
NikhilBirgade: I am planning to modify the get_tick SLANT code also, to work in unlock mode. For that I will need sibling node. Just to have coherent code, I am storing l_sibling and clearing up it in level_cleanup(). We anyways had allocated memory for l_sibling, so assigning a node to this will not cause any performance degradation.

btree/btree.c
```diff
@@ -1708,13 +1725,42 @@ static int64_t node_get(struct node_op *op, struct td *tree,
...
+		} else
+			node_refcnt_update(op->no_node, true);
+	} else {
 		/**
 		 * If node descriptor is not present allocate a new one
-		 * and assign to node. */
+		 * and assign to node.
+		 */
 		node = m0_alloc(sizeof *node);
+		/**
```
shashank-parulekar, line 1759: What if 2 threads reach this same line at the same time on different CPUs !!!
Currently we do not have any protection from allocating 2 node descriptors for the same node !!!

btree/btree.c
```diff
@@ -1786,16 +1833,24 @@ static int64_t node_alloc(struct node_op *op, struct td *tree, int size,
 	return nxt_state;
 }
 
-#ifndef __KERNEL__
 static int64_t node_free(struct node_op *op, struct nd *node,
 			 struct m0_be_tx *tx, int nxt)
 {
 	int shift = node->n_type->nt_shift(node);
```
shashank-parulekar, line 1839: Add an assert to make sure that all the records in this node have been removed. This is helpful during debugging.

btree/btree.c
```diff
@@ -1786,16 +1833,24 @@ static int64_t node_alloc(struct node_op *op, struct td *tree, int size,
...
+
+	node_refcnt_update(node, false);
```
shashank-parulekar, line 1841: I assume the code which will wait for all the references to have gone to 0 will be coming in a later PR.

btree/btree.c
```diff
@@ -2705,17 +2777,16 @@ static struct m0_btree_oimpl *level_alloc(int height)
 	return oi;
 }
 
-static void level_cleanup(struct m0_btree_oimpl *oi,
+static void level_cleanup(struct m0_btree_oimpl *oi, struct td *tree,
```
shashank-parulekar, line 2780: As I mentioned above the tree_put does not need a pointer to the tree descriptor so the tree pointer parameter to level_cleanup can also be removed.

shashank-parulekar (changes requested):

btree/btree.c
```diff
@@ -1434,20 +1434,23 @@ int m0_btree_mod_init(void)
 	trees_loaded = 0;
 	m0_rwlock_init(&trees_lock);
```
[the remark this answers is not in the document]
NikhilBirgade: It was going to be covered in a separate task along with tree code restructuring. Or is there a change in plan and it should bee included with these changes.

btree/btree.c
```diff
@@ -1074,7 +1074,7 @@ static void    tree_put   (struct td *tree);
 static int64_t    node_get  (struct node_op *op, struct td *tree,
 			     struct segaddr *addr, int nxt);
 #ifndef __KERNEL__
-static void       node_put  (struct nd *node);
+static void       node_put  (struct nd *node, struct td *tree);
```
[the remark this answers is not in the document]
NikhilBirgade: Ok.

btree/btree.c
```diff
@@ -1391,6 +1392,19 @@ static void node_del(const struct nd *node, int idx, struct m0_be_tx *tx)
...
+/**
+ * Updates the node reference count
+ *
+ * @param node The node descriptor whose ref count needs to be updated.
+ * @param increment If true increase ref count.
+ *		    If false decrease ref count.
+ */
+static void node_refcnt_update(struct nd *node, bool increment)
+{
+	m0_rwlock_write_lock(&node->n_lock);
```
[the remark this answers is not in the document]
NikhilBirgade: ok.

btree/btree.c
```diff
@@ -1434,20 +1448,23 @@ int m0_btree_mod_init(void)
 	trees_loaded = 0;
 	m0_rwlock_init(&trees_lock);
 
+	/** Initialtise lru list and lock. */
```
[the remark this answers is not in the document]
NikhilBirgade: Not here , it should be added in `m0_btree_mod_fini()` . I had skipped it as per our discussion and put a comment `/** Skipping lru list teardown. */`. I will add clearing lru list changes.

btree/btree.c
RadhaGulhane13: okay

btree/btree.c
RadhaGulhane13: okay

btree/btree.c
NikhilBirgade: Yes. Will  merge those changes once they are pushed.

btree/btree.c
NikhilBirgade: Thanks, removed tree pointer.

btree/btree.c
NikhilBirgade: If I add `node_count()` check here , it will call `node_invariant()`. Node invariant has check `node->n_skip_rec_count_check || ergo(h->ff_level > 0, h->ff_used > 0)`. This check will fail for `btree_del_kv_tick() --> case P_FREENODE -->  node_free()`, since it will delete internal nodes with 0 records. There is a task for enhancing invariant, is it okay to put this as part of the same task? This task will make sure to enhance invariant and also add invariants in functions

btree/btree.c
NikhilBirgade: Added nd::n_lock to protect node descriptor field update.

btree/btree.c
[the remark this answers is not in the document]
shashank-parulekar: correct in the fini() routine

btree/btree.c
shashank-parulekar: Even this does not help because both the threads would still allocate nodes because they have no way to know that other thread has already allocated. Need to think about this corner case. This is the time where Nikita mentioned about a global lock which can be used here. If you do not want to introduce a new lock you can maybe use the lru lock for now.

btree/btree.c
shashank-parulekar: Good thing you reminded me. We need to rethink about the invariant. Some way we need to make the invariant aware of these operations or else the invariant should not be too strict.

btree/btree.c
```diff
@@ -1828,15 +1892,17 @@ static int64_t node_alloc(struct node_op *op, struct td *tree, int size,
...
 	int shift = node->n_type->nt_shift(node);
-	node->n_ref--;
+
+	node_refcnt_update(node, false);
 	node->n_type->nt_fini(node);
 
 	if (node->n_ref == 0) {
+		ndlist_tlink_del_fini(node);
+		m0_rwlock_fini(&node->n_lock);
 		m0_free(node);
```
RadhaGulhane13, line 1911: before m0_free(node), we should copy segaddr   op->no_addr = node->n_addr

btree/btree.c
NikhilBirgade: Oh ok. Yes, will add it under lru lock for now and put a comment to have a separate global lock once the hash functionality is implemented.

btree/btree.c
```diff
@@ -1828,15 +1892,17 @@ static int64_t node_alloc(struct node_op *op, struct td *tree, int size,
...
 {
 	int shift = node->n_type->nt_shift(node);
-	node->n_ref--;
+
+	node_refcnt_update(node, false);
 	node->n_type->nt_fini(node);
 
 	if (node->n_ref == 0) {
+		ndlist_tlink_del_fini(node);
+		m0_rwlock_fini(&node->n_lock);
```
RadhaGulhane13, line 1910: after reaching to this point, some other thread may request same  node, and increase n_ref . Such scenarios needs to be think. node_get and node_free should be in sync

btree/btree.c
NikhilBirgade: Yes, we need to think about it because current invariants might not catching bugs. In my opinion we should have strict invariants for development phase. May be for development/verification phase we can have some invariants (like function specific) under `M0_PRE_EX`, which we can remove later and some invariants under `M0_PRE` which we can keep.

btree/btree.c
```diff
@@ -2028,13 +2097,29 @@ static int64_t mem_node_get(struct node_op *op, struct td *tree,
...
 {
-	/**
-	 * This implementation does not perform any action, but the final one
-	 * should decrement the reference count of the node and if this
-	 * reference count goes to '0' then detach this node descriptor from the
-	 * existing node and make this node descriptor available for 'node_get'.
-	 */
-	node->n_ref--;
+	node_refcnt_update(node, false);
+	if (node->n_ref == 0) {
```
RadhaGulhane13, line 2106: Let's consider scenario: there are 2 threads refering to same node, therefore n_ref ==2, now both executed line 2100 one after other and now both will reach node->n_ref = 0, and execute if part which should not happen

btree/btree.c
```diff
@@ -3472,7 +3551,7 @@ static int64_t btree_put_kv_tick(struct m0_sm_op *smop)
 		return P_CLEANUP;
 	}
 	case P_CLEANUP: {
-		level_cleanup(oi, bop->bo_tx);
+		level_cleanup(oi,  bop->bo_tx);
```
RadhaGulhane13, line 3559: extra space before bop->bo_tx

btree/btree.c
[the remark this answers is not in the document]
NikhilBirgade: Modified code to handle race condition.

btree/btree.c
NikhilBirgade: Moved `n_delayed_free = true` to top of the function, so that race condition is handled.

btree/btree.c
NikhilBirgade: Modified code to handle race condition.

btree/btree.c
```diff
@@ -1743,16 +1796,56 @@ static int64_t node_get(struct node_op *op, struct td *tree,
...
+		m0_rwlock_write_unlock(&op->no_node->n_lock);
+	} else {
 		/**
 		 * If node descriptor is not present allocate a new one
-		 * and assign to node. */
+		 * and assign to node.
+		 */
+		/**
+		 * TODO: Adding lru_lock to protect from two threads allocating
+		 * two node descriptors for the same node. Replace it with a
```
RadhaGulhane13, line 1839: comment should be modified; it is not necessary that there will be two threads only, you might want o  use more than one thread , more than one node descriptor than two thread and two node descriptor , respectively.

RadhaGulhane13 (approved):

btree/btree.c
NikhilBirgade: Sure, will modify comment.

btree/btree.c
```diff
@@ -1778,24 +1872,25 @@ static int64_t node_get(struct node_op *op, struct td *tree,
...
- *             not been loaded.
- * @param addr node address in the segment.
- * @param nxt state to return on successful completion
+ * @param node node descriptor.
+ * @param tree pointer to tree.
  *
- * @return next state
  */
 static void node_put(struct node_op *op, struct nd *node, bool lock_acquired,
 		     struct m0_be_tx *tx)
```
ydb242, line 1896: Comment given for node_put parameters and the actual parameters dont match.
Is this expected?

btree/btree.c
```diff
@@ -1759,16 +1812,57 @@ static int64_t node_get(struct node_op *op, struct td *tree,
...
-		 * and assign to node. */
+		 * and assign to node.
+		 */
+		/**
+		 * TODO: Adding lru_lock to protect from multiple threads 
+		 * allocating more than one node descriptors for the same node. 
+		 * Replace it with a different global lock once hash 
+		 * functionality is implemented.
+		 */
+		m0_rwlock_write_lock(&lru_lock);
```
shashank-parulekar, line 1859: This will still not help because after the lock a check is needed to confirm if the node descriptor is present for the node in BE segment.
In the current case 2 threads will again allocate node descriptor(s) for the same node in different threads.

btree/btree.c
```diff
@@ -2050,13 +2157,31 @@ static int64_t mem_node_get(struct node_op *op, struct td *tree,
...
-	node->n_ref--;
+	m0_rwlock_write_lock(&node->n_lock);
+	node_refcnt_update(node, false);
+	if (node->n_ref == 0) {
+		/**
+		 * The node descriptor is in tree's active list. Remove from
+		 * active list and add to lru list
+		 */
+		tree_lock(node->n_tree, lock_acquired);
+		/* M0_ASSERT_EX(ndlist_tlist_contains(&node->n_tree->t_active_nds, node)); */
```
ydb242, line 2168: Comment exceeding 80characters.

btree/btree.c
shashank-parulekar: yes we can have M0_PRE_EX but we still need to find how can the invariant inspect the node condition and apply appropriate checks instead of taking hints given  by the caller.

shashank-parulekar (changes requested):

btree/btree.c
NikhilBirgade: Done.

btree/btree.c
NikhilBirgade: Oh yes. Added code to check if node descriptor is allocated. If yes increase the n_ref count and return.

shashank-parulekar (approved):

ydb242 (approved):

PR #803, opened 2021-06-28
Update Quick-Start-Guide.rst

mengwanguc:
Add `-y` when installing `epel-release` and `ansible` using yum.

These two `yum install` commands are put together in the same code block. If users copy the two lines together and paste them in the shell, shell will ask this when installing `epel-release`:
```
Installing:
 epel-release                         noarch                         7-11                          extras                          15 k
Total download size: 15 k
Installed size: 24 k
Is this ok [y/d/N]:
```
And shell will automatically treat `sudo yum install ansible` as the answer to `Is this ok [y/d/N]`.

In this case, shell won't run `sudo yum install ansible` as a command, as it's treated as an answer to `Is this ok [y/d/N]`.

To avoid this problem, it's better if we add option `-y` when running `yum install`.

### DIFF
--- a/doc/Quick-Start-Guide.rst
+++ b/doc/Quick-Start-Guide.rst
@@ -12,8 +12,8 @@ The prerequisite that is necessary to install the Motr component is mentioned be
 
 - **Ansible** is needed::
 
-    sudo yum install epel-release # Install EPEL yum repo
-    sudo yum install ansible
+    sudo yum install -y epel-release # Install EPEL yum repo
+    sudo yum install -y ansible
 
 Get the Sources
 ===============


cortx-admin:

huanghua78 (commented):
This has been discussed before.
Please note, this "Quick-Start-Guide" is not a shell script.
It shows what to do.
Developers should read and understand the steps before copying and pasting these commands into a terminal.

These two two separate commands, and they should not be mixed.
If you read on, and you will see 
```
    sudo modprobe lnet
    sudo lctl list_nids
```
You will also treat these as two commands, and will not mix them in a single line.

mengwanguc:
Closing this PR as this issue has been discussed before and a decision had been made.

PR #804, opened 2021-06-28
Update Quick-Start-Guide.rst

mengwanguc:
need to be root user in order to run `python -m pip uninstall pip setuptools`. Otherwise will get permission errors.


mengwanguc:
Closing this because found another command that needs root user permission.

Put them together in #805

PR #806, opened 2021-06-29
Adding new Motr gatekeeper.

huanghua78:


truptiatseagate (approved):
Increasing the gatekeepers count temporarily to 5 until end of July. We will get it back to 4 at the end of July. (Huanghua has now transitioned to ADG)

PR #807, opened 2021-06-29
Update Quick-Start-Guide.rst

mengwanguc:
I met error `insmod: ERROR: could not insert module /home/cc/cortx-motr/extra-libs/galois/src/linux_kernel/galois.ko: File exists
Error unloading /home/cc/cortx-motr/m0tr.ko.`.

I was running system tests. Before the tests finish, I forcefully stopped the tests. Then when I run `sudo scripts/m0 run-st` again, I met the error above.

The solution comes from https://github.com/Seagate/cortx/blob/main/doc/Build-Installation-FAQ.md

### DIFF
--- a/doc/Quick-Start-Guide.rst
+++ b/doc/Quick-Start-Guide.rst
@@ -146,6 +146,36 @@ Troubleshooting
        sudo systemctl restart lnet
        sudo lctl list_nids
 
+- If you see error::
+
+    rmmod: ERROR: Module m0tr is in use or rmmod: ERROR: Module galois is in use by: m0tr
+    
+  or::
+  
+    insmod: ERROR: could not insert module /root/cortx-motr/extra-libs/galois/src/linux_kernel/galois.ko: File exists Error unloading /root/cortx-motr/m0tr.ko.
+    
+  This happens because previous runs may have exited abruptly or because the runs was forcefully stopped using ctrl+c. To resolve it, do the following:
+  
+  1. List the modules::
+  
+       $ [root@localhost ~]# lsmod
+         Module                  Size  Used by
+         m0ctl                  48916  0
+         m0tr                13901353  1 m0ctl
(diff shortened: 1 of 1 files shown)


cortx-admin:

doc/Quick-Start-Guide.rst
```diff
@@ -146,6 +146,36 @@ Troubleshooting
        sudo systemctl restart lnet
        sudo lctl list_nids
 
+- If you see error::
```
huanghua78, line 149: >The solution comes from https://github.com/Seagate/cortx/blob/main/doc/Build-Installation-FAQ.md

We don't duplicate the information everywhere.
Adding a reference to the doc is a better idea.

This is a quick start guide, and let's keep it as simple as possible.
We may have another document to list FAQ, troubleshooting, or other information.

mengwanguc: Thanks, that makes sense. 
I duplicated the solution because I saw Qs2 from that doc is already duplicated in the Quick Start Guide, so I wanted to follow the same style.

I agree that adding a reference to the doc will be a better idea. I will close this one and create another PR.
In that case, do you think we should remove the solution to "lctl list_nids command does not show any output" problem? It's also duplicated as Qs2 from the doc.

huanghua78: I would prefer them to be kept untouched.
Because, developers may see this network issue before doing any testing. So they may be blocked by that.

But the problem you mentioned, only happens after a non-clean stop of tests. (if some tests fail and the kernel modules are not removed by the testing script)

mengwanguc: Got it thanks

mengwanguc:
Closing this issue as discussed above. Opened another issue #811 to add a reference link to the FAQs page.

PR #809, opened 2021-06-30
bug fixes :  Adding i_height to determine height change

RadhaGulhane13:


btree/btree.c
```diff
@@ -3468,14 +3471,21 @@ static int64_t btree_put_kv_tick(struct m0_sm_op *smop)
...
 
 			oi->i_key_found = node_find(&node_slot,
 						    &bop->bo_rec.r_key);
 			lev->l_idx = node_slot.s_idx;
 			if (node_level(node_slot.s_node) > 0) {
+				if (oi->i_used == oi->i_height - 1) {
+					/* If height of tree increased. */
+					return m0_sm_op_sub(&bop->bo_op,
+							    P_CLEANUP, P_SETUP);
+				}
```
shashank-parulekar, line 3487: Better to relocate this check below right after the oi->i_used++ on line 3498. With this the code become easier to understand and guess. Also after the increment of i_used the check will be if (oi->i_used >= ot->i_height)

ydb242 (approved):

btree/btree.c
```diff
@@ -4150,14 +4161,21 @@ static int64_t btree_get_kv_tick(struct m0_sm_op *smop)
...
 			oi->i_key_found = node_find(&node_slot,
 						    &bop->bo_rec.r_key);
 			lev->l_idx = node_slot.s_idx;
 
 			if (node_level(node_slot.s_node) > 0) {
+				if (oi->i_used == oi->i_height - 1) {
+					/* If height of tree increased. */
+					return m0_sm_op_sub(&bop->bo_op,
+							    P_CLEANUP, P_SETUP);
+				}
```
shashank-parulekar, line 4177: same comment as above.

shashank-parulekar (changes requested):

lib/list.c
```diff
@@ -47,6 +47,7 @@ M0_EXPORTED(m0_list_is_empty);
 
 M0_INTERNAL bool m0_list_link_invariant(const struct m0_list_link *link)
 {
+#if 0
```
NikhilBirgade, line 50: Please revert changes from list.c and tlist.c. We can disable expensive invariants at compile time.

btree/btree.c
```diff
@@ -4327,13 +4345,20 @@ int64_t btree_iter_kv_tick(struct m0_sm_op *smop)
...
-				m0_sm_op_sub(&bop->bo_op, P_CLEANUP, P_SETUP);
+				return m0_sm_op_sub(&bop->bo_op, P_CLEANUP,
+						    P_SETUP);
 			}
 
 			oi->i_key_found = node_find(&s, &bop->bo_rec.r_key);
 			lev->l_idx = s.s_idx;
 
 			if (node_level(s.s_node) > 0) {
+				if (oi->i_used == oi->i_height - 1) {
```
NikhilBirgade, line 4356: similar check is needed at P_SIBLING for `i_pivot`.

btree/btree.c
RadhaGulhane13: Then we unnecessarily have to perform node_child() and address_in_Segment() check . To avoid this if height increased, I am doing check at line 3483

btree/btree.c
RadhaGulhane13: Please refer above comment

lib/list.c
RadhaGulhane13: ok.

btree/btree.c
RadhaGulhane13: done.

NikhilBirgade (approved):

btree/btree.c
shashank-parulekar: That is fine but we need to make sure the code looks nice and clean and compartmentalized.
Also again this is not a very frequent occurrence where the height would have changed so please make the change as I suggested as long as this move is not going to cause any crashes.

shashank-parulekar (changes requested):

shashank-parulekar (approved):

PR #813, opened 2021-07-01
EOS-21221: Move active nd list from tree descriptor to global scope.

NikhilBirgade:
In multithread UT, there was deadlock between node lock and tree lock.
Moved active node descriptor list to global scope to avoid deadlock.

### DIFF
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -1254,7 +1254,7 @@ static struct m0_rwlock trees_lock;
 /**
  * Node descriptor LRU list.
  * Following actions will be performed on node descriptors:
- * 1. If nds are not active, they will be  moved from btree_active_nds to
+ * 1. If nds are not active, they will be moved from btree_active_nds to
  * btree_lru_nds list head.
  * 2. If the nds in btree_lru_nds become active, they will be moved to
  * btree_active_nds list head.
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -814,13 +814,6 @@ struct td {
 	 * expected.
 	 */
 	m0_time_t               t_starttime;
-
-	/**
-	 * Active node descriptor list contains the node descriptors that are
-	 * currently in use by the tree.
-	 * Node descriptors are linked through nd:n_linkage to this list.
-	 */
-	struct m0_tl                t_active_nds;
 };
 
 /** Special values that can be passed to node_move() as 'nr' parameter. */
@@ -1000,7 +993,10 @@ struct nd {
 	 */
 	bool                    n_skip_rec_count_check;
 
-	/** Linkage into node descriptor list. ndlist_tl, td::t_active_nds. */
+	/**
+	 * Linkage into node descriptor list.
+	 * ndlist_tl, btree_active_nds, btree_lru_nds.
+	 */
 	struct m0_tlink	        n_linkage;
 	uint64_t                n_magic;
 
@@ -1098,14 +1094,12 @@ static int64_t tree_delete(struct node_op *op, struct td *tree,
 			   struct m0_be_tx *tx, int nxt);
 #endif
 static void    tree_put   (struct td *tree);
-static void    tree_lock(struct td *tree, bool lock_acquired);
-static void    tree_unlock(struct td *tree, bool lock_acquired);
 
 static int64_t    node_get  (struct node_op *op, struct td *tree,
-			     struct segaddr *addr, bool lock_acquired, int nxt);
+			     struct segaddr *addr, int nxt);
 #ifndef __KERNEL__
 static void       node_put  (struct node_op *op, struct nd *node,
-			     bool lock_acquired, struct m0_be_tx *tx);
+			     struct m0_be_tx *tx);
 #endif
 
 
@@ -1115,7 +1109,7 @@ static struct nd *node_try  (struct td *tree, struct segaddr *addr);
 
 static int64_t    node_alloc(struct node_op *op, struct td *tree, int size,
 			     const struct node_type *nt, int ksize, int vsize,
-			     bool lock_aquired, struct m0_be_tx *tx, int nxt);
+			     struct m0_be_tx *tx, int nxt);
 static int64_t    node_free(struct node_op *op, struct nd *node,
 			    struct m0_be_tx *tx, int nxt);
 #ifndef __KERNEL__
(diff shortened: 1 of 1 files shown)


btree/btree.c
```diff
@@ -3992,42 +3968,56 @@ int64_t btree_open_tree_tick(struct m0_sm_op *smop)
...
 		/**
 		 * This code is meant for debugging. In future, this case needs
 		 * to be handled in a better way.
 		 */
-		if (ndlist_tlist_length(&td_curr->t_active_nds) > 1) {
-			if (m0_time_seconds(m0_time_now() -
-					    td_curr->t_starttime) > 5) {
-				td_curr->t_starttime = 0;
-				return M0_ERR(-ETIMEDOUT);
+		m0_rwlock_read_lock(&lru_lock);
```
shashank-parulekar, line 3993: Why are you taking a read lock instead of write lock ?

btree/btree.c
```diff
@@ -3992,42 +3968,56 @@ int64_t btree_open_tree_tick(struct m0_sm_op *smop)
...
+						return M0_ERR(-ETIMEDOUT);
+					}
+					m0_rwlock_read_unlock(&lru_lock);
+					return P_TIMECHECK;
+				}
 			}
-			return P_TIMECHECK;
-		}
+		} m0_tl_endfor;
+		m0_rwlock_write_unlock(&lru_lock);
```
shashank-parulekar, line 4011: Above you took read lock and here you are releasing write lock !!!

btree/btree.c
```diff
@@ -3992,42 +3968,56 @@ int64_t btree_open_tree_tick(struct m0_sm_op *smop)
...
 	case P_INIT:
-		M0_ASSERT(td_curr->t_ref != 0);
-		if (td_curr->t_ref > 1) {
-			tree_put(td_curr);
+		if (tree->t_ref > 1) {
+			tree_put(tree);
 			return P_DONE;
 		}
-		td_curr->t_starttime = m0_time_now();
+		tree->t_starttime = m0_time_now();
```
shashank-parulekar, line 3985: We should 'put' the root node here.

btree/btree.c
```diff
@@ -3992,42 +3968,56 @@ int64_t btree_open_tree_tick(struct m0_sm_op *smop)
...
 		 */
-		if (ndlist_tlist_length(&td_curr->t_active_nds) > 1) {
-			if (m0_time_seconds(m0_time_now() -
-					    td_curr->t_starttime) > 5) {
-				td_curr->t_starttime = 0;
-				return M0_ERR(-ETIMEDOUT);
+		m0_rwlock_read_lock(&lru_lock);
+		m0_tl_for(ndlist, &btree_active_nds, node) {
+			if (node->n_tree == tree && node->n_ref > 0) {
+				root = node;
```
shashank-parulekar, line 3996: The first node descriptor in the active list matches our tree we can check how much time has elapsed. There is no need to continue with the rest of the list elements.

btree/btree.c
```diff
@@ -3992,42 +3968,56 @@ int64_t btree_open_tree_tick(struct m0_sm_op *smop)
...
+		} m0_tl_endfor;
+		m0_rwlock_write_unlock(&lru_lock);
 		/** Fallthrough to P_ACT */
 
 	case P_ACT:
-		if (nd_head == td_curr->t_root)
-			node_put(nd_head->n_op, nd_head, false, bop->bo_tx);
+		M0_ASSERT(active_node_count == 1 && root == tree->t_root);
+		if (root == tree->t_root)
+			node_put(root->n_op, root, bop->bo_tx);
```
shashank-parulekar, line 4017: Check my comment above regarding the release of the root node.

shashank-parulekar (changes requested):

btree/btree.c
```diff
@@ -1905,24 +1882,28 @@ static int64_t node_get(struct node_op *op, struct td *tree,
...
-static void node_put(struct node_op *op, struct nd *node, bool lock_acquired,
-		     struct m0_be_tx *tx)
+static void node_put(struct node_op *op, struct nd *node, struct m0_be_tx *tx)
 {
 	M0_PRE(node != NULL);
 	int        shift = node->n_type->nt_shift(node);
-	segops->so_node_put(node, lock_acquired);
+	segops->so_node_put(node);
 
+	m0_rwlock_read_lock(&node->n_lock);
```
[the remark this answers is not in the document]
shashank-parulekar: Maybe that is fine in this implementation, but in the new implementation by Yash the reference counter will be decremented in the node_put and hence write lock might be needed.

btree/btree.c
```diff
@@ -3992,42 +3968,56 @@ int64_t btree_open_tree_tick(struct m0_sm_op *smop)
...
+			if (node->n_tree == tree && node->n_ref > 0) {
+				root = node;
+				active_node_count++;
+				if (active_node_count > 1) {
+					if (m0_time_seconds(m0_time_now() -
+							    tree->t_starttime)
+					    > 5) {
+						m0_rwlock_read_unlock(&lru_lock);
+						tree->t_starttime = 0;
+						return M0_ERR(-ETIMEDOUT);
```
shashank-parulekar: Do not return error here. Instead crash by calling ASSERT so that the crash dump can be used for debugging if any unexpected logic is present.

shashank-parulekar (changes requested):

btree/btree.c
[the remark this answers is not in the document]
NikhilBirgade: Okay, I will change read locks to write locks so that we do not miss later.

btree/btree.c
NikhilBirgade: okay, I will add assert.

shashank-parulekar (approved):

btree/btree.c
```diff
@@ -1923,24 +1897,28 @@ static int64_t node_get(struct node_op *op, struct td *tree,
...
-static void node_put(struct node_op *op, struct nd *node, bool lock_acquired,
-		     struct m0_be_tx *tx)
+static void node_put(struct node_op *op, struct nd *node, struct m0_be_tx *tx)
 {
 	M0_PRE(node != NULL);
 	int        shift = node->n_type->nt_shift(node);
-	segops->so_node_put(node, lock_acquired);
+	segops->so_node_put(node);
 
+	node_lock(node);
```
RadhaGulhane13, line 1909: With current implementation, if node_free and node_put get called at the same time with n_ref =2,now one thread mem_node_put will decrease ref counter and then other thread can take lock and decrease the n_ref and it will see n_ref as 0 and will free the node, the first thread is pointing out to the node which is not valid, which contains garbage values. 
move node_lock(node) before segops->so_node_put(node); and remove the lock taken in mem_node_put. as lock will be taken in node_put() function .

RadhaGulhane13: The same thing is possible when we call node_get and node_free/ node_put for same node with ref count as 1

btree/btree.c
```diff
@@ -1255,21 +1249,28 @@ static struct m0_rwlock trees_lock;
 /**
  * Node descriptor LRU list.
  * Following actions will be performed on node descriptors:
- * 1. If nds are not active, they will be  moved from td::t_active_nds to
+ * 1. If nds are not active, they will be  moved from btree_active_nds to
```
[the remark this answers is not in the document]
NikhilBirgade: Done.

btree/btree.c
NikhilBirgade: Added global lock for node_* operations.

RadhaGulhane13 (approved):

PR #815, opened 2021-07-01
fix example1.c build instruction

mengwanguc:
When building example1.c, should use `export LD_LIBRARY_PATH=/work/cortx-motr/motr/.libs/`.
Added this `export`.
Without this `export`, will get error `./example1: error while loading shared libraries: libmotr.so.2: cannot open shared object file: No such file or directory`.

### DIFF
--- a/motr/examples/example1.c
+++ b/motr/examples/example1.c
@@ -36,7 +36,7 @@
  * Please change the configuration according to you development environment.
  *
  * How to run:
- * LD_LIBRARY_PATH=/work/cortx-motr/motr/.libs/                              \
+ * export LD_LIBRARY_PATH=/work/cortx-motr/motr/.libs/                              \
  * ./example1 172.16.154.179@tcp:12345:34:1 172.16.154.179@tcp:12345:33:1000 \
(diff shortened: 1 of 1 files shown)


cortx-admin:

motr/examples/example1.c
```diff
@@ -36,7 +36,7 @@
  * Please change the configuration according to you development environment.
  *
  * How to run:
- * LD_LIBRARY_PATH=/work/cortx-motr/motr/.libs/                              \
+ * export LD_LIBRARY_PATH=/work/cortx-motr/motr/.libs/                              \
```
huanghua78, line 39: Please note the last char in this line, the back slash.
That means, this line, the next, and the next next line, should be put in a single line in shell.

So, you don't need to "export LD_LIBRARY_PATH", because it is done for this run.

huanghua78 (changes requested):
This change is not needed.

motr/examples/example1.c
mengwanguc: Got it. Sorry about that

PR #819, opened 2021-07-02
EOS-21926 - Initial commit for remove node descriptor from ff_*()

ydb242:
Signed-off-by: Yash Bhamare <yash.bhamare@seagate.com>

### DIFF
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -865,47 +865,47 @@ struct node_type {
 			int vsize, uint32_t ntype, struct m0_be_tx *tx);
 
 	/** Cleanup of the node if any before deallocation */
-	void (*nt_fini)(const struct nd *node);
+	void (*nt_fini)(const struct segaddr *addr);
 
 	/** Returns count of keys in the node */
-	int  (*nt_count)(const struct nd *node);
+	int  (*nt_count)(const struct segaddr *addr);
 
 	/** Returns count of records/values in the node*/
-	int  (*nt_count_rec)(const struct nd *node);
+	int  (*nt_count_rec)(const struct segaddr *addr);
 
 	/** Returns the space (in bytes) available in the node */
-	int  (*nt_space)(const struct nd *node);
+	int  (*nt_space)(const struct segaddr *addr);
 
 	/** Returns level of this node in the btree */
-	int  (*nt_level)(const struct nd *node);
+	int  (*nt_level)(const struct segaddr *addr);
 
 	/** Returns size of the node (as a shift value) */
-	int  (*nt_shift)(const struct nd *node);
+	int  (*nt_shift)(const struct segaddr *addr);
 
 	/**
 	 * Returns size of the key of node. In case of variable key size return
 	 * -1.
 	 */
-	int  (*nt_keysize)(const struct nd *node);
+	int  (*nt_keysize)(const struct segaddr *addr);
 
 	/**
 	 * Returns size of the value of node. In case variable value size
 	 * return -1.
 	 */
-	int  (*nt_valsize)(const struct nd *node);
+	int  (*nt_valsize)(const struct segaddr *addr);
 
 	/**
 	 * If predict is set as true, function determines if there is
 	 * possibility of underflow else it determines if there is an underflow
 	 * at node.
 	 */
-	bool  (*nt_isunderflow)(const struct nd *node, bool predict);
+	bool  (*nt_isunderflow)(const struct segaddr *addr, bool predict);
 
 	/** Returns true if there is possibility of overflow. */
-	bool  (*nt_isoverflow)(const struct nd *node);
+	bool  (*nt_isoverflow)(const struct segaddr *addr);
 
 	/** Returns unique FID for this node */
-	void (*nt_fid)  (const struct nd *node, struct m0_fid *fid);
+	void (*nt_fid)  (const struct segaddr *addr, struct m0_fid *fid);
 
 	/** Returns record (KV pair) for specific index. */
 	void (*nt_rec)  (struct slot *slot);
@@ -938,20 +938,21 @@ struct node_type {
 	 *  All the changes to the node have completed. Any post processing can
 	 *  be done here.
 	 */
-	void (*nt_fix)  (const struct nd *node, struct m0_be_tx *tx);
+	void (*nt_fix)  (const struct segaddr *addr, struct m0_be_tx *tx);
 
 	/**
 	 *  Changes the size of the value (increase or decrease) for the
 	 *  specified key
 	 */
-	void (*nt_cut)  (const struct nd *node, int idx, int size,
+	void (*nt_cut)  (const struct segaddr *addr, int idx, int size,
 			 struct m0_be_tx *tx);
 
 	/** Deletes the record from the node at specific index */
-	void (*nt_del)  (const struct nd *node, int idx, struct m0_be_tx *tx);
+	void (*nt_del)  (const struct segaddr *addr, int idx, 
+			 struct m0_be_tx *tx);
 
 	/** Updates the level of node */
-	void (*nt_set_level)  (const struct nd *node, uint8_t new_level,
+	void (*nt_set_level)  (const struct segaddr *addr, uint8_t new_level,
 			       struct m0_be_tx *tx);
 
 	/** Moves record(s) between nodes */
@@ -959,13 +960,13 @@ struct node_type {
 			 enum dir dir, int nr, struct m0_be_tx *tx);
 
 	/** Validates node composition */
-	bool (*nt_invariant)(const struct nd *node);
+	bool (*nt_invariant)(const struct segaddr *addr, bool check);
 
 	/** 'Does a thorough validation */
-	bool (*nt_verify)(const struct nd *node);
+	bool (*nt_verify)(const struct segaddr *addr);
 
 	/** Does minimal (or basic) validation */
-	bool (*nt_isvalid)(const struct nd *node);
+	bool (*nt_isvalid)(const struct segaddr *addr);
 	/** Saves opaque data. */
 	void (*nt_opaque_set)(const struct segaddr *addr, void *opaque);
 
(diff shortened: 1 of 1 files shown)


btree/btree.c
```diff
@@ -1913,7 +1955,7 @@ static void node_put(struct node_op *op, struct nd *node, bool lock_acquired,
 		     struct m0_be_tx *tx)
 {
 	M0_PRE(node != NULL);
-	int        shift = node->n_type->nt_shift(node);
+	int        shift = node->n_type->nt_shift(&node->n_addr);
 	segops->so_node_put(node, lock_acquired);
```
shashank-parulekar, line 1959: These have to be changed so that there are no more mem_node_* functions.

shashank-parulekar (changes requested):

btree/btree.c
```diff
@@ -1927,7 +1919,8 @@ static void node_put(struct node_op *op, struct nd *node, struct m0_be_tx *tx)
 			shift = node->n_type->nt_shift(node);
 			m0_free(node);
 			m0_rwlock_write_unlock(&list_lock);
-			segops->so_node_free(op, shift, tx, 0);
+			m0_free_aligned(segaddr_addr(&op->no_addr),
```
NikhilBirgade, line 1922: We need to capture this operation in transaction. Also `m0_free_aligned` is used at two places.  Put `m0_free_aligned` in another function and call this function from the two places. for eg. 
```
seg_node_free(op, shift, tx)
{
m0_free_aligned(segaddr_addr(&op->no_addr),1ULL << shift, shift);
/** Capture in transaction */
}

File: btree/btree.c [hunk 3]
```diff
@@ -941,7 +941,8 @@ struct node_type {
 			 struct m0_be_tx *tx);
 
 	/** Deletes the record from the node at specific index */
-	void (*nt_del)  (const struct nd *node, int idx, struct m0_be_tx *tx);
+	void (*nt_del)  (const struct nd *node, int idx,
+			 struct m0_be_tx *tx);
```
RadhaGulhane13, line 945: struct m0_be_tx *tx can be added to previous line

btree/btree.c
```diff
@@ -1962,9 +1955,16 @@ static int64_t node_alloc(struct node_op *op, struct td *tree, int size,
...
+	void      *area;
+	int        actual_size = 1ULL << size;
 
-	nxt_state = segops->so_node_alloc(op, tree, size, nt, tx, nxt);
+	M0_PRE(op->no_opc == NOP_ALLOC);
+	M0_PRE(node_shift_is_valid(size));
+	area = m0_alloc_aligned(actual_size, size);
+	M0_ASSERT(area != NULL);
+	op->no_addr = segaddr_build(area, size);
+	op->no_tree = tree;
```
NikhilBirgade, line 1967: put it in a function say 'seg_node_alloc`

btree/btree.c
```diff
@@ -2422,21 +2360,18 @@ static bool ff_invariant(const struct nd *node)
 {
 	const struct ff_head *h = ff_data(node);
 
-	return  _0C(h->ff_shift == segaddr_shift(&node->n_addr)) &&
-		_0C(node->n_skip_rec_count_check ||
-		    ergo(h->ff_level > 0, h->ff_used > 0));
+	return  _0C(h->ff_shift == segaddr_shift(&node->n_addr));
```
NikhilBirgade, line 2363: Why the check is removed ?

btree/btree.c
```diff
@@ -2489,15 +2424,17 @@ static void ff_fini(const struct nd *node)
 
 static int ff_count(const struct nd *node)
 {
-	int used = ff_data(node)->ff_used;
-	if (ff_data(node)->ff_level > 0)
+	struct ff_head *h = ff_data(node);
+	int used = h->ff_used;
```
NikhilBirgade, line 2428: I do not see any issue with previous code, why it is changed?

btree/btree.c
```diff
@@ -2489,15 +2424,17 @@ static void ff_fini(const struct nd *node)
...
+	if (h->ff_level > 0)
 		used --;
 	return used;
 }
 
 static int ff_count_rec(const struct nd *node)
 {
-	return ff_data(node)->ff_used;
+	struct ff_head *h = ff_data(node);
+	return h->ff_used;
```
NikhilBirgade, line 2437: Earlier code was correct, why it is changed here and at other places ?

btree/btree.c
```diff
@@ -1797,7 +1788,10 @@ static int64_t node_get(struct node_op *op, struct td *tree,
 	struct nd              *node;
 	bool                    in_lrulist;
 
-	nxt_state =  segops->so_node_get(op, tree, addr, nxt);
+	nxt_state = nxt;
+	if (tree == NULL) {
+		nxt_state = mem_tree_get(op, addr, nxt);
+	}
```
RadhaGulhane13, line 1794: No need for nxt_state varaible ,  mem_tree_get is returning the same state which u are passing as an argument.

btree/btree.c
ydb242: Sure. I will update it.

btree/btree.c
ydb242: I bought these changes in node_alloc as we wanted to remove the dependency of a mem_node_*() function.
I will make a new function to call this block of code.

btree/btree.c
ydb242: Initially, it was removed to accommodate segaddr since we were not using node descriptor. Shashank suggested that this ergo check is not useful. Hence, after reversing the segaddr changes, I didnt update this function.

RadhaGulhane13, line 2363: Is this the final definition of ff_invariant

btree/btree.c
ydb242: This particular variable "struct ff_head *h" was added to so that segaddr data can be assigned to it.
the code was updated to: struct ff_head *h =  segaddr_addr(addr);
On reversing those changes, I just replaced "segaddr_addr(addr)" with "ff_data(node)"
Incase in future, we again decide to go with segaddr, it will be easy to update.

btree/btree.c
ydb242: This particular variable "struct ff_head *h" was added to so that segaddr data can be assigned to it.
the code was updated to: struct ff_head *h = segaddr_addr(addr);
On reversing those changes, I just replaced "segaddr_addr(addr)" with "ff_data(node)"
Incase in future, we again decide to go with segaddr, it will be easy to update.

btree/btree.c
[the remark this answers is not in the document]
ydb242: Oh Yes.
I will update it.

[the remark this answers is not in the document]
ydb242: Yes

ydb242: Yes

btree/btree.c
ydb242: ok

btree/btree.c
ydb242: No, This was just done to accomodate segaddr changes, but for now I kept them as it is.
In the future jira, that you will work on, will have the final definition of ff_invariant.

RadhaGulhane13: ok

NikhilBirgade: We need the check till we have efficient invariant function. 
@RadhaGulhane13 Please confirm if this check is needed or not?

btree/btree.c
NikhilBirgade: In that code also you could have just replaced ff_data with segaddr_addr.

btree/btree.c
NikhilBirgade: In that code also you could have just replaced ff_data with segaddr_addr.

btree/btree.c
ydb242: Ok. I will update it!

btree/btree.c
ydb242: Ok. I will update it!

NikhilBirgade (approved):

btree/btree.c
RadhaGulhane13: I have added this check to ignore this check : h->ff_level > 0, h->ff_used > 0
If @ydb242  is removing this check, h->ff_level > 0, h->ff_used > 0, no need to check the flag. 
But on the other hand, if we make this change, invariant does not seems to be inefficient. So, the other option is if you want to remove flag check , you can remove h->ff_level>0 check but keep h->ff_used>=0 check as it is .
But I think we need the suggestion from @shashank-parulekar  as well .

ydb242: @RadhaGulhane13 https://github.com/Seagate/cortx-motr/pull/819#discussion_r664072942

RadhaGulhane13: okay .

RadhaGulhane13 (approved):

btree/btree.c
```diff
@@ -1772,6 +1762,29 @@ static void tree_put(struct td *tree)
...
+	return nxt;
+}
+
+static void seg_node_free(struct node_op *op, int shift, struct m0_be_tx *tx,
+			     int nxt)
+{
+	m0_free_aligned(segaddr_addr(&op->no_addr), 1ULL << shift, shift);
+	/** Capture in transaction */
+}
+
```
shashank-parulekar, line 1787: It seems you have just changed the names of the functions mem_node_*() to seg_node_*() !!!
How will that help !!!

shashank-parulekar (changes requested):

btree/btree.c
ydb242: Updated the code.

btree/btree.c
```diff
@@ -2387,7 +2325,7 @@ static struct ff_head *ff_data(const struct nd *node)
 
 static void *ff_key(const struct nd *node, int idx)
 {
-	struct ff_head *h    = ff_data(node);
+	struct ff_head *h  = ff_data(node);
```
shashank-parulekar, line 2328: Fix the '=' alignment here and also in the below functions.

ydb242: Done.

shashank-parulekar (approved):

PR #827, opened 2021-07-06
EOS-21221: Fix deadlock at P_LOCKALL state.

NikhilBirgade:
Issue: A thread was waiting to take tree lock at P_LOCKALL. In the meantime
another thread took lock, inserted records and changed height of the tree.
When thread one took tree lock, it found height is changed and returned to next
state without unlocking the tree.
Fix: Changed the P_LOCKALL and P_SETUP order for lockall case.

### DIFF
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -3549,16 +3549,13 @@ static int64_t btree_put_kv_tick(struct m0_sm_op *smop)
 		if (!path_check(oi, tree, &bop->bo_rec.r_key.k_cookie)) {
 			oi->i_trial++;
 			if (oi->i_trial >= MAX_TRIALS) {
-				if (bop->bo_flags & BOF_LOCKALL) {
-					lock_op_unlock(bop->bo_arbor->t_desc);
-					return fail(bop, -ETOOMANYREFS);
-				} else {
-					bop->bo_flags |= BOF_LOCKALL;
-					lock_op_unlock(tree);
-					return m0_sm_op_sub(&bop->bo_op,
-							    P_CLEANUP,
-							    P_LOCKALL);
-				}
+				M0_ASSERT_INFO((bop->bo_flags & BOF_LOCKALL) ==
+					       0, "Put record failure in tree"
+					       "lock mode");
+				bop->bo_flags |= BOF_LOCKALL;
+				lock_op_unlock(tree);
+				return m0_sm_op_sub(&bop->bo_op, P_CLEANUP,
+						    P_LOCKALL);
 			}
 			if (oi->i_height != tree->t_height) {
 				/* If height has changed. */
@@ -4201,16 +4198,13 @@ static int64_t btree_get_kv_tick(struct m0_sm_op *smop)
 		if (!path_check(oi, tree, &bop->bo_rec.r_key.k_cookie)) {
 			oi->i_trial++;
 			if (oi->i_trial >= MAX_TRIALS) {
-				if (bop->bo_flags & BOF_LOCKALL) {
-					lock_op_unlock(bop->bo_arbor->t_desc);
-					return fail(bop, -ETOOMANYREFS);
-				} else {
-					bop->bo_flags |= BOF_LOCKALL;
-					lock_op_unlock(tree);
-					return m0_sm_op_sub(&bop->bo_op,
-							    P_CLEANUP,
-							    P_LOCKALL);
-				}
+				M0_ASSERT_INFO((bop->bo_flags & BOF_LOCKALL) ==
+					       0, "Get record failure in tree"
+					       "lock mode");
+				bop->bo_flags |= BOF_LOCKALL;
+				lock_op_unlock(tree);
+				return m0_sm_op_sub(&bop->bo_op, P_CLEANUP,
+						    P_LOCKALL);
 			}
 			if (oi->i_height != tree->t_height) {
 				/* If height has changed. */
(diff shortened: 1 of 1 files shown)


btree/btree.c
```diff
@@ -3554,8 +3552,13 @@ static int64_t btree_put_kv_tick(struct m0_sm_op *smop)
 				if (bop->bo_flags & BOF_LOCKALL) {
 					lock_op_unlock(bop->bo_arbor->t_desc);
 					return fail(bop, -ETOOMANYREFS);
-				} else
+				} else {
```
shashank-parulekar, line 3555: you need to reset the oi->i_trial to 0 here when setting the BOF_LOCKALL flag.

shashank-parulekar: This **if (bop->bo_flags & BOF_LOCKALL) {** is incorrect. Instead you should have an assert to make sure that if path_check fails then BOF_LOCKALL is NOT set. If the path changes even with LOCKALL then we have some serious issue which we need to know of during our debugging.

btree/btree.c
```diff
@@ -4206,8 +4204,13 @@ static int64_t btree_get_kv_tick(struct m0_sm_op *smop)
...
```
shashank-parulekar, line 4207: same comments as above.

shashank-parulekar (changes requested):

btree/btree.c
NikhilBirgade: I will add assert if we `path_check` fails when `BOF_LOCKALL` flag is set. 
I think it will be better to not reset `oi->i_trial` to 0 when setting the BOF_LOCKALL flag. In case of any failure we can refer to this value if needed.

btree/btree.c
NikhilBirgade: Noted.

btree/btree.c
```diff
@@ -3429,21 +3429,19 @@ static int64_t btree_put_kv_tick(struct m0_sm_op *smop)
 		if (cookie_is_valid(tree, &bop->bo_rec.r_key.k_cookie) &&
 		    !node_isoverflow(oi->i_cookie_node))
 			return P_LOCK;
-		else
-			return P_SETUP;
-	case P_SETUP: {
+		/** Fall through if cookie is invalid or node overflow. */
+	case P_LOCKALL:
+		if (bop->bo_flags & BOF_LOCKALL)
```
RadhaGulhane13, line 3434: we should have assert instead of if (bop->bo_flags & BOF_LOCKALL), because we should reach P_LOCKALL only if BOF_LOCKALL flag is set
and we should not have fall-through from BOF_LOCKALL, else there needs to transitions from P_LOCKALL-->P_NEXTDOWN

NikhilBirgade: I can do that too. In that case I need to avoid falling from P_COOKIE to P_LOCKALL and jump from P_COOKIE to P_SETUP in case of cookie is invalid.

RadhaGulhane13: Yes.

shashank-parulekar (approved):

btree/btree.c
```diff
@@ -3469,19 +3479,17 @@ static int64_t btree_put_kv_tick(struct m0_sm_op *smop)
...
-	case P_SETUP: {
+	case P_LOCKALL:
+		M0_ASSERT(bop->bo_flags & BOF_LOCKALL);
+		return lock_op_init(&bop->bo_op, &bop->bo_i->i_nop,
+				    bop->bo_arbor->t_desc, P_SETUP);
+	case P_SETUP:
 		oi->i_height = tree->t_height;
 		level_alloc(oi, oi->i_height);
 		if (oi->i_level == NULL)
 			return fail(bop, M0_ERR(-ENOMEM));
```
RadhaGulhane13, line 3490: If lock_acquired flag is set and oi->i_level == NULL, unlock the tree lock at line 3490and then return fail()

btree/btree.c
```diff
@@ -3596,21 +3604,24 @@ static int64_t btree_put_kv_tick(struct m0_sm_op *smop)
 		if (!path_check(oi, tree, &bop->bo_rec.r_key.k_cookie)) {
 			oi->i_trial++;
 			if (oi->i_trial >= MAX_TRIALS) {
-				if (bop->bo_flags & BOF_LOCKALL) {
-					lock_op_unlock(bop->bo_arbor->t_desc);
-					return fail(bop, -ETOOMANYREFS);
-				} else
-					bop->bo_flags |= BOF_LOCKALL;
+				M0_ASSERT_INFO((bop->bo_flags & BOF_LOCKALL) ==
```
RadhaGulhane13, line 3607: Do you need this ASSERT or you forgot to remove from here. Please check once

btree/btree.c
```diff
@@ -3596,21 +3604,24 @@ static int64_t btree_put_kv_tick(struct m0_sm_op *smop)
...
+				lock_op_unlock(tree);
+				return m0_sm_op_sub(&bop->bo_op, P_CLEANUP,
+						    P_LOCKALL);
 			}
 			if (oi->i_height != tree->t_height) {
 				/* If height has changed. */
 				lock_op_unlock(tree);
 				return m0_sm_op_sub(&bop->bo_op, P_CLEANUP,
-					            P_SETUP);
+				                    P_SETUP);
```
RadhaGulhane13, line 3619: Use of spaces instead of tabs

btree/btree.c
NikhilBirgade: Done.

btree/btree.c
NikhilBirgade: We need it. Suppose we have already set LOCKALL and still path check fails then there is something seriously wrong with code.

btree/btree.c
NikhilBirgade: Done.

RadhaGulhane13 (approved):

Issue #826, opened 2021-07-06
m0_op_launch fails when running go/mkv


mengwanguc:
Hello,

I was running [go/mkv](https://github.com/Seagate/cortx-motr/blob/main/bindings/go/mio/mkv.go) with the following command:

`sudo ./mkv -hax 10.52.3.172@tcp:12345:34:1 -ep 10.52.3.172@tcp:12345:33:1000 --prof 0x7000000000000001:0 -proc 0x7200000000000001:64 111:12345670 aaa aaaaaaa`

Basically here I want to put a key-value (aaa-aaaaaaa) to idx `111:12345670`. However, I got the following error:

```
motr[58733]:  2e70  FATAL  [lib/assert.c:50:m0_panic]  panic: (({ typeof (ifid_type) __x = (ifid_type); ((__x) == (m0_dix_fid_type.ft_id) || ((__x) == (m0_cas_index_fid_type.ft_id))); })) at idx_is_distributed() (motr/idx_dix.c:103)  [git: Cred-Test-9-g70701ee] 
Motr panic: (({ typeof (ifid_type) __x = (ifid_type); ((__x) == (m0_dix_fid_type.ft_id) || ((__x) == (m0_cas_index_fid_type.ft_id))); })) at idx_is_distributed() motr/idx_dix.c:103 (errno: 22) (last failed: none) [git: Cred-Test-9-g70701ee] pid: 58733  
../../../motr/.libs/libmotr.so.2(m0_arch_backtrace+0x20)[0x7f3e9053df30]
../../../motr/.libs/libmotr.so.2(m0_arch_panic+0xe6)[0x7f3e9053e0e6]
../../../motr/.libs/libmotr.so.2(+0x370f04)[0x7f3e9052cf04]
../../../motr/.libs/libmotr.so.2(+0x3b68e4)[0x7f3e905728e4]
../../../motr/.libs/libmotr.so.2(+0x3b7dce)[0x7f3e90573dce]
../../../motr/.libs/libmotr.so.2(+0x3b7ff7)[0x7f3e90573ff7]
../../../motr/.libs/libmotr.so.2(+0x3b8d7a)[0x7f3e90574d7a]
../../../motr/.libs/libmotr.so.2(m0_op_launch_one+0x11b)[0x7f3e9055449b]
../../../motr/.libs/libmotr.so.2(**m0_op_launch**+0x44)[0x7f3e905545e4]
./mkv[0x466750]
SIGABRT: abort

[...]

PR #828, opened 2021-07-06
go/mkv: add check for index fid format

andriytk:
Check index fid provided by user to make sure it starts with `0x78`.

This should help to avoid issues like https://github.com/Seagate/cortx-motr/issues/826.

### DIFF
--- a/bindings/go/mio/mkv.go
+++ b/bindings/go/mio/mkv.go
@@ -45,12 +45,19 @@ type Mkv struct {
     idx    *C.struct_m0_idx
 }
 
+func uint128_to_fid(u C.struct_m0_uint128) (f C.struct_m0_fid) {
+    f.f_container = u.u_hi
+    f.f_key       = u.u_lo
+    return f
+}
+
 func (mkv *Mkv) idxNew(id string) (err error) {
     mkv.idxID, err = ScanID(id)
     if err != nil {
         return err
     }
-    if (mkv.idxID.u_hi >> 56) != 0x78 {
+    fid := uint128_to_fid(mkv.idxID)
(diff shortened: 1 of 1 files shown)


bindings/go/mio/mkv.go
```diff
@@ -50,6 +50,10 @@ func (mkv *Mkv) idxNew(id string) (err error) {
     if err != nil {
         return err
     }
+    if (mkv.idxID.u_hi >> 56) != 0x78 {
```
huanghua78, line 53: Do you think we forcibly XORing the 0x78 will be better?
```
mkv.idxID.u_hi |= 0x7800000000000000;
```
Or maybe not.
"MSByte", what does it mean?

We need to say that, 0x78 is the type of object fid type.

andriytk: > forcibly XORing the 0x78 will be better?

Not sure. User may use this utility for debug purposes and use the same index in other apps, so better being clear about its fid.

MSByte is Most Significant Byte, see https://en.wikipedia.org/wiki/Bit_numbering#Most_significant_byte.

> need to say that, 0x78 is the type of object fid type.

May be. Wouldn't that be too much for the error msg?

huanghua78 (approved):
Looks fine to me

bindings/go/mio/mkv.go
madhavemuri: [Suggestion]
It would be better if code can be changed in a following way,

 if (m0_fid_tget(dix_fid) == m0_dix_fid_type.ft_id )

madhavemuri (approved):

bindings/go/mio/mkv.go
andriytk: Good idea, done.

PR #829, opened 2021-07-06
EOS-22162: config script changes for lvm names.

atulsdeshmukh2312:

### DIFF
--- a/scripts/install/opt/seagate/cortx/motr/bin/motr_setup
+++ b/scripts/install/opt/seagate/cortx/motr/bin/motr_setup
@@ -123,7 +123,7 @@ class PostInstallCmd(Cmd):
             config_lvm(self)
         else:
             self.logger.error("There are LVM configurations present on the system. "
-                              "Please wipe them and re-run the step.\n")
+                              "Skipping LVM creation.\n")
 
 class ConfigCmd(Cmd):
     """Config Setup Cmd"""
--- a/scripts/install/opt/seagate/cortx/motr/bin/motr_setup
+++ b/scripts/install/opt/seagate/cortx/motr/bin/motr_setup
@@ -123,8 +123,6 @@ class PostInstallCmd(Cmd):
             config_lvm(self)
         else:
             self.logger.error("There are LVM configurations present on the system\n")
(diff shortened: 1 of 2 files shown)


scripts/install/opt/seagate/cortx/motr/bin/motr_setup
```diff
@@ -122,10 +122,8 @@ class PostInstallCmd(Cmd):
         if not lvm_exist(self):
             config_lvm(self)
         else:
-            self.logger.warning("lvm already exists. "
-                             "Skipping lvm creation.\n")
-            self.logger.warning("Updating bgsize.\n")
-            update_bgsize(self)
+            self.logger.error("There are LVM configurations present on the system\n")
+            sys.exit()
```
madhavemuri, line 126: Post install can be called again, so in such cases LVM's created in previous call are valid.

Just make sure that if valid LVM's are present proceed ahead, otherwise exit with error message that "Invalid/unclean LVM's are present and need to wipe them and re-run the step."

scripts/install/opt/seagate/cortx/motr/bin/motr_setup
```diff
@@ -123,7 +123,7 @@ class PostInstallCmd(Cmd):
             config_lvm(self)
         else:
             self.logger.error("There are LVM configurations present on the system. "
-                              "Please wipe them and re-run the step.\n")
```
madhavemuri: Can we ensure here, if they are valid LVM's or not.
If not then exit here with message as clear the LVM's and created the valid ones.

atulsdeshmukh2312: We list LV names from lvdisplay and doing some grepping on it to extract LV path.
Once we get all LV paths, we check with cortx paths like '/dev/vg_srvnode-1_md1/lv_raw_md1' and 
/dev/vg_srvnode-1_md1/lv_main_swap1

Since we are getting this list from lvm commands, we assume that they are valid.
We have not validated that if these found cortx lvm paths are from our md disks or from some other.

madhavemuri (approved):

Issue #830, opened 2021-07-06
list all keys in a client index key-value store


mengwanguc:
Hello,

I'm wondering is it possible to list all the keys in a client index key-value store?

By reading [idx.h](https://github.com/Seagate/cortx-motr/blob/main/motr/idx.h), looks like it's not supported yet, as it only implements 4 types of index operations (PUT, GET, DEL, NEXT).
But I want to double check in case I missed anything.

---
The background is that I want to implement a [go-datastore](https://github.com/ipfs/go-datastore/blob/master/datastore.go#L94) based on motr, and the datastore requires a [`Query(q query.query)`](https://github.com/ipfs/go-datastore/blob/master/query/query.go) function.

I assume currently index Key-Value store doesn't support queries. 

So a native way to implement this query function is to list all the keys, and then check the keys one by one to see if each key satisfies the query.

That's why I want to check if it's possible to list all the keys, which sounds easier to implement than query functionalities.

hessio:
Hi @mengwanguc I think you are right it is not available.

cc @andriytk

andriytk:
NEXT operation is there exactly for this purpose.

mengwanguc:
> NEXT operation is there exactly for this purpose.

Ah I see. Do you mean NEXT is like a circular list iterator so that we can do:

k = NEXT(firstKey)
while (k != firstKey) {
    k = NEXT(k)
}

I thought it's a one-direction list so NEXT will reach end somehow. But if it's a circular iterator then it makes a lot of sense.

mengwanguc:
Hi @andriytk , I think my another question is how can user get the start key, if the user had no knowledge of the key-value store? (i.e. the user doesn't know any key at all)

andriytk:
Read more about it here - https://github.com/Seagate/cortx-motr/blob/e822f3af0eab46fc76cb52c6274b94cb5eaa3022/motr/client.h#L1445-L1458

mengwanguc:
> cortx-motr/motr/client.h

Thanks! This answers all my questions!

johnbent:
@mengwanguc , can you please try to figure out why you didn't find this info yourself?  I don't think it is your failure!  I think the documentation needs to be improved.  So, can you please retrace your steps searching yourself for this information and find the place where it should be but isn't?  And then add it?  And then please add @andriytk as a reviewer to the PR?

mengwanguc:
Hi @johnbent , sure!

[...]

PR #835, opened 2021-07-07
Update idx.h about NEXT operation info

mengwanguc:
Add information about what will happen if the user doesn't specify start key for the NEXT operation, and what if the specified start key doesn't exist in the index.

Add this info because otherwise users may wrongly assume that they must specify start keys for the NEXT operation, which is not true.

A corresponding issue was reported at #830.

### DIFF
--- a/motr/idx.h
+++ b/motr/idx.h
@@ -46,7 +46,10 @@
  *     - DEL: given a set of keys, delete the matching records from the index;
  *
  *     - NEXT: given a set of keys, return the records with the next (in the
- *       ascending key order) keys from the index.
+ *       ascending key order) keys from the index. If the user doesn't specify 
+ *       the start key, then start key is treated as the smallest possible key
+ *       of the index. If start key doesn't exist in the index, then retrieved
+ *       records will start with the smallest key following the starting key.
  *
  * @todo Currently, the user can specify only one start key for NEXT operation,
  * not a set.


cortx-admin:

motr/idx.h
```diff
@@ -46,7 +46,10 @@
  *     - DEL: given a set of keys, delete the matching records from the index;
  *
  *     - NEXT: given a set of keys, return the records with the next (in the
- *       ascending key order) keys from the index.
+ *       ascending key order) keys from the index. If the user doesn't specify 
+ *       the start key, then start key is treated as the smallest possible key
```
johnbent, line 50: I haven't looked at the rest of the documentation so my comment maybe can be ignored.  Does the word "smallest" make sense for keys?  It talks about "ascending" so perhaps "lowest" would be better.

mengwanguc: Thanks John. I used "smallest" because in [client.h](https://github.com/Seagate/cortx-motr/blob/e822f3af0eab46fc76cb52c6274b94cb5eaa3022/motr/client.h#L1445-L1458) they used "smallest".

johnbent:
Cool.  Carry on then.  🙂
________________________________

mengwanguc:
Hi @andriytk , can you please review this PR?
(Unfortunately I'm not authorized to request specific reviewers, so I have to @ you here in the comment :)

andriytk:
If this is just a copy-paste from https://github.com/Seagate/cortx-motr/blob/e822f3af0eab46fc76cb52c6274b94cb5eaa3022/motr/client.h#L1445-L1453, I'm not sure if it is necessary. Let's keep the detailed description of the interface in one place - at client.h.

mengwanguc:
> If this is just a copy-paste from
> 
> https://github.com/Seagate/cortx-motr/blob/e822f3af0eab46fc76cb52c6274b94cb5eaa3022/motr/client.h#L1445-L1453
> 
> , I'm not sure if it is necessary. Let's keep the detailed description of the interface in one place - at client.h.

Thanks Andriy,  yeah that makes sense.

Can you please take a look at #838 ? There I add a reference to client.h instead.

Also, I found that idx.h already has a copy-paste of comments from client.h. So I removed the duplicate comments in that PR.

PR #842, opened 2021-07-09
Changes in P_ALLOC phase

RadhaGulhane13:
Fixed bug in P_ALLOC phase:
Freeing node which is allocated and not stored or used.

some other changes:
using oi->pivot for P_ALLOC phase instead of oi->i_used to avoid resetting of oi->i_used

### DIFF
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -3618,8 +3618,9 @@ static struct m0_sm_state_descr btree_states[P_NR] = {
 	[P_NEXTDOWN] = {
 		.sd_flags   = 0,
 		.sd_name    = "P_NEXTDOWN",
-		.sd_allowed = M0_BITS(P_NEXTDOWN, P_ALLOC_REQUIRE, P_STORE_CHILD,
-				      P_CLEANUP, P_SETUP, P_LOCK, P_SIBLING),
+		.sd_allowed = M0_BITS(P_NEXTDOWN, P_ALLOC_REQUIRE,
+				      P_STORE_CHILD, P_SIBLING, P_CLEANUP,
+				      P_LOCK),
 	},
 	[P_SIBLING] = {
 		.sd_flags   = 0,
@@ -3711,7 +3712,6 @@ static struct m0_sm_trans_descr btree_trans[] = {
 	{ "get-nextdown-next", P_NEXTDOWN, P_LOCK },
 	{ "iter-nextdown-sibling", P_NEXTDOWN, P_SIBLING },
 	{ "kvop-nextdown-failed", P_NEXTDOWN, P_CLEANUP },
-	{ "kvop-nextdown-setup", P_NEXTDOWN, P_SETUP},
 	{ "iter-sibling-repeat", P_SIBLING, P_SIBLING },
 	{ "iter-sibling-next", P_SIBLING, P_LOCK },
 	{ "iter-sibling-failed", P_SIBLING, P_CLEANUP },
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -1234,7 +1234,9 @@ struct m0_btree_oimpl {
 	/** Level from where sibling nodes needs to be loaded. **/
 	int             i_pivot;
 
-	int             i_alloc;
+	/** i_alloc_lev will be used for P_ALLOC_REQUIRE and P_ALLOC_STORE phase
+	 * to indicate current level index. **/
+	int             i_alloc_lev;
 
 	/** Store node_find() output. */
 	bool            i_key_found;
@@ -3465,11 +3467,10 @@ static int64_t btree_put_kv_tick(struct m0_sm_op *smop)
 				if (oi->i_key_found)
 					return P_LOCK;
 				/**
-				 * i_alloc will be used for P_ALLOC_REQUIRE and
-				 * P_ALLOC_STORE phase to indicate current level
-				 * index.
-				 **/
-				oi->i_alloc = oi->i_used;
+				 * Initialized i_alloc_lev to level of leaf
+				 * node.
+				 */
+				oi->i_alloc_lev = oi->i_used;
 				return P_ALLOC_REQUIRE;
 			}
 		} else {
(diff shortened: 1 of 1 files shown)


NikhilBirgade (approved):

btree/btree.c
```diff
@@ -2966,7 +2971,7 @@ static void level_cleanup(struct m0_btree_oimpl *oi, struct m0_be_tx *tx)
 	 * already taken by this thread.
 	 */
 	int i;
-	for (i = 0; i <= oi->i_used; ++i) {
+	for (i = 0; i < oi->i_height; ++i) {
```
shashank-parulekar: Why did we move from i_used to i_height ?

btree/btree.c
```diff
@@ -3547,46 +3466,96 @@ static int64_t btree_put_kv_tick(struct m0_sm_op *smop)
 				node_unlock(lev->l_node);
 				if (oi->i_key_found)
 					return P_LOCK;
-				return P_ALLOC;
+				/**
+				 * Initialized i_alloc_lev to level of leaf
```
shashank-parulekar, line 3383: s/Initialized/Initialize/

btree/btree.c
```diff
@@ -3547,46 +3466,96 @@ static int64_t btree_put_kv_tick(struct m0_sm_op *smop)
...
-		bool alloc = false;
+	case P_ALLOC_REQUIRE:{
 		do {
-			lev = &oi->i_level[oi->i_used];
-			/**
-			 * Validate node to dertmine if lev->l_node is still
-			 * exists.
-			 */
+			lev = &oi->i_level[oi->i_alloc_lev];
 			if (!node_isvalid(lev->l_node)) {
```
shashank-parulekar, line 3396: Remove '{' for single-line if-statement.

btree/btree.c
```diff
@@ -3547,46 +3466,96 @@ static int64_t btree_put_kv_tick(struct m0_sm_op *smop)
...
+					oi->i_nop.no_opc = NOP_ALLOC;
+					return node_alloc(&oi->i_nop, tree,
+							  shift,
+							  lev->l_node->n_type,
+							  ksize, vsize,
+							  lock_acquired,
+							  bop->bo_tx,
+							  P_ALLOC_STORE);
+
+				} else if (oi->i_extra_node == NULL){
```
NikhilBirgade, line 3543: spacing `({`

btree/btree.c
```diff
@@ -3547,46 +3466,96 @@ static int64_t btree_put_kv_tick(struct m0_sm_op *smop)
...
+					oi->i_extra_node = oi->i_nop.no_node;
+					return P_LOCK;
+				} else
+					M0_ASSERT(0);
```
NikhilBirgade, line 3547: For root level , it will come here only if `l_alloc` is Null of `i_extra_node` is Null. I think we can eliminate this else check

shashank-parulekar (changes requested):

btree/btree.c
RadhaGulhane13: Yes else check can be removed, but it is added intentionally to ensure that we should not reach this step.

btree/btree.c

btree/btree.c
```diff
@@ -3547,46 +3466,96 @@ static int64_t btree_put_kv_tick(struct m0_sm_op *smop)
...
+				/**
+				 * Depending on the level of node, shift can be
+				 * updated.
+				 */
+				int ksize = node_keysize(lev->l_node);
+				int vsize = node_valsize(lev->l_node);
+				int shift = node_shift(lev->l_node);
+				oi->i_nop.no_opc = NOP_ALLOC;
+				return node_alloc(&oi->i_nop, tree,
+						  shift,lev->l_node->n_type,
```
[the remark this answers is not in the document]

btree/btree.c
```diff
@@ -3650,18 +3618,24 @@ static struct m0_sm_state_descr btree_states[P_NR] = {
 	[P_NEXTDOWN] = {
 		.sd_flags   = 0,
 		.sd_name    = "P_NEXTDOWN",
-		.sd_allowed = M0_BITS(P_NEXTDOWN, P_ALLOC, P_STORE_CHILD,
+		.sd_allowed = M0_BITS(P_NEXTDOWN, P_ALLOC_REQUIRE, P_STORE_CHILD,
```
ydb242, line 3621: Are we within 80 char limit here?

btree/btree.c
RadhaGulhane13: So, it might be a corner case scenario.
Consider the scenario where at first traversal we find the height of tree 4 and at worst case allocated new nodes at all the levels from 0 to 3 level. Now in P_CHECK, we found that path_check function fails but height is same. Now, we start traversing again from P_DOWN , now consider height of tree decreased, now height is 3 i.e. our oi->i_used will be maximum 2. so, consider we havn't used any of the allocated node allocated at the time of 1st traversal ,  so ideally we should free all the allocated node from level 0 to 3 .But for second traversal oi->i_used will be 2.
 Therefore , we should used oi->i_height at time of node_free instead of oi->i_used.

btree/btree.c
```diff
@@ -3460,46 +3379,95 @@ static int64_t btree_put_kv_tick(struct m0_sm_op *smop)
...
-		oi->i_used = oi->i_height - 1;
+			}
+			oi->i_alloc_lev--;
+		} while (oi->i_alloc_lev >= 0);
 		return P_LOCK;
 	}
+	case P_ALLOC_STORE: {
+		lev = &oi->i_level[oi->i_alloc_lev];
+		if (oi->i_nop.no_op.o_sm.sm_rc == 0) {
+			if (oi->i_alloc_lev == 0) {
```
NikhilBirgade, line 3425: You can rearrange the code as below to avoid multiple level of indentation.

```
if (oi->i_nop.no_op.o_sm.sm_rc != 0) {
                      if (lock_acquired)
				lock_op_unlock(tree);
			node_op_fini(&oi->i_nop);
			return fail(bop, oi->i_nop.no_op.o_sm.sm_rc);
}
lev = &oi->i_level[oi->i_alloc_lev];
if (oi->i_alloc_lev == 0) {
     ------
     ------
}

btree/btree.c

shashank-parulekar (approved):

ydb242 (approved):

PR #844, opened 2021-07-12
Add node validation check in node_get().

NikhilBirgade:
Issue: Thread 1 is waiting to take global lock in node_get(). In the meantime
thread 2 deleted the same node. Thread 1 took lock and failed as the node is
deleted.
Fix: Add node validation in node_get(). If node validation fails, restart
respective tick function.

### DIFF
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -6345,7 +6345,6 @@ static void btree_ut_kv_oper_thread_handler(struct btree_ut_thread_info *ti)
 			for (i = 1; i < ARRAY_SIZE(value); i++)
 				value[i] = value[0];
 
-			M0_SET0(&kv_op);
 			M0_BTREE_OP_SYNC_WITH_RC(&kv_op,
 						 m0_btree_put(tree, &rec,
 							      &ut_cb, 0,
@@ -6384,7 +6383,6 @@ static void btree_ut_kv_oper_thread_handler(struct btree_ut_thread_info *ti)
 
 		get_data.check_value = true; /** Compare value with key */
 
-		M0_SET0(&kv_op);
 		M0_BTREE_OP_SYNC_WITH_RC(&kv_op,
 					 m0_btree_get(tree,
 						      &rec.r_key, &ut_get_cb,
@@ -6394,7 +6392,6 @@ static void btree_ut_kv_oper_thread_handler(struct btree_ut_thread_info *ti)
 		keys_found_count++;
 
 		while (1) {
-			M0_SET0(&kv_op);
 			M0_BTREE_OP_SYNC_WITH_RC(&kv_op,
 						 m0_btree_iter(tree,
 							       &rec.r_key,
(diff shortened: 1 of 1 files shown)


btree/btree.c
```diff
@@ -6368,6 +6382,7 @@ static void btree_ut_kv_oper_thread_handler(struct btree_ut_thread_info *ti)
 			for (i = 1; i < ARRAY_SIZE(value); i++)
 				value[i] = value[0];
 
+			M0_SET0(&kv_op);
```
shashank-parulekar, line 6385: Is this needed ? The reason is if similar code is present in Motr where the m0_btree_op is not initialized then will we crash?

shashank-parulekar (approved):

btree/btree.c
```diff
@@ -1812,15 +1812,24 @@ static int64_t node_get(struct node_op *op, struct td *tree,
 	 * implemented
 	 * */
 	nt = &fixed_format;
```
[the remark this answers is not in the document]
NikhilBirgade: Yes. I had added as a TODO item. I thought of added it with adding when we will add other node types. I will add the function in this PR.

btree/btree.c
NikhilBirgade: While debugging this crash, had doubts that we might be passing uninitialized `kv_op` to kv operations and it could have garbage data which might be causing this issue. Hence initialized it as a safety measure. But yes, it is optional as I have not seen any issues with uninitialized `kv_op`

btree/btree.c
```diff
@@ -1812,15 +1812,24 @@ static int64_t node_get(struct node_op *op, struct td *tree,
...
 	 * Replace it with a different global lock once hash
 	 * functionality is implemented.
 	 */
 	m0_rwlock_write_lock(&list_lock);
+	/**
+	 * If the node was deleted before node_get can acquire list_lock, then
+	 * restart the tick funcions.
+	 */
+	if (!nt->nt_isvalid(addr)) {
+		op->no_op.o_sm.sm_rc = ECHILD;
```
RadhaGulhane13, line 1827: Please check if error code should be  M0_ERR(-ECHILD) format instead of ECHILD. Please make similar changes at line 1837

NikhilBirgade: Ok.

btree/btree.c
shashank-parulekar: Remove this so that if we ever have issues related to not initializing the op variables with '0' then we need to catch them in our ut's and fix the m0_btree_put() to reset all the members in kv_op if it needs to.

shashank-parulekar (changes requested):

btree/btree.c
NikhilBirgade: Ok.

RadhaGulhane13 (approved):

shashank-parulekar (approved):

ydb242 (approved):

PR #855, opened 2021-07-15
Eos 21225: invariant update

RadhaGulhane13:
- Updated definition of invariant
- Added expensive invariant which will validate if keys are in ascending order

### DIFF
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -2424,7 +2424,7 @@ static bool segaddr_header_isvalid(const struct segaddr *addr)
 	struct ff_head       *h = segaddr_addr(addr);
 	struct m0_format_tag  tag;
 
-	if (h->ff_fmt.hd_magic == 0)
+	if (h->ff_fmt.hd_magic != M0_FORMAT_HEADER_MAGIC)
 		return false;
 
 	m0_format_header_unpack(&tag, &h->ff_fmt);
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -2362,8 +2362,7 @@ static bool ff_rec_is_valid(const struct slot *slot)
 	   _0C(val_is_valid);
 }
 
-static bool ff_iskey_smaller(const struct nd *node, int prev_key_idx,
-			int next_key_idx)
+static bool ff_iskey_smaller(const struct nd *node, int cur_key_idx)
 {
 	struct ff_head          *h;
 	struct m0_btree_key      key_prev;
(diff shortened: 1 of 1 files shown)


btree/btree.c
```diff
@@ -2421,6 +2473,8 @@ static void ff_fini(const struct nd *node)
 		.ot_type          = 0,
 		.ot_footer_offset = offsetof(struct ff_head, ff_foot)
```
shashank-parulekar, line 2474: ot_footer_offset should also be set to 0.

shashank-parulekar (changes requested):

btree/btree.c
```diff
@@ -2357,11 +2362,55 @@ static bool ff_rec_is_valid(const struct slot *slot)
 	   _0C(val_is_valid);
 }
 
+static bool ff_iskey_smaller(const struct nd *node, int prev_key_idx,
+			int next_key_idx)
```
NikhilBirgade, line 2366: Alignment

btree/btree.c
```diff
@@ -2357,11 +2362,55 @@ static bool ff_rec_is_valid(const struct slot *slot)
 	   _0C(val_is_valid);
 }
 
+static bool ff_iskey_smaller(const struct nd *node, int prev_key_idx,
```
NikhilBirgade, line 2365: The function is checking record at current jey index with next key index. It will be better to have second parameter name as `cur_key_idx`

btree/btree.c
```diff
@@ -2357,11 +2362,55 @@ static bool ff_rec_is_valid(const struct slot *slot)
...
+
 static bool ff_invariant(const struct nd *node)
 {
 	const struct ff_head *h = ff_data(node);
 
-	return  _0C(h->ff_shift == segaddr_shift(&node->n_addr));
+	/* TBD: add check for h_tree_type after initializing it in node_init. */
+	return  _0C(h->ff_fmt.hd_magic == M0_FORMAT_HEADER_MAGIC) &&
+		_0C(h->ff_seg.h_node_type == BNT_FIXED_FORMAT) &&
+		_0C(h->ff_ksize != 0) && (h->ff_vsize != 0) &&
```
NikhilBirgade, line 2412: either have both ksize and vsize check in same _0C or and separate _0C for vsize.

btree/btree.c
RadhaGulhane13: made this change.

btree/btree.c

NikhilBirgade (approved):

shashank-parulekar (approved):

ydb242 (approved):

PR #861, opened 2021-07-19
EOS-22548: Credit calculation for btree callbacks

NikhilBirgade:

### DIFF
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -3087,7 +3087,6 @@ static int64_t btree_put_root_split_handle(struct m0_btree_op *bop,
 	node_done(&node_slot, bop->bo_tx, true);
 	node_seq_cnt_update(lev->l_node);
 	node_fix(lev->l_node, bop->bo_tx);
-	btree_callback_add(oi->i_callback, lev->l_node);
 	btree_callback_add(oi->i_callback, oi->i_extra_node);
 
 	/* Increase height by one */
@@ -4740,6 +4739,9 @@ static int64_t btree_del_resolve_underflow(struct m0_btree_op *bop)
 
 	node_move(root_child, lev->l_node, D_RIGHT, NR_MAX, bop->bo_tx);
 	M0_ASSERT(node_count_rec(root_child) == 0);
+
+	btree_callback_add(oi->i_callback, lev->l_node);
+	btree_callback_add(oi->i_callback, root_child);
 
 	lev->l_node->n_skip_rec_count_check = false;
 	oi->i_level[1].l_sibling->n_skip_rec_count_check = false;
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -619,6 +619,8 @@ enum {
 	MAX_KEY_SIZE             = 8,
 	MAX_VAL_SIZE             = 8,
 	MAX_TRIALS               = 3,
+	MAX_TREE_HEIGHT          = 5,
+	BTREE_CALLBACK_CREDIT    = MAX_TREE_HEIGHT * 2 + 1,
 	INTERNAL_NODE_VALUE_SIZE = sizeof(void *),
 };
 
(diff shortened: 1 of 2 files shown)


btree/btree.c
```diff
@@ -1247,6 +1249,14 @@ struct m0_btree_oimpl {
 	/** Node descriptor for cookie if it is going to be used. **/
 	struct nd      *i_cookie_node;
 
+	/**
+	 * Array to hold addresses of node descriptors which are going to be
+	 * captured in transaction. This will be used by transaction engine for
+	 * callbacks on transaction commit. Array size is the worst case based
+	 * on btree_callback_credit() calculation.
+	 */
+	void           *i_callback[BTREE_CALLBACK_CREDIT];
```
shashank-parulekar, line 1258: Do not assign the max array size here. This will not work if the current transaction is going to capture multiple trees and each of them is currently of MAX_HEIGHT. Instead the call to the credit calculator needs to be used for every tree.
Also the type of this should not be void * - instead it should be some structure which contains 2 memberes: callback function pointer and callback data pointer,

btree/btree.c
```diff
@@ -2933,6 +2969,25 @@ static void level_cleanup(struct m0_btree_oimpl *oi, struct m0_be_tx *tx)
 	m0_free(oi->i_level);
 }
 
+/**
+ * Adds unique node descriptor address to m0_btree_oimpl::i_callback.
+ */
+static void btree_callback_add(void **arr, void *addr)
```
shashank-parulekar, line 2975: This function needs to accept three parameters:
1) Tx in which to storre the callback information.
2) Callback function pointer.
3) Callback data to be passed to the callback function on transaction commit complete.

shashank-parulekar (changes requested):

btree/btree.c
```diff
@@ -3032,6 +3087,8 @@ static int64_t btree_put_root_split_handle(struct m0_btree_op *bop,
 	node_done(&node_slot, bop->bo_tx, true);
 	node_seq_cnt_update(lev->l_node);
 	node_fix(lev->l_node, bop->bo_tx);
+	btree_callback_add(oi->i_callback, lev->l_node);
```
RadhaGulhane13, line 3090: you can avoid  calling btree_callback_add(oi->i_callback, lev->l_node); as it will get already added in tree_put_makespace_phase at line no. 3224 or 3300

btree/btree.c
NikhilBirgade: Agree with the below point
```
Also the type of this should not be void * - instead it should be some structure which contains 2 memberes: callback function pointer and callback data pointer
```
For below point, this structure is in `m0_btree_oimpl`, so it will be called for each operation. And for multiple btrees credit calculator will also needs to be called for each btree. I will discuss with you offline.
```
Do not assign the max array size here. This will not work if the current transaction is going to capture multiple trees and each of them is currently of MAX_HEIGHT. Instead the call to the credit calculator needs to be used for every tree.
```

btree/btree.c
NikhilBirgade: the callback information will be stored in the tx capture based on i_callback.

btree/btree.c
```diff
@@ -4633,6 +4700,7 @@ static int64_t btree_del_resolve_underflow(struct m0_btree_op *bop)
 		}
 		node_seq_cnt_update(lev->l_node);
 		node_fix(node_slot.s_node, bop->bo_tx);
+		btree_callback_add(oi->i_callback, lev->l_node);
```
RadhaGulhane13, line 4703: you might need to add btree_callback_add(oi->i_callback, lev->l_node); and btree_callback_add(root_child); as well at line no.: 4746 or you might need to add btree_callback_add() in node_move for src and tgt node

shashank-parulekar (approved):

btree/btree.c
NikhilBirgade: ok.

btree/btree.c
NikhilBirgade: Since I do not have oimpl in node_move, added callbacks after node_move for root_child and lev->l_node.

RadhaGulhane13 (approved):

PR #869, opened 2021-07-20
EOS-23898 : motr-freespace-monitor: Add a delay between retry if capacity_info is 0

yeshpal-jain-seagate:
EOS-23898 : motr-freespace-monitor: Add a delay between retry if capacity_info

Problem : Hare may take sometime to retrieve capcacity info during
startup, if the capacity_info is not recieved, free space monitor would
timeout with error.
Solution: Add delay between retry, After 10 retries, if it still fails
than return failure.

Signed-off-by: Yeshpal Jain <yeshpal.jain@seagate.com>

### DIFF
--- a/scripts/install/usr/libexec/cortx-motr/motr-free-space-monitor
+++ b/scripts/install/usr/libexec/cortx-motr/motr-free-space-monitor
@@ -83,11 +83,13 @@ def execute_command(command):
         console_output = json.loads(process.stdout.decode('utf-8'))
         capacity_info = console_output.get('filesystem',{}).get('stats',{})
         if not capacity_info:
+            time.sleep(iem_alert_check_interval)
             continue
         return capacity_info
 
     msg = " Returned with error " + str(rc)
     log.error(f"Failed to process command after {iem_max_retry} retry '{command}':{msg}")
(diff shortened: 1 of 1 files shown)


scripts/install/usr/libexec/cortx-motr/motr-free-space-monitor
```diff
@@ -83,6 +83,7 @@ def execute_command(command):
         console_output = json.loads(process.stdout.decode('utf-8'))
         capacity_info = console_output.get('filesystem',{}).get('stats',{})
         if not capacity_info:
+            time.sleep(iem_alert_check_interval)
```
madhavemuri, line 86: Check  here whether ioservices are online or not, if not then log it.

yeshpal-jain-seagate: IO service status can be determined using cluster.log file and we would never know how many ioservices would be there to check, instead added logs to output the result of filesystem stats which will help us to know why the service failed.

madhavemuri (approved):

mehjoshi:
@yeshpal-jain-seagate , Is there a jira ticket for this? If not could you please create one so that we do not lose context of why this change was made?
Thanks,

mehjoshi (approved):
Looks good to me.

PR #870, opened 2021-07-20
EOS-22551: Build callback list in transaction.

NikhilBirgade:

### DIFF
--- a/be/tx.c
+++ b/be/tx.c
@@ -450,7 +450,6 @@ static void be_tx_state_move_ast(struct m0_be_tx *tx, enum m0_be_tx_state state)
 	}
 }
 
-#define ARRAY_ALLOC_NZ(arr, nr) ((arr) = m0_alloc_nz((nr) * sizeof ((arr)[0])))
 static int be_tx_memory_allocate(struct m0_be_tx *tx)
 {
 	int rc;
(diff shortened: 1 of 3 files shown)


be/tx.c
```diff
@@ -439,6 +450,7 @@ static void be_tx_state_move_ast(struct m0_be_tx *tx, enum m0_be_tx_state state)
 	}
 }
 
+#define ARRAY_ALLOC_NZ(arr, nr) ((arr) = m0_alloc_nz((nr) * sizeof ((arr)[0])))
```
shashank-parulekar, line 453: Why not call m0_alloc(). This way if we ever have a bug during development and access NULL memory then it will be easy to debug this.
You should call M0_ALLOC_ARR instead of this macro.in be_tx_memory_allocate()

shashank-parulekar (changes requested):

be/tx.c
NikhilBirgade: Since reg_area related functions were also using ARRAY_ALLOC_NZ macro, so I thought of using it. I will change it to M0_ALLOC_ARR.

shashank-parulekar: There is a different reason for them to use ARRAY_ALLOC_NZ since they would be capturing a 'lot' of data in the region to commit. In our case it is only a small list of callbacks and the data pointers. Also we need all the help we can to catch errors early during development.

shashank-parulekar (approved):

ydb242 (approved):

be/tx.c
NikhilBirgade: Okay.

PR #872, opened 2021-07-21
Expensive check changes

RadhaGulhane13:
- Removed expensive invariant check from ut.
- Added expensive invariant check in insert and delete routine whenever there is modification in node.
- Performing expensive invariant check in node lock mode.

### DIFF
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -3162,7 +3162,8 @@ static int64_t btree_put_root_split_handle(struct m0_btree_op *bop,
 
 	/* Increase height by one */
 	tree->t_height++;
-
+	M0_ASSERT(node_expensive_invariant(lev->l_node));
+	M0_ASSERT(node_expensive_invariant(oi->i_extra_node));
 	node_unlock(lev->l_node);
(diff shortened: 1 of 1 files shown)


btree/btree.c
```diff
@@ -3271,6 +3276,8 @@ static int64_t btree_put_root_split_handle(struct m0_btree_op *bop,
 	/* Increase height by one */
 	tree->t_height++;
 
+	M0_ASSERT(node_expensive_invariant(lev->l_node));
+	M0_ASSERT(node_expensive_invariant(oi->i_extra_node));
```
shashank-parulekar, line 3280: You need to have TBD message here which tells the expensive_invariant call should be removed.

shashank-parulekar (changes requested):

shashank-parulekar (approved):

NikhilBirgade (approved):

btree/btree.c
RadhaGulhane13: done.

PR #875, opened 2021-07-22
n_delayed flag changes

RadhaGulhane13:
renamed n_delayed_free to n_be_node_valid

### DIFF
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -1992,7 +1992,7 @@ static int64_t node_get(struct node_op *op, struct td *tree,
 	if (op->no_node != NULL &&
 	    op->no_node->n_addr.as_core == addr->as_core) {
 
-		if (op->no_node->n_be_node_valid) {
+		if (!op->no_node->n_be_node_valid) {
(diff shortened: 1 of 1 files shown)


btree/btree.c
```diff
@@ -2039,15 +2037,15 @@ static int64_t node_get(struct node_op *op, struct td *tree,
...
-		node->n_ref          = 1;
-		node->n_txref        = 0;
-		node->n_delayed_free = false;
+		node->n_addr          = *addr;
+		node->n_tree          = tree;
+		node->n_type          = nt;
+		node->n_seq           = m0_time_now();
+		node->n_ref           = 1;
+		node->n_txref         = 0;
+		node->n_be_node_valid = false;
```
NikhilBirgade, line 2046: Well, the variable name does not match with the use. If we have allocated the node then it means node is valid but here it is set to false.

RadhaGulhane13: Right, made changes according to new name.

shashank-parulekar (approved):

ydb242 (approved):

NikhilBirgade (approved):

PR #877, opened 2021-07-22
Eos 22550 : changes for capturing transactions for KV Delete operation.

RadhaGulhane13:
- calling P_CAPTURE before P_FREENODE
- unlocking tree after node_free operation.

### DIFF
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -7652,6 +7652,7 @@ static void ut_put_del_operation(void)
 					      .ksize = 8,
 					      .vsize = 8, };
 	struct m0_be_tx        *tx          = NULL;
+	struct m0_be_seg       *seg         = NULL;
 	struct m0_btree_op      b_op        = {};
 	struct m0_btree        *tree;
 	void                   *temp_node;
@@ -7681,7 +7682,7 @@ static void ut_put_del_operation(void)
 	temp_node = m0_alloc_aligned((1024 + sizeof(struct nd)), 10);
 	M0_BTREE_OP_SYNC_WITH_RC(&b_op,
 				 m0_btree_create(temp_node, 1024, &btree_type,
-						 nt, &b_op, tx));
+						 nt, &b_op, seg, tx));
 
 	tree = b_op.bo_arbor;
 	inc = false;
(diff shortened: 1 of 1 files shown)


btree/btree.c
```diff
@@ -5360,29 +5366,38 @@ static int64_t btree_del_kv_tick(struct m0_sm_op *smop)
...
-		lock_op_unlock(tree);
-		return m0_sm_op_sub(&bop->bo_op, P_CLEANUP, P_FINI);
+		M0_ASSERT(0);
 	}
+	case P_CAPTURE:
+		/**
+		 * TBD: uncomment function call below once node_free changes
+		 * done.
+		 */
+		/* btree_tx_node_capture(oi, bop->bo_tx); */
```
shashank-parulekar, line 5376: Please rename this routine to btree_tx_capture() since there might be more than one nodes for this btree which would be captured.

NikhilBirgade (approved):

btree/btree.c
```diff
@@ -5360,29 +5366,38 @@ static int64_t btree_del_kv_tick(struct m0_sm_op *smop)
...
-			oi->i_used --;
-			node = lev->l_node;
-			lev->l_node = NULL;
+		int i;
+		for (i = oi->i_used; i >= 0; i--) {
+			lev = &oi->i_level[i];
+			if (lev->l_freenode) {
+				M0_ASSERT(oi->i_used > 0);
+				oi->i_nop.no_opc = NOP_FREE;
+				node_free(&oi->i_nop, lev->l_node,
```
shashank-parulekar, line 5385: Where is the call to node_fini() as we discussed. It needs to be done before we reach this point.

shashank-parulekar (changes requested):

btree/btree.c
RadhaGulhane13: okay.

btree/btree.c
RadhaGulhane13: I will be covering it different PR as it may require some more changes.

shashank-parulekar (approved):

PR #880, opened 2021-07-23
Eos 22550 modified : changes for capturing transactions for KV Delete operation.

RadhaGulhane13:
This is the final PR. 
It includes following changes :
- calling node_fini separately than node_free.
- calling node_fini before P_CAPTURE and then calling P_FREENODE so that issue of freeing node_descriptor bfore node_capture will be resolved.
- Added node_lock at necessary places whenever we call node_isvalid()

### DIFF
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -1458,12 +1458,16 @@ static bool node_verify(const struct nd *node)
 #ifndef __KERNEL__
 /**
  * This function should get called in node_lock or tree_lock mode. As, it deals
- * with header update which will get modified only in tree_lock and node_lock
+ * with header update and n_be_node_valid flag check which will get modified
+ * only node_fini() in node lock mode or node_free() in tree_lock and node_lock
  * mode.
  */
 static bool node_isvalid(const struct nd *node)
 {
-	return node->n_type->nt_isvalid(&node->n_addr);
+	if (node->n_be_node_valid)
+		return node->n_type->nt_isvalid(&node->n_addr);
+
+	return false;
 }
 
 #endif
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -1456,9 +1456,8 @@ static bool node_verify(const struct nd *node)
 #ifndef __KERNEL__
 /**
  * This function should get called in node_lock or tree_lock mode. As, it deals
- * with header update and n_be_node_valid flag check which will get modified
- * only node_fini() in node lock mode or node_free() in tree_lock and node_lock
- * mode.
+ * with header and n_be_node_valid flag which will get updated in node_fini()
+ * and node_free() respectively in tree_lock or node_lock mode only.
  */
 static bool node_isvalid(const struct nd *node)
 {
@@ -3239,6 +3238,13 @@ static void btree_tx_commit_cb(void *payload)
 	node_lock(node);
 	M0_ASSERT(node->n_txref != 0);
 	node->n_txref--;
+	if (!node->n_be_node_valid && node->n_txref == 0) {
+		ndlist_tlink_del_fini(node);
+		node_unlock(node);
+		m0_rwlock_fini(&node->n_lock);
+		m0_free(node);
+		return;
+	}
 	node_unlock(node);
 }
 
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -7652,6 +7652,7 @@ static void ut_put_del_operation(void)
 					      .ksize = 8,
 					      .vsize = 8, };
 	struct m0_be_tx        *tx          = NULL;
+	struct m0_be_seg       *seg         = NULL;
 	struct m0_btree_op      b_op        = {};
 	struct m0_btree        *tree;
 	void                   *temp_node;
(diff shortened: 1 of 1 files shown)


shashank-parulekar (approved):

btree/btree.c
```diff
@@ -1449,9 +1454,18 @@ static bool node_verify(const struct nd *node)
 #endif
 
 #ifndef __KERNEL__
+/**
+ * This function should get called in node_lock or tree_lock mode. As, it deals
+ * with header update and n_be_node_valid flag check which will get modified
+ * only node_fini() in node lock mode or node_free() in tree_lock and node_lock
+ * mode.
```
NikhilBirgade, line 1461: Please update the comment.

btree/btree.c
```diff
@@ -2082,16 +2101,11 @@ static void node_put(struct node_op *op, struct nd *node, struct m0_be_tx *tx)
...
 			ndlist_tlink_del_fini(node);
 			m0_rwlock_fini(&node->n_lock);
-			op->no_addr = node->n_addr;
-			shift = node->n_type->nt_shift(node);
-			node->n_type->nt_fini(node, tx);
 			m0_free(node);
 			m0_rwlock_write_unlock(&list_lock);
-			m0_free_aligned(segaddr_addr(&op->no_addr),
-					1ULL << shift, shift);
 			/** Capture in transaction */
```
NikhilBirgade, line 2109: This comment can be removed as we are not dealing with be segment node here now.

btree/btree.c
```diff
@@ -2112,21 +2126,22 @@ static int64_t node_free(struct node_op *op, struct nd *node,
...
 		ndlist_tlink_del_fini(node);
+		node_unlock(node);
 		m0_rwlock_fini(&node->n_lock);
-		op->no_addr = node->n_addr;
-		node->n_type->nt_fini(node, tx);
 		m0_free(node);
 		m0_rwlock_write_unlock(&list_lock);
-		m0_free_aligned(segaddr_addr(&op->no_addr), 1ULL << shift,
-				shift);
 		/** Capture in transaction */
```
NikhilBirgade, line 2141: The comment can be removed from here. We should capture in transaction at line 2133.

btree/btree.c
```diff
@@ -2112,21 +2126,22 @@ static int64_t node_free(struct node_op *op, struct nd *node,
...
 
 	m0_rwlock_write_lock(&list_lock);
+	node_lock(node);
 	node->n_ref--;
 	node->n_be_node_valid = false;
+	op->no_addr = node->n_addr;
+	m0_free_aligned(segaddr_addr(&op->no_addr), 1ULL << shift, shift);
 
-	if (node->n_ref == 0) {
+	if (node->n_ref == 0 && node->n_txref == 0) {
```
NikhilBirgade, line 2135: What if the node_free and/or node_put gets called before transaction is committed and callback is called? In that case n_ref could reach zero but txref will still be non-zero in the node_free/node_put calls. Similar check should be added in callback to free up node descriptor in callback.

RadhaGulhane13: For any node which we want to free we are calling btree_tx_nodes_capture for that node and then node_free, so this scenario will not occur. If you see we are calling P_CAPTURE before P_FREENODE. so n_ref will not become zero.

NikhilBirgade: In btree_tx_nodes_capture we are just setting the callback btree_tx_commit_cb(). This will be invoked on tx_done, which we do not know when will it happen.

shashank-parulekar: @RadhaGulhane13 this needs to be taken care of in the capture callback since we cannot free the descriptor now because even if the n_ref is zero the n_txref is still non-zero. After sometime when the Tx commit completes and the Tx callback gets called then the node's n_txref will be decremented and become zero. At that point we need to free the node descriptor.
@NikhilBirgade  this might need to be done in the node_tx_callback routine !!!

RadhaGulhane13: So, we have node->n_txref == 0 check as well, and it is getting decremented on in btree_tx_commit_cb()

btree/btree.c
RadhaGulhane13: Updated Please check

btree/btree.c
RadhaGulhane13: made changes according to discussion.

btree/btree.c
```diff
@@ -3239,6 +3238,13 @@ static void btree_tx_commit_cb(void *payload)
 	node_lock(node);
 	M0_ASSERT(node->n_txref != 0);
 	node->n_txref--;
+	if (!node->n_be_node_valid && node->n_txref == 0) {
```
shashank-parulekar, line 3241: You also need to check if (node->n_ref == 0)

shashank-parulekar (changes requested):

btree/btree.c

btree/btree.c
```diff
@@ -1455,9 +1455,7 @@ static bool node_verify(const struct nd *node)
 
 #ifndef __KERNEL__
 /**
- * This function should get called in node_lock or tree_lock mode. As, it deals
- * with header and n_be_node_valid flag which will get updated in node_fini()
- * and node_free() respectively in tree_lock or node_lock mode only.
+ * This function should get called in node_lock mode only.
```
shashank-parulekar, line 1458: Replace the comment with:  This function should be called after acquiring node lock.

shashank-parulekar (approved):

NikhilBirgade (approved):

ydb242 (approved):

PR #882, opened 2021-07-23
Fix trees_loaded variable's memory corruption.

NikhilBirgade:

### DIFF
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -2241,7 +2241,7 @@ static int64_t mem_tree_get(struct node_op *op, struct segaddr *addr, int nxt)
 			M0_ASSERT(offset != 0);
 			offset--;
 			trees_in_use[i] |= (1ULL << offset);
-			offset += (i * sizeof trees_in_use[0]);
+			offset += (i * (sizeof trees_in_use[0]) * 8);
 			tree = &trees[offset];
 			trees_loaded++;
 			break;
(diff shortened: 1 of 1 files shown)


btree/btree.c
```diff
@@ -2332,8 +2332,8 @@ static void mem_tree_put(struct td *tree)
 		m0_rwlock_write_lock(&trees_lock);
 		M0_ASSERT(trees_loaded > 0);
 		i = tree - &trees[0];
-		array_offset = i / sizeof(trees_in_use[0]);
-		bit_offset_in_array = i % sizeof(trees_in_use[0]);
+		array_offset = i / (sizeof(trees_in_use[0]) * 8);
+		bit_offset_in_array = i % (sizeof(trees_in_use[0]) * 8);
```
RadhaGulhane13, line 2336: 8 is hardcoded can we replace it by defining cost variable and use it here

NikhilBirgade: Multiplying by 8 to convert it to bits. Anyways, this code will be replaced after tree_get/put changes.

shashank-parulekar, line 2336: Good catch. This same change needs to be done in tree_get routine where the offset it calculated at line 2244 in this file.

RadhaGulhane13: okay

shashank-parulekar (changes requested):

RadhaGulhane13 (approved):

shashank-parulekar (approved):

btree/btree.c
NikhilBirgade: Oh yes. Thanks.

PR #886, opened 2021-07-26
Reset phase of motr mini provisioner

atulsdeshmukh2312:


scripts/install/opt/seagate/cortx/motr/bin/motr_mini_prov.py
```diff
@@ -736,3 +731,32 @@ def lnet_self_ping(self):
        if ret != 0:
             return False
     return True
+
+def lvm_clean(self):
+    vol_grps=execute_command(self, "vgs|grep vg_srvnode|awk '{print $1}'")[0].split('\n')[0:-1]
```
madhavemuri, line 736: For the case of stale LVM's after re-image and LVM creation is failing, we have a method of wipefs and dmsetup logic, can you include it as well ?

PR #889, opened 2021-07-26
[DTM0] EOS-22352: PoC of Versioned ordinary catalogues.

ivan-alekhin:
The patch introduces tombstones and versions into CAS.
Tombstones and versions may be used to enable the
"conflict-free" behavior: a set of key-value pairs
can always be merged with another set (for example,
a set of REDO records sent from another process).

Notes:
This PR is a continuation of https://github.com/Seagate/cortx-motr/pull/662. Initially, versions had been added into btree but Nikita recommended to put them into CAS.
The PR is not ready for a formal review -- DTM team needs to do internal review first.

### DIFF
--- a/cas/cas.h
+++ b/cas/cas.h
@@ -136,6 +136,21 @@ struct m0_cas_kv_vec {
 } M0_XCA_SEQUENCE M0_XCA_DOMAIN(rpc);
 
 /**
+ * Version of a CAS record.
+ * A version comprises a physical timestamp and a tombstone flag.
+ * The timestamp is set on the client side whenever the corresponding
+ * record was supposed to be modified by PUT or DEL request.
+ * The tombstone flag is set when the corresponding record has been
+ * "logicaly" removed from the storage (although, it still exists
+ * there as a "dead" record).
+ * See ::COF_VERSIONED for details.
+ */
+struct m0_cas_kv_ver {
+	struct m0_dtm0_ts ckv_ts;
+	bool              ckv_tombstone;
+} M0_XCA_RECORD M0_XCA_DOMAIN(rpc);
+
+/**
  * CAS index record.
  */
 struct m0_cas_rec {
@@ -196,6 +211,13 @@ struct m0_cas_rec {
 	 * records.
 	 */
 	uint64_t             cr_rc;
+
+	/**
+	 * Optional version of this record.
+	 * The version is returned as a reply to GET and NEXT requests
+	 * when COF_VERSIONED is specified in the request.
+	 */
+	struct m0_cas_kv_ver cr_ver;
 } M0_XCA_RECORD M0_XCA_DOMAIN(rpc);
 
 /**
@@ -290,6 +312,19 @@ enum m0_cas_op_flags {
 	 * use the flag but they do not depend on enabled DTM0).
 	 */
 	COF_VERSIONED = 1 << 8,
+
+	/**
+	 * Makes NEXT return "dead" records (with tombstones) when
+	 * version-aware behavior is specified (see ::COF_VERSIONED).
+	 * By default, COF_VERSIONED does not return dead records for NEXT
+	 * requests but when both flags are specified (VERSIONED | SHOW_DEAD)
+	 * then it yields all records whether they are alive or not.
+	 * It might be used by the client when it wants to analyse the records
+	 * received from several catalogue services, so that it could properly
+	 * merge the results of NEXT operations eliminating inconsistencies.
+	 * In other words, it might be used in degraded mode.
+	 */
+	COF_SHOW_DEAD = 1 << 9,
 };
 
 enum m0_cas_opcode {
--- a/cas/cas.h
+++ b/cas/cas.h
@@ -136,19 +136,26 @@ struct m0_cas_kv_vec {
 } M0_XCA_SEQUENCE M0_XCA_DOMAIN(rpc);
 
 /**
- * Version of a CAS record.
+ * CAS record version and tombstone encoded in on-disk/on-wire format.
+ * Format:
+ *     MSB                             LSB
+ *     +----------------+----------------+
+ *     | 1 bit          | 63 bits        |
+ *     |<- tombstone -> | <- timestamp ->|
+ *     +----------------+----------------+
+ *
  * A version comprises a physical timestamp and a tombstone flag.
  * The timestamp is set on the client side whenever the corresponding
  * record was supposed to be modified by PUT or DEL request.
  * The tombstone flag is set when the corresponding record has been
  * "logicaly" removed from the storage (although, it still exists
  * there as a "dead" record).
- * See ::COF_VERSIONED for details.
+ *
+ * See ::COF_VERSIONED, ::M0_CRV_VER_NONE for details.
  */
-struct m0_cas_kv_ver {
-	struct m0_dtm0_ts ckv_ts;
-	bool              ckv_tombstone;
-} M0_XCA_RECORD M0_XCA_DOMAIN(rpc);
+struct m0_crv {
+	uint64_t crv_encoded;
+} M0_XCA_RECORD M0_XCA_DOMAIN(rpc|be);
 
 /**
  * CAS index record.
@@ -217,7 +224,7 @@ struct m0_cas_rec {
 	 * The version is returned as a reply to GET and NEXT requests
 	 * when COF_VERSIONED is specified in the request.
 	 */
-	struct m0_cas_kv_ver cr_ver;
+	struct m0_crv cr_ver;
 } M0_XCA_RECORD M0_XCA_DOMAIN(rpc);
 
 /**
@@ -463,6 +470,48 @@ M0_INTERNAL void m0_cas_id_fini(struct m0_cas_id *cid);
 M0_INTERNAL bool m0_cas_id_invariant(const struct m0_cas_id *cid);
 
 M0_INTERNAL bool cas_in_ut(void);
+
+enum {
+	/* Tombstone flag: marks a dead kv pair. We use the MSB here. */
+	M0_CRV_TBS = 1L << (sizeof(uint64_t) * CHAR_BIT - 1),
+	/*
+	 * A special value for the empty version.
+	 * A record with the empty version is always overwritten by any
+	 * PUT or DEL operation that has a valid non-empty version.
+	 * A PUT or DEL operation with the empty version always ignores
+	 * the version-aware behavior: records are actually removed by DEL,
+	 * and overwritten by PUT, no matter what was stored in the catalogue.
+	 */
+	M0_CRV_VER_NONE = 0,
+	/* The maximum possible value of a version. */
+	M0_CRV_VER_MAX = (UINT64_MAX & ~M0_CRV_TBS) - 1,
+	/* The minimum possible value of a version. */
+	M0_CRV_VER_MIN = M0_CRV_VER_NONE + 1,
+};
+
+/*
+ * 100:a == alive record with version 100
+ * 123:d == dead record with version 123
+ */
+#define CRV_F "%" PRIu64 ":%c"
+#define CRV_P(__crv) m0_crv_ts(__crv).dts_phys, m0_crv_tbs(__crv) ? 'd' : 'a'
+
+#define M0_CRV_INIT_NONE ((struct m0_crv) { .crv_encoded = M0_CRV_VER_NONE })
+
+M0_INTERNAL void m0_crv_init(struct m0_crv           *crv,
+			     const struct m0_dtm0_ts *ts,
+			     bool                     tbs);
+
+M0_INTERNAL bool m0_crv_is_none(const struct m0_crv *crv);
+M0_INTERNAL int m0_crv_cmp(const struct m0_crv *left,
+			   const struct m0_crv *right);
+
+M0_INTERNAL bool m0_crv_tbs(const struct m0_crv *crv);
+M0_INTERNAL void m0_crv_tbs_set(struct m0_crv *crv, bool tbs);
+
+M0_INTERNAL struct m0_dtm0_ts m0_crv_ts(const struct m0_crv *crv);
+M0_INTERNAL void m0_crv_ts_set(struct m0_crv           *crv,
+			       const struct m0_dtm0_ts *ts);
 
 /** @} end of cas_dfspec */
 #endif /* __MOTR_CAS_CAS_H__ */
--- a/cas/cas.h
+++ b/cas/cas.h
@@ -251,7 +251,45 @@ enum m0_cas_op_flags {
 	 * For PUT/DEL operation, instructs it to delay reply from CAS service
 	 * until BE transaction is persisted.
 	 */
-	COF_SYNC_WAIT = 1 << 7
+	COF_SYNC_WAIT = 1 << 7,
+
+	/**
+	 * Enables version-aware behavior for PUT, DEL, GET, and NEXT requests.
+	 *
+	 * Overview
+	 * --------
+	 *   Versions are taken from m0_cas_op::cg_txd. When this flag is set,
+	 * the following change happens in the logic of the mentioned request
+	 * types:
+	 *     - PUT does not overwrite "newest" (version-wise) records.
+	 *       Requirements:
+	 *         x COF_OVERWRITE is set.
+	 *         x Transaction descriptor has a valid DTX ID.
+	 *     - DEL puts a tombstone instead of an actual removal. In this
+	 *       mode, DEL does not return -ENOENT in the same way as
+	 *       PUT-with-COF_OVERWRITE does not return -EEXIST.
+	 *       Requirements:
+	 *         x Transaction descriptor has a valid DTX ID.
+	 *     - GET returns only "alive" entries (without tombstones).
+	 *     - NEXT also skips entries with tombstones.
+	 *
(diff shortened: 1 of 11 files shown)


cas/cas.h
```diff
@@ -135,6 +135,21 @@ struct m0_cas_kv_vec {
...
+ * The timestamp is set on the client side whenever the corresponding
+ * record was supposed to be modified by PUT or DEL request.
+ * The tombstone flag is set when the corresponding record has been
+ * "logicaly" removed from the storage (although, it still exists
+ * there as a "dead" record).
+ * See ::COF_VERSIONED for details.
+ */
+struct m0_cas_kv_ver {
+	struct m0_dtm0_ts ckv_ts;
+	bool              ckv_tombstone;
```
just-now, line 150: a very strange question...
do you think we should avoid using `bool` types in network messages for any reason?

cas/client.h
```diff
@@ -475,11 +479,11 @@ M0_INTERNAL int m0_cas_get(struct m0_cas_req      *req,
...
  * A version-aware version of m0_cas_get.
- * It returns the value associated with the given key only this value
- * does not have a tombstone set (i.e., it was not removed).
- * Ssee ::COF_VERSIONED.
+ * It returns the values associated with the given keys only if this values
+ * do not have tombstones set (i.e., they have not been removed).
+ * See ::COF_VERSIONED.
  *
- * TODO: This function will be disolved once ::m0_cas_get gets an extra
+ * TODO: This function may be disolved if ::m0_cas_get gets an extra
```
just-now, line 486: RE: m0_cas_versioned_get()

maybe having just m0_cas_get(req) with preceding req->flags |= VERSIONED will be sufficient?

just-now (commented):
looks good.

cas/cas.h
ivan-alekhin: That's a good question. Generally speaking, we should not use non-fixed-size types for on-disk/on-wire data types. However, in our case, the platform is fixed, so xcode should always generate the same u8 parsers for this field.

Nevertheless, I believe the idea of exposing versions/tombstones using a separate type did not serve its purpose, so I just removed this structure and now we use m0_crv everywhere (see https://github.com/Seagate/cortx-motr/pull/889/commits/262f68da0c955caa0674a60222fd947e52571724)

cas/client.h
ivan-alekhin: Nope. m0_cas_req has no flags. Unfortunately, there is no field to set before or after m0_cas_get call.

cas/cas.h
```diff
@@ -135,6 +135,28 @@ struct m0_cas_kv_vec {
...
+ * A version comprises a physical timestamp and a tombstone flag.
+ * The timestamp is set on the client side whenever the corresponding
+ * record was supposed to be modified by PUT or DEL request.
+ * The tombstone flag is set when the corresponding record has been
+ * "logicaly" removed from the storage (although, it still exists
+ * there as a "dead" record).
+ *
+ * See ::COF_VERSIONED, ::M0_CRV_VER_NONE for details.
+ */
+struct m0_crv {
```
madhavemuri, line 156: Why the structure type is used here ?

ivan-alekhin: Do you mean "why don't we just use uint64_t instead?". I asked myself the same question. There are two reasons:

1. It might be error-prone. The version (the pair "tombstone+timestamp") of a key-value record is a data type. The domain of this type is ([0, 1], [1..2^63]) | (0, 0). The domain of uint64_t is just 0..UINT64_MAX. A separate data type should make the reader of this code to think about that. For example, we have uint128 and fid data types for different purposes even though they have the same size.

2. There is a chance that we need to add more information here. For example, we might want to add the fid of the originator into this data type. I doubt we ever do that, but it is much easier to extend/modify a structure than a bare uint64_t.

cas/client.c
```diff
@@ -1765,8 +1783,12 @@ M0_INTERNAL int m0_cas_next(struct m0_cas_req *req,
 	M0_PRE(start_keys != NULL);
 	M0_PRE(m0_cas_req_is_locked(req));
 	M0_PRE(m0_cas_id_invariant(index));
-	/* Only slant and exclude start key flags are allowed. */
-	M0_PRE((flags & ~(COF_SLANT | COF_EXCLUDE_START_KEY)) == 0);
+	/* Only slant, exclude start key, and versioned flags are allowed. */
```
huanghua78, line 1786: This flag is from user.
It's better not to panic, but return -EINVAL.

cas/ctg_store.c
```diff
@@ -160,82 +213,196 @@ static int ctg_berc(struct m0_ctg_op *ctg_op)
...
 		return M0_ERR(-ENOMEM);
 }
 
 /**
- * Allocate memory and unpack value having format length + data.
- *
- * @param val destination buffer
- * @param buf source buffer
- * @return 0 if ok else error code
+ * Unpack an an on-disk value data into in-memory format.
```
huanghua78, line 267: remove extra 'an'

cas/ctg_store.c
```diff
@@ -160,82 +213,196 @@ static int ctg_berc(struct m0_ctg_op *ctg_op)
...
+ * The function makes "buf" to point to the user-specific data associated
+ * with the value (see ::generic_value::gv_data).
+ * @param[out] Optional storage for the version of the record.
```
huanghua78, line 270: add "crv"

cas/ctg_store.c
```diff
@@ -160,82 +213,196 @@ static int ctg_berc(struct m0_ctg_op *ctg_op)
...
+		return M0_ERR_INFO(-EPROTO, "%" PRIu64 " < %" PRIu64,
+				   buf->b_nob, sizeof(*value));
+	if (value->gv_length != buf->b_nob - sizeof(*value))
+		return M0_ERR(-EPROTO);
+
+	if (crv != NULL)
+		*crv = value->gv_version;
+
+	buf->b_nob = value->gv_length;
+	buf->b_addr = &value->gv_data[0];
```
huanghua78, line 292: Any problems if we call m0_buf_free() later?

huanghua78: The data is in BE segment? So we can not call m0_free_buf()?

cas/ctg_store.c
```diff
@@ -160,82 +213,196 @@ static int ctg_berc(struct m0_ctg_op *ctg_op)
...
+	return M0_RC(0);
+}
+
+/**
+ * Get a m0_cas_ctg pointer from in-memory representation of a meta-index
+ * value.
+ */
+static int ctg_vbuf_as_ctg(const struct m0_buf *buf, struct m0_cas_ctg **ctg)
+{
+	struct meta_value *mv = buf->b_addr - sizeof(struct generic_value);
```
huanghua78, line 303: Should be?
struct meta_value *mv = buf->b_addr + sizeof(struct generic_value);

huanghua78, line 303: This is frightening, because this is to access some memory prior to the "buf->b_addr".

cas/ctg_store.c
```diff
@@ -160,82 +213,196 @@ static int ctg_berc(struct m0_ctg_op *ctg_op)
...
+
+	M0_ENTRY();
+
+	if (buf->b_nob == sizeof(struct m0_cas_ctg *)) {
+		*ctg = mv->mv_ctg;
```
huanghua78, line 308: Why don't see just use:
*ctg = (struct m0_cas_ctg *)buf->b_addr;

cas/ctg_store.c
```diff
@@ -160,82 +213,196 @@ static int ctg_berc(struct m0_ctg_op *ctg_op)
...
 
-static void ctg_fid_key_fill(void *key, const struct m0_fid *fid)
+/**
+ * Get the layout from an unpacked ::layout_value (cctidx value).
+ */
+static int ctg_vbuf_as_layout(const struct m0_buf    *buf,
+			      struct m0_dix_layout **layout)
 {
-	ctg_memcpy(key, fid, sizeof(struct m0_fid));
+	struct layout_value *lv = buf->b_addr - sizeof(struct generic_value);
```
huanghua78, line 320: The same as previous one.

cas/cas.h
```diff
@@ -135,6 +135,28 @@ struct m0_cas_kv_vec {
...
+ *     MSB                             LSB
+ *     +----------------+----------------+
+ *     | 1 bit          | 63 bits        |
+ *     |<- tombstone -> | <- timestamp ->|
+ *     +----------------+----------------+
+ *
+ * A version comprises a physical timestamp and a tombstone flag.
+ * The timestamp is set on the client side whenever the corresponding
+ * record was supposed to be modified by PUT or DEL request.
+ * The tombstone flag is set when the corresponding record has been
```
huanghua78, line 150: This flag is set by the CAS server internally, not the user, right?

cas/ctg_store.c
```diff
@@ -1883,6 +2080,289 @@ static const struct m0_be_btree_kv_ops cas_btree_ops = {
...
+
+	anchor->ba_value.b_nob = ctg_vbuf_packed_size(&ctg_op->co_val);
+	rc = M0_BE_OP_SYNC_RET(op,
+			       m0_be_btree_save_inplace(btree, tx, &op,
+							key, anchor, true,
+							ctg_op_zones(ctg_op)),
+			       bo_u.u_btree.t_rc);
+
+	if (rc == 0) {
+		ctg_vbuf_pack(&ctg_op->co_anchor.ba_value,
```
huanghua78, line 2235: Can we:
s/&ctg_op->co_anchor.ba_value/&anchor->ba_value
?

cas/ctg_store.c
```diff
@@ -1883,6 +2080,289 @@ static const struct m0_be_btree_kv_ops cas_btree_ops = {
...
+			       bo_u.u_btree.t_rc) ?:
+		ctg_vbuf_unpack(&anchor->ba_value, &ctg_op->co_out_ver);
+
+	if (rc == 0) {
+		if (m0_crv_tbs(&ctg_op->co_out_ver))
+			rc = -ENOENT;
+		else
+			ctg_op->co_out_val = anchor->ba_value;
+	}
+
```
huanghua78, line 2274: Do we need to call  m0_be_btree_release(NULL, anchor); here?

In versioned_put_sync(), the same operation is called.
Why is it not called here in this function?

cas/ctg_store.c
```diff
@@ -1181,39 +1385,29 @@ M0_INTERNAL int m0_ctg_dead_index_insert(struct m0_ctg_op  *ctg_op,
...
+	return ctg_exec(ctg_op, ctg, &M0_BUF_INIT_PTR(&ctg), next_phase);
 }
 
 static int ctg_exec(struct m0_ctg_op    *ctg_op,
 		    struct m0_cas_ctg   *ctg,
 		    const struct m0_buf *key,
 		    int                  next_phase)
 {
-	int ret = M0_FSO_AGAIN;
+	int               ret     = M0_FSO_AGAIN;
```
siningwuseagate, line 1402: remove extra space

cas/ctg_store.c
```diff
@@ -1883,6 +2080,289 @@ static const struct m0_be_btree_kv_ops cas_btree_ops = {
...
+		ctg_vbuf_pack(&ctg_op->co_anchor.ba_value,
+			      &ctg_op->co_val,
+			      &new_version);
+
+		m0_ctg_state_inc_update(tx,
+					key->b_nob -
+					sizeof(struct generic_key) +
+					ctg_op->co_val.b_nob);
+	}
+
```
huanghua78, line 2244: Do we need to call m0_be_btree_release(NULL, anchor); here?

cas/ctg_store.c
```diff
@@ -1612,6 +1820,15 @@ M0_INTERNAL void m0_ctg_delete_credit(struct m0_cas_ctg      *ctg,
...
 	m0_be_btree_delete_credit(&ctg->cc_tree, 1, knob, vnob, accum);
+
+	/* XXX: performance.
+	 * At this moment we do not know for sure if the CAS operation should
+	 * have version-aware behavior. So that we are doing a pessimistic
+	 * estimate here: reserving credits for both delete and insert.
+	 */
+	if (ctg_is_ordinary(ctg)) {
+		m0_be_btree_insert_credit2(&ctg->cc_tree, 1, knob, vnob, accum);
+	}
```
siningwuseagate, line 1831: the { and } are not needed

cas/ctg_store.c
```diff
@@ -160,82 +213,196 @@ static int ctg_berc(struct m0_ctg_op *ctg_op)
...
+	if (buf->b_nob == sizeof(struct m0_dix_layout)) {
+		*layout = &lv->lv_layout;
+		return M0_RC(0);
+	} else
+		return M0_ERR(-EPROTO);
 }
 
-static m0_bcount_t ctg_ksize(const void *key)
+/**
+ * Convert a versioned variable length buffer (on-disk data) into user-specific
```
siningwuseagate, line 332: s/Convert/Converts

cas/ctg_store.c
```diff
@@ -1883,6 +2080,289 @@ static const struct m0_be_btree_kv_ops cas_btree_ops = {
...
+		if (has_txd)
+			return M0_RC(true);
+		else
+			return M0_RC_INFO(false,
+					  "%s request is has an empty txd.",
+					  ctg_op->co_opcode == CO_PUT ?
+					  "PUT" : "DEL");
+	case CTG_OP_COMBINE(CO_GET, CT_BTREE):
+	case CTG_OP_COMBINE(CO_CUR, CT_BTREE):
+		return M0_RC(true);
```
huanghua78, line 2132: All CAS_GET and CAS_NEXT are forced to be versioned operations?

cas/ctg_store.c
ivan-alekhin: Yes, this data lives in BE segment -- we cannot call m0_buf_free on it.

cas/ctg_store.c
ivan-alekhin: Yes. I think an M0_AMB or container_of would look less scary. I mean something like that:

```c
struct generic_value *gv = M0_AMB(gv, (uint8_t **) buf->b_addr, gv_data);
struct meta_value    *mv = M0_AMB(mv, gv, mv_gval);
```

cas/ctg_store.c
ivan-alekhin: You already have an answer: if we do `*ctg = (struct m0_cas_ctg *)buf->b_addr;` then we are copying the wrong data. b_addr is a double pointer (pointer to pointer to catalogue). It should be something like that:

```c
*ctg = *((struct m0_cas_ctg **) buf->b_addr);
```

Playing with double pointers could be dangerous, so I decided to protect myself and the reader from making such mistakes. I'm using the power of data types to copy the pointer in a relatively safe way. What do you think about that?

cas/cas.h
ivan-alekhin: Yes. The catalogue store marks a record as dead if the user wants to delete the record.

cas/ctg_store.c
ivan-alekhin: Yes, makes sense.

cas/ctg_store.c
ivan-alekhin: versioned_put_sync reuses the same anchor:
```
// PUT:
anchor = ctg_op->op_anchor;
lookup_inplace(..., anchor, ...);
release(anchor);
save_inplace(..., anchor, ...);
// release(anchor); will be called by m0_ctg_op_fini
```
Here we do not reuse it:
```
// GET:
save_inplace(..., anchor, ...);
// release(anchor); will be called by m0_ctg_op_fini
```

cas/ctg_store.c
ivan-alekhin: It is done by `m0_ctg_op_fini`.

cas/ctg_store.c
ivan-alekhin: Not all. Only GET/NEXT that have COF_VERSIONED flag set. Please take a look at the previous conditions.

huanghua78: Yes. OK.

cas/ctg_store.c
huanghua78: OK

cas/ctg_store.c
huanghua78: Removed all extra spaces/tabs. Done.

cas/ctg_store.c
huanghua78: OK. I see.

cas/ctg_store.c
huanghua78: Yes, agree.
When I did the ctg dump task, I found it is confusing here. 
So, when I read this patch, it is much easier to understand.

cas/ctg_store.c
huanghua78: OK

cas/ctg_store.c
huanghua78: OK

cas/client.c
huanghua78: OK

madhavemuri:
@ivan-alekhin : Is this PR still needed after PR-1515 is merged ?

ivan-alekhin:
Closing the PR because the patch was merged within https://github.com/Seagate/cortx-motr/pull/1515

PR #890, opened 2021-07-27
Ut changes group size and indexsize tunable.

nihar1313:


nihar1313:
This is no longer valid, new PR has been raised

PR #894, opened 2021-07-29
Eos 16852 : Support OVERWRITE record value in new bTree when new value size is same as value size of existing record.

RadhaGulhane13:


btree/btree.c
```diff
@@ -6808,6 +6902,43 @@ static void btree_ut_kv_oper_thread_handler(struct btree_ut_thread_info *ti)
...
+			 *  in a loaded system.
+			 */
+			key[0] = (key_first << (sizeof(ti->ti_thread_id) * 8)) +
+				 ti->ti_thread_id;
+			key[0] = m0_byteorder_cpu_to_be64(key[0]);
+			for (i = 1; i < ARRAY_SIZE(key); i++)
+				key[i] = key[0];
+
+			value[0] = key[0];
+			for (i = 1; i < ARRAY_SIZE(value); i++)
```
RadhaGulhane13, line 6926: keeping value same as key, as btree_kv_get_cb will fail if value is different than key

btree/btree.c
```diff
@@ -3889,23 +3903,40 @@ static int64_t btree_put_kv_tick(struct m0_sm_op *smop)
...
+		int                  rc;
+
+		if (bop->bo_opc == M0_BO_PUT && oi->i_key_found)
+			rec.r_flags = M0_BSC_KEY_EXISTS;
+		else if (bop->bo_opc == M0_BO_UPDATE && !oi->i_key_found)
+			rec.r_flags = M0_BSC_KEY_NOT_FOUND;
+		else
+			rec.r_flags = M0_BSC_SUCCESS;
+
+		if (rec.r_flags) {
```
shashank-parulekar, line 3920: s/rec.r_flags/rec.r_flags != M0_BSC_SUCCESS/

btree/btree.c
```diff
@@ -3865,17 +3875,21 @@ static int64_t btree_put_kv_tick(struct m0_sm_op *smop)
 			}
 		}
 		/** Fall through if path_check is successful. */
+	case P_DECIDE_NXT: {
```
shashank-parulekar, line 3878: s/P_DECIDE_NXT/P_SANITY_CHECK/

btree/btree.c
```diff
@@ -3865,17 +3875,21 @@ static int64_t btree_put_kv_tick(struct m0_sm_op *smop)
...
+		if (bop->bo_opc == M0_BO_PUT) {
+			if (oi->i_key_found)
+				return P_ACT;
+			else
+				return  P_MAKESPACE;
+		} else {
+			M0_ASSERT(bop->bo_opc == M0_BO_UPDATE);
+			return P_ACT;
+		}
```
shashank-parulekar: Remove these lines and move the code at lines 3917 - 3929 here with appropriate changes.

shashank-parulekar (changes requested):

btree/btree.c

shashank-parulekar (changes requested):

btree/btree.c
```diff
@@ -3875,16 +3875,11 @@ static int64_t btree_put_kv_tick(struct m0_sm_op *smop)
...
-			if (oi->i_key_found)
-				return P_ACT;
-			else
-				return  P_MAKESPACE;
-		} else {
-			M0_ASSERT(bop->bo_opc == M0_BO_UPDATE);
+	case P_SANITY_CHECK: {
+		if (bop->bo_opc == M0_BO_PUT && !oi->i_key_found)
+			return  P_MAKESPACE;
+		else
```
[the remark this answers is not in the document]
RadhaGulhane13: It is necessary to have checks in P_ACT, as in  P_SANITY_CHECK we can not call fail routine directly as callback needs to be done first. and callback will be called in P_ACT.

btree/btree.c
```diff
@@ -3736,6 +3737,15 @@ static int64_t btree_put_kv_tick(struct m0_sm_op *smop)
 						&child_node_addr, P_NEXTDOWN);
 			} else {
 				node_unlock(lev->l_node);
+				if (bop->bo_opc == M0_BO_UPDATE) {
+					/**
+					 * TBD : Revisit  this part, while
+					 * implmenting UPDATE operatoion for
```
NikhilBirgade, line 3743: operation.

btree/btree.c
```diff
@@ -3917,14 +3939,17 @@ static int64_t btree_put_kv_tick(struct m0_sm_op *smop)
...
-		int rc = bop->bo_cb.c_act(&bop->bo_cb, rec);
+		node_slot.s_rec.r_flags = M0_BSC_SUCCESS;
+		rc = bop->bo_cb.c_act(&bop->bo_cb, &node_slot.s_rec);
 		if (rc) {
 			/* handle if callback fail i.e undo make */
-			node_del(node_slot.s_node, node_slot.s_idx, bop->bo_tx);
-			node_done(&node_slot, bop->bo_tx, true);
-			node_seq_cnt_update(lev->l_node);
-			node_fix(lev->l_node, bop->bo_tx);
+			if (bop->bo_opc == M0_BO_PUT) {
```
NikhilBirgade, line 3946: `if ( rc && bop->bo_opc == M0_BO_PUT)`

RadhaGulhane13: line 3946 will get execute only if rec is non zero. no need to check for rc flag again

btree/btree.c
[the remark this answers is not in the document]
shashank-parulekar: If the check fails and we can return the proper error code then I am thinking we do not need the callback. I understand we can provide an error status in the callback but maybe we can just skip this part - I think it will be more useful for the caller to debug his code as he can check for the return value of the function instead of looking for the error in the callback. I hope these are not a lot of changes.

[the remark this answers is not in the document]
RadhaGulhane13: Yes we can do that , but currently ut is is checking callback flag which will get set in callback routine only, If we do above changes, we may need to change error handling in ut. 
Also, it can happen that user request to put the key which is already exists, in such case don't we need to do callback , which will tell user that key already exists.

btree/btree.c
NikhilBirgade: I meant the if condition could be clubbed with `if(rc)`.

RadhaGulhane13: I have not done that  as we need to execute line 3954-3956 if rc is non zero and  irrespective of bop->bo_opc code.
If we  (rc && bop->bo_opc == M0_BO_PUT)->execute error handling for PUT then again we need to check for rc, if it is non zero then we will execute line 3954-3956. I think it won't be readable.

btree/btree.c
```diff
@@ -3865,19 +3879,27 @@ static int64_t btree_put_kv_tick(struct m0_sm_op *smop)
...
+		int  rc;
+		if (bop->bo_opc == M0_BO_PUT && oi->i_key_found)
+			rc = M0_BSC_KEY_EXISTS;
+		else if (bop->bo_opc == M0_BO_UPDATE && !oi->i_key_found)
+			rc = M0_BSC_KEY_NOT_FOUND;
+		else
+			rc = M0_BSC_SUCCESS;
+
+		if (rc != M0_BSC_SUCCESS) {
+			oi->i_sm_rc = rc;
```
NikhilBirgade, line 3892: I do not see the use of `i_sm_rc` variable. It is getting updated when `rc != M0_BSC_SUCCESS` and from here tick sm will exit via fail routine.

btree/btree.c
```diff
@@ -3865,19 +3879,27 @@ static int64_t btree_put_kv_tick(struct m0_sm_op *smop)
...
 			lock_op_unlock(tree);
-			return m0_sm_op_sub(&bop->bo_op, P_CLEANUP, P_FINI);
+			return fail(bop, M0_ERR(-rc));
```
NikhilBirgade, line 3894: `enum m0_btree_status_codes` return codes should not be used for returning tick machine return codes. We are using this enum as flag in record. State machine return codes should be linux codes.

btree/btree.c
RadhaGulhane13: Yes in P_FINI, we are checking if sm_rc is non-zero, if it is non-zero, we are returning ERROR

btree/btree.c
RadhaGulhane13: okay, I am using m0_btree_status_codes as return codes , as it will help debugging. but if it is necessary to s=use linux codes for return code I will update.

btree/btree.c
NikhilBirgade: In fail routine `bop->bo_op.o_sm.sm_rc = rc` will update the state machine return code, we don't need the error returning code in P_FINI.

RadhaGulhane13: bop->bo_op.o_sm.sm_rc  will get reinitialized to 0  in op.c: m0_sm_op_tick() after returning next state . That's why we can not use bop->bo_op.o_sm.sm_rc. I also thought the same but because of this issue, had to add one more flag in oimpl structure.

sm/op.c
```diff
@@ -152,8 +152,8 @@ bool m0_sm_op_tick(struct m0_sm_op *op)
 		}
 		wait = (result & M0_SMOP_WAIT) != 0;
 		M0_ASSERT(ergo(wait, ceo->oe_vec->eo_is_armed(ceo)));
-		if (fail)
-			m0_sm_state_and_rc_set(&op->o_sm, 
+		if (fail || op->o_sm.sm_rc!= 0)
```
NikhilBirgade: `fail` check can be removed from all places. `op->o_sm.sm_rc!= 0` check should be sufficient.

btree/btree.c
```diff
@@ -3865,19 +3875,26 @@ static int64_t btree_put_kv_tick(struct m0_sm_op *smop)
...
-			}
+	case P_SANITY_CHECK: {
+		int  rc;
+		if (bop->bo_opc == M0_BO_PUT && oi->i_key_found)
+			rc = M0_BSC_KEY_EXISTS;
+		else if (bop->bo_opc == M0_BO_UPDATE && !oi->i_key_found)
+			rc = M0_BSC_KEY_NOT_FOUND;
+		else
+			rc = M0_BSC_SUCCESS;
+
```
NikhilBirgade, line 3886: initialize `rc =0`.  
/s/M0_BSC_KEY_EXISTS/EEXIST
/s/M0_BSC_KEY_NOT_FOUND/ENOENT
remove `else rc = M0_BSC_SUCCESS; `
/s/EPERM/rc

sm/op.c
```diff
@@ -152,8 +152,8 @@ bool m0_sm_op_tick(struct m0_sm_op *op)
 		}
 		wait = (result & M0_SMOP_WAIT) != 0;
 		M0_ASSERT(ergo(wait, ceo->oe_vec->eo_is_armed(ceo)));
-		if (fail)
-			m0_sm_state_and_rc_set(&op->o_sm, 
+		if (fail || op->o_sm.sm_rc != 0)
```
NikhilBirgade, line 155: 'fail' check can be removed from all the places. `op->o_sm.sm_rc != 0` is sufficient.

NikhilBirgade (approved):

shashank-parulekar (approved):

PR #895, opened 2021-07-29
EOS-22075: Segregate tree_*() from ff_*() and mem_*()

ydb242:


btree/btree.c
```diff
@@ -1885,12 +1869,45 @@ static int64_t mem_tree_get(struct node_op *op, struct segaddr *addr, int nxt);
  */
 static int64_t tree_get(struct node_op *op, struct segaddr *addr, int nxt)
 {
-	int        nxt_state;
+	struct td              *tree = NULL;
+	struct nd              *node = NULL;
+
+	if (addr != NULL) {
+		nxt  = node_get(op, NULL, addr, nxt);
```
NikhilBirgade, line 1876: Need to handle error codes returned by node_get() .

btree/btree.c
```diff
@@ -1885,12 +1869,45 @@ static int64_t mem_tree_get(struct node_op *op, struct segaddr *addr, int nxt);
...
+			m0_rwlock_write_lock(&tree->t_lock);
+			tree->t_ref++;
+		}
+
+		op->no_node = tree->t_root;
+		op->no_tree = tree;
+		m0_rwlock_write_unlock(&tree->t_lock);
+
+	} else
+		op->no_op.o_sm.sm_rc = M0_ERR(-EINVAL);
```
NikhilBirgade, line 1906: Are we assuming that a that tree_create will always be called with a valid `addr` ?

btree/btree.c
```diff
@@ -1885,12 +1869,45 @@ static int64_t mem_tree_get(struct node_op *op, struct segaddr *addr, int nxt);
...
 static int64_t tree_get(struct node_op *op, struct segaddr *addr, int nxt)
 {
-	int        nxt_state;
+	struct td              *tree = NULL;
+	struct nd              *node = NULL;
+
+	if (addr != NULL) {
+		nxt  = node_get(op, NULL, addr, nxt);
+		node = op->no_node;
+		if (node != NULL) {
```
NikhilBirgade, line 1878: In which condition node can be NULL?

btree/btree.c
```diff
@@ -1908,10 +1925,28 @@ static int64_t tree_get(struct node_op *op, struct segaddr *addr, int nxt)
 static int64_t tree_create(struct node_op *op, struct m0_btree_type *tt,
 			   int rootshift, struct m0_be_tx *tx, int nxt)
 {
-	return segops->so_tree_create(op, tt, rootshift, tx, nxt);
+	struct td *tree;
```
NikhilBirgade, line 1928: This code is under `#if 0`. are we not going to need this code?

btree/btree.c
```diff
@@ -1926,9 +1961,15 @@ static int64_t tree_delete(struct node_op *op, struct td *tree,
 			   struct m0_be_tx *tx, int nxt)
 {
 	M0_PRE(tree != NULL);
-	return segops->so_tree_delete(op, tree, tx, nxt);
+	struct nd *root = tree->t_root;
+
+	op->no_tree = tree;
+	op->no_node = root;
+	node_free(op, op->no_node, tx, nxt);
+	tree_put(tree);
```
NikhilBirgade, line 1969: In line 1968, you have freed the root node and then taking lock for root in tree_put(), it will result in error.

btree/btree.c
```diff
@@ -1976,9 +2023,9 @@ static int64_t node_get(struct node_op *op, struct td *tree,
 	bool                    in_lrulist;
 	uint32_t                ntype;
 
-	if (tree == NULL) {
-		nxt = mem_tree_get(op, addr, nxt);
-	}
+	// if (tree == NULL) {
```
NikhilBirgade, line 2026: Remove code if not needed.

btree/btree.c
```diff
@@ -4306,8 +4204,20 @@ int64_t btree_open_tree_tick(struct m0_sm_op *smop)
...
+			oi->i_nop.no_tree->t_type = bop->b_data.bt;
+
+		/**
+		 * When tree_open() is called after tree_close() and the tree
+		 * descriptor associated with the tree is freed during the
+		 * tree_close(), we will get a new tree_descriptor which wont
+		 * contain the tree_height. This data is then filled using the
+		 * tree's t_height.
+		 */
+		if (oi->i_nop.no_tree->t_height == 0)
```
NikhilBirgade, line 4217: It will be better to update height as per the height stored in be node. The `if` check can be removed.

btree/btree.c
```diff
@@ -4365,6 +4278,8 @@ int64_t btree_close_tree_tick(struct m0_sm_op *smop)
 	case P_ACT:
 		tree->t_starttime = 0;
 		tree_put(tree);
+		M0_SET0(bop->bo_arbor->t_desc);
```
NikhilBirgade, line 4281: Not needed. If you want to set tree descriptor to NULL after freeing, use m0_free0 in tree_put().

btree/btree.c
```diff
@@ -5529,8 +5448,11 @@ void m0_btree_destroy(struct m0_btree *arbor, struct m0_btree_op *bop,
 {
 	bop->bo_arbor = arbor;
 	bop->bo_tx    = tx;
-	bop->bo_seg   = arbor->t_desc->t_seg;
-
+	if (arbor->t_desc != NULL)
+		bop->bo_seg = arbor->t_desc->t_seg;
+	else
+		bop->bo_seg = NULL;
```
NikhilBirgade, line 5454: Can you please explain the logic behind lines 5451-5454.

btree/btree.c
```diff
@@ -5627,15 +5549,15 @@ static bool                     btree_ut_initialised = false;
 static void btree_ut_init(void)
 {
 	if (!btree_ut_initialised) {
-		segops = (struct seg_ops *)&mem_seg_ops;
+		//segops = (struct seg_ops *)&mem_seg_ops;
```
NikhilBirgade, line 5552: Remove if not needed.

btree/btree.c
```diff
@@ -5627,15 +5549,15 @@ static bool                     btree_ut_initialised = false;
...
 		m0_btree_mod_init();
 		btree_ut_initialised = true;
 	}
 }
 
 static void btree_ut_fini(void)
 {
-	segops = NULL;
+	//segops = NULL;
```
NikhilBirgade, line 5560: Remove if not needed.

btree/btree.c
```diff
@@ -5985,28 +5900,24 @@ static void ut_basic_tree_oper(void)
 	temp_node = m0_alloc_aligned((1024 + sizeof(struct nd)), 10);
 	btree = m0_alloc(sizeof *btree);
 	rc = M0_BTREE_OP_SYNC_WITH_RC(&b_op, m0_btree_create(temp_node, 1024,
-							     &btree_type, nt,
-							     &b_op, seg, tx));
+							      						 &btree_type, nt, &b_op,
```
NikhilBirgade, line 5903: Alignment.

btree/btree.c
```diff
@@ -5985,28 +5900,24 @@ static void ut_basic_tree_oper(void)
...
+														 seg, tx));
 	M0_ASSERT(rc == 0);
 
-	rc = M0_BTREE_OP_SYNC_WITH_RC(&b_op, m0_btree_close(b_op.bo_arbor,
-							    &b_op));
+	rc = M0_BTREE_OP_SYNC_WITH_RC(&b_op, m0_btree_close(b_op.bo_arbor, &b_op));
```
NikhilBirgade, line 5907: Alignment.

btree/btree.c
```diff
@@ -5985,28 +5900,24 @@ static void ut_basic_tree_oper(void)
...
 	M0_ASSERT(rc == 0);
 	temp_btree = b_op.bo_arbor;
 	rc = M0_BTREE_OP_SYNC_WITH_RC(&b_op, m0_btree_open(temp_node, 1024,
-							   &btree, &b_op));
+													   &btree, &b_op));
```
NikhilBirgade, line 5911: Alignment.

btree/btree.c
```diff
@@ -5985,28 +5900,24 @@ static void ut_basic_tree_oper(void)
...
 	b_op.bo_arbor = temp_btree;
 
-	if (b_op.bo_arbor->t_desc->t_ref > 0) {
-		rc = M0_BTREE_OP_SYNC_WITH_RC(&b_op,
-					      m0_btree_destroy(b_op.bo_arbor,
-							       &b_op, tx));
-		M0_ASSERT(rc == 0);
-	}
+	rc = M0_BTREE_OP_SYNC_WITH_RC(&b_op, m0_btree_destroy(b_op.bo_arbor, &b_op,
+														  tx));
```
NikhilBirgade, line 5919: Alignment.

btree/btree.c
```diff
@@ -6078,12 +5989,9 @@ static void ut_basic_tree_oper(void)
 	 */
 
 	/** Destory it */
-	if (b_op.bo_arbor->t_desc->t_ref > 0) {
-		rc = M0_BTREE_OP_SYNC_WITH_RC(&b_op,
-					      m0_btree_destroy(b_op.bo_arbor,
-							       &b_op, tx));
-		M0_ASSERT(rc == 0);
-	}
+	rc = M0_BTREE_OP_SYNC_WITH_RC(&b_op, m0_btree_destroy(b_op.bo_arbor, &b_op,
```
NikhilBirgade, line 5992: Alignment.

btree/btree.c
```diff
@@ -7354,13 +7244,9 @@ static void btree_ut_tree_oper_thread_handler(struct btree_ut_thread_info *ti)
...
 
-		if (b_op.bo_arbor->t_desc->t_ref > 0) {
-			rc = M0_BTREE_OP_SYNC_WITH_RC(&b_op,
-						      m0_btree_destroy(tree,
-								       &b_op,
-								       tx));
-			M0_ASSERT(rc == 0);
-		}
+
+		rc = M0_BTREE_OP_SYNC_WITH_RC(&b_op, m0_btree_destroy(tree,	&b_op, tx));
```
NikhilBirgade, line 7248: Alignment.

btree/btree.c
```diff
@@ -7155,26 +7058,13 @@ static void btree_ut_num_threads_num_trees_kv_oper(int32_t thread_count,
...
-	/**
-	 * Commenting this code as the delete operation is not done here.
-	 * Due to this, the destroy operation will crash.
-	 *
-	 *
-	 * M0_BTREE_OP_SYNC_WITH_RC(&b_op,
-	 *				 m0_btree_destroy(b_op.bo_arbor, &b_op));
-	 */
+	rc = M0_BTREE_OP_SYNC_WITH_RC(&b_op, m0_btree_destroy(b_op.bo_arbor, 
+														  &b_op, tx));
```
NikhilBirgade, line 7066: Alignment.

btree/btree.c
ydb242: Yes. I missed that part.
I will update it.

btree/btree.c
ydb242: This code is used in UT only as this code considers that we dont have space allocated for root node, and then it goes ahead an creates a tree with empty root node. This code has been put under if #0 by shashank as there is a bit dilemma whether to keep it or not.

btree/btree.c
ydb242: It is a question between node in BE segment and node descriptor. The memory being free is of node in BE segment whereas the lock taken is of node descriptor. I dont think this will result in an error.

btree/btree.c
ydb242: Yes.

btree/btree.c
ydb242: Ok

btree/btree.c
ydb242: Ok.
I will check this and update.

btree/btree.c
ydb242: I wanted to make the m0_btree_destroy code more robust so that it handles incorrect cases inside it. Thus, in the case of tree_close() -> tree_destroy(), it might be possible that t_desc has been freed earlier. Therefore, we might get a segmentation fault here. To avoid that and handle that case properly, I have added these lines.

btree/btree.c
ydb242: Yes. Will do.

btree/btree.c
ydb242: Yes. Will do.

btree/btree.c
ydb242: Tab size on the editor was 4 instead of 8. I will update this.

btree/btree.c
ydb242: Ok. I will update it.

btree/btree.c
ydb242: Ok. I will update it.

btree/btree.c
ydb242: Ok. I will update it.

btree/btree.c
ydb242: Ok. I will update it.

btree/btree.c
ydb242: Ok. I will update it.

btree/btree.c
ydb242: Ok. I will update it.

btree/btree.c
ydb242: Hmm. I realise that it wont be NULL.
I will remove this check.

btree/btree.c
ydb242: Currently, yes. I have assumed that. I need to see if we have a check to determine this.

btree/btree.c
```diff
@@ -1926,9 +1961,15 @@ static int64_t tree_delete(struct node_op *op, struct td *tree,
 			   struct m0_be_tx *tx, int nxt)
 {
 	M0_PRE(tree != NULL);
-	return segops->so_tree_delete(op, tree, tx, nxt);
+	struct nd *root = tree->t_root;
+
+	op->no_tree = tree;
+	op->no_node = root;
+	node_free(op, op->no_node, tx, nxt);
```
RadhaGulhane13, line 1990: you have not called node_fini before freeing node. I think . it is required, please check it once.

btree/btree.c
RadhaGulhane13: when node_free is called , it will free BE segment memory but if n_ref and node->n_txref, both are 0, it will also free node_descriptor.

btree/btree.c
ydb242: I think we will need to keep this as after freeing the tree_descriptor and again opening the tree, this is the only way to get the  tree height as we dont store height in BE node.

btree/btree.c
```diff
@@ -4237,37 +4134,38 @@ int64_t btree_destroy_tree_tick(struct m0_sm_op *smop)
...
 
-	tree_put(bop->bo_arbor->t_desc);
-	/**
-	 * TODO: Currently just deleting the tree root node, as the assumption
-	 * is tree will only have root node at this point.
-	 */
-	m0_rwlock_write_lock(&list_lock);
-	ndlist_tlink_del_fini(bop->bo_arbor->t_desc->t_root);
-	m0_rwlock_write_unlock(&list_lock);
+	if (bop->bo_arbor->t_desc != NULL &&  bop->bo_arbor->t_desc->t_ref > 0) {
```
RadhaGulhane13: Extra space after &&.

btree/btree.c
```diff
@@ -4237,37 +4134,38 @@ int64_t btree_destroy_tree_tick(struct m0_sm_op *smop)
...
-	 * Only those fields that have changed need to be
-	 * updated.
-	 */
+		/** The following pre-condition is currently a
+		 *  compulsion as the delete routine has not been
+		 *  implemented yet.
+		 *  Once it is implemented, this pre-condition can be
+		 *  modified to compulsorily remove the records and get
+		 *  the node count to 0.
+		 */
```
RadhaGulhane13, line 4192: make changes as described in comments as delete routine has been implemented and remove the comments.

btree/btree.c
```diff
@@ -4306,8 +4204,20 @@ int64_t btree_open_tree_tick(struct m0_sm_op *smop)
 		return tree_get(&oi->i_nop, &oi->i_nop.no_addr, P_ACT);
 
 	case P_ACT:
+		if (!oi->i_nop.no_tree->t_type)
+			oi->i_nop.no_tree->t_type = bop->b_data.bt;
+
+		/**
+		 * When tree_open() is called after tree_close() and the tree
+		 * descriptor associated with the tree is freed during the
+		 * tree_close(), we will get a new tree_descriptor which wont
```
RadhaGulhane13, line 4258: apostrophe: wont->won't

btree/btree.c
```diff
@@ -4306,8 +4204,20 @@ int64_t btree_open_tree_tick(struct m0_sm_op *smop)
...
+		 * contain the tree_height. This data is then filled using the
+		 * tree's t_height.
+		 */
+		if (oi->i_nop.no_tree->t_height == 0)
+			oi->i_nop.no_tree->t_height = bop->b_data.tree->t_height;
```
RadhaGulhane13, line 4263: exceed 80 char limit.

btree/btree.c
```diff
@@ -1885,12 +1869,45 @@ static int64_t mem_tree_get(struct node_op *op, struct segaddr *addr, int nxt);
...
+	struct nd              *node = NULL;
+
+	if (addr != NULL) {
+		nxt  = node_get(op, NULL, addr, nxt);
+		node = op->no_node;
+		if (node != NULL) {
+			tree = node->n_tree;
+		} else
+			op->no_op.o_sm.sm_rc = M0_ERR(-EINVAL);
+
```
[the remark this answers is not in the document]
ydb242: I have removed this part in the latest commit.
So please check once. I think you have commented on 1 commit back.

btree/btree.c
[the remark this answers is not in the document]
ydb242: Yes, what you say is true but i dont think we should worry about this for now. The reason is that in future we might actually want to return from this position if needed due to the async changes or we might run a sub_op() here. Hence, changes can be avoided.

btree/btree.c
ydb242: Yes. I may have missed that. Let me check.

btree/btree.c
ydb242: It has been updated in the latest commit.

btree/btree.c
ydb242: ok.

btree/btree.c
ydb242: ok.

btree/btree.c
ydb242: Oh, is it? I need to check this.

btree/btree.c
ydb242: Hmm.
There is an assert in tree_put that will catch the condition if tree->t_root is NULL or not.
That being said, I need to verify the flow once again.

btree/btree.c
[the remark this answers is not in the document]
RadhaGulhane13: okay

btree/btree.c
RadhaGulhane13: But , after freeing node descriptor(t_root), it is not necessary that tree->t_root will become NULL, but it will point to the memory location which is already  been freed.

ydb242: This becomes a case of dangling pointer then which needs to be handled.
Anyways I need to rethink over this.

btree/btree.c
NikhilBirgade: `bop->b_data.tree->t_height`  is a user provided parameter. Suppose you have a tree and its height increased to 5. Then you called tree_close and shutdown the cluster. Now you boot up the cluster and call tree_open. This time you do not have any idea about tree height. You have to read the m0_btree:t_height(which is in be segment) and update the td:t_height and use it.

btree/btree.c
ydb242: On discussion, it is a prerequisite that all records should be deleted before reaching destroy.
Hence, there is no need to add the delete routine.
Comment has been removed.

btree/btree.c
```diff
@@ -1908,10 +1947,28 @@ static int64_t tree_get(struct node_op *op, struct segaddr *addr, int nxt)
...
-	return segops->so_tree_create(op, tt, rootshift, tx, nxt);
+	struct td *tree;
+
+	/**
+	 * Creates root node and then assigns a tree descriptor for this root
+	 * node.
+	 */
+
+	tree_get(op, NULL, nxt);
+
```
RadhaGulhane13, line 1958: error handling, as tree_get can return error code

btree/btree.c
```diff
@@ -1885,12 +1892,44 @@ static int64_t mem_tree_get(struct node_op *op, struct segaddr *addr, int nxt);
...
+			tree->t_starttime = 0;
+			tree->t_root = node;
+			tree->t_height = node_level(node) + 1;
+			node_lock(node);
+			node->n_tree = tree;
+			node_unlock(node);
+
+		} else {
+			m0_rwlock_write_lock(&tree->t_lock);
+			tree->t_ref++;
```
RadhaGulhane13, line 1920: It may happen by the time you reach this point tree descriptor may become freed by other thread executing tree_put operation. So, in this case you are trying to take on lock which doesn't exists.

Better to use global lock instead of tree_lock and this global locks need to be taken at the beginning of tree_get and tree_put operation and released at the end of these operation, In this way you can avoid above scenario.

btree/btree.c
```diff
@@ -4306,9 +4245,12 @@ int64_t btree_open_tree_tick(struct m0_sm_op *smop)
 		return tree_get(&oi->i_nop, &oi->i_nop.no_addr, P_ACT);
 
 	case P_ACT:
+		if (!oi->i_nop.no_tree->t_type)
```
RadhaGulhane13, line 4248: Better to check if (oi->i_nop.no_op.o_sm.sm_rc == 0) as currently code will work, as we always returning -ve values in  case of error condition, But in future we may change op.c which will persist sm_rc and return next state , so in such case you may want to check oi->i_nop.no_op.o_sm.sm_rc in next state i.e. P_ACT. you may refer to P_NEXTDOWN state which is doing same thing. 
Or at least for now, you can have ASSERT(oi->i_nop.no_op.o_sm.sm_rc == 0); it will make sure that tree_get operation was successful.

btree/btree.c
ydb242: This routine has not been touched yet as it is under consideration.
I will update once changes are to be made here.

btree/btree.c
ydb242: The purpose of this if condition was just to ensure that the tree being returned has tree type intact, but with the new changes it is hard to reach such a scenario.
It wont be replacement actually if we use an sm_rc based check, but that check will be useful.
I will add that.

btree/btree.c
ydb242: Ok Yes.
I will add that lock!

btree/btree.c
```diff
@@ -7879,7 +7841,8 @@ struct m0_ut_suite btree_ut = {
 	.ts_init = ut_btree_suite_init,
 	.ts_fini = ut_btree_suite_fini,
 	.ts_tests = {
-		{"basic_tree_op",                   ut_basic_tree_oper},
+		{"basic_tree_op_cp",                ut_basic_tree_oper_cp},
+		//{"basic_tree_op_icp",               ut_basic_tree_oper_icp},
```
NikhilBirgade, line 7845: Any reason behind disabling this test ?

btree/btree.c
```diff
@@ -5975,66 +5915,35 @@ static void ut_basic_tree_oper(void)
...
 	M0_ASSERT(rc == 0);
-	b_op.bo_arbor = temp_btree;
 
-	if (b_op.bo_arbor->t_desc->t_ref > 0) {
-		rc = M0_BTREE_OP_SYNC_WITH_RC(&b_op,
-					      m0_btree_destroy(b_op.bo_arbor,
-							       &b_op, tx));
-		M0_ASSERT(rc == 0);
-	}
+	rc = M0_BTREE_OP_SYNC_WITH_RC(&b_op, m0_btree_destroy(b_op.bo_arbor,
```
NikhilBirgade, line 5937: In m0_btree_close the tree descriptor will get freed. After that t_desc is deferenced in m0_btree_destroy.

btree/btree.c
ydb242: Yes, It will be modified later @NikhilBirgade

btree/btree.c
```diff
@@ -1976,17 +2048,27 @@ static int64_t node_get(struct node_op *op, struct td *tree,
...
 	 * accessing the same node descriptor concurrently.
 	 * Replace it with a different global lock once hash
 	 * functionality is implemented.
 	 */
+
 	m0_rwlock_write_lock(&list_lock);
+
+	/**
+	 * TODO: Consider adding a state here to return as we might require time
+	 * to get the list_lock.
```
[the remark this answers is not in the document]
ydb242: Ok I will update it.

btree/btree.c
```diff
@@ -2109,6 +2190,7 @@ static void node_put(struct node_op *op, struct nd *node, struct m0_be_tx *tx)
 			node_unlock(node);
 			m0_rwlock_fini(&node->n_lock);
 			m0_free(node);
+			node = NULL;
```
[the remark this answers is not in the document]
ydb242: Ok. I will update it.

btree/btree.c
ydb242: This test is the incorrect path test.
At the moment, one of the scenario i.e. calling destroy after tree has been already destroyed is not gracefully failing.
Hence, the test has been disabled.

shashank-parulekar (changes requested):

btree/btree.c
NikhilBirgade: Is it going to be part of any other ticket?

btree/btree.c
NikhilBirgade: Ok, resolving based on above comment.

btree/btree.c
NikhilBirgade: Ok.

shashank-parulekar (approved):

RadhaGulhane13 (approved):

NikhilBirgade (approved):

PR #905, opened 2021-07-30
Eos 23143 : Credit calculation for KV UPDATE

RadhaGulhane13:
- Adding credit calculation for Update Operation

### DIFF
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -2995,23 +2995,19 @@ static void btree_callback_credit(struct m0_be_tx_credit *accum)
 }
 
 /**
- * This function will calculate credits required to update @nr nodes and it will
- * add those credits to @accum.
+ * This function will calculate credits required to update node and it will add
+ * those credits to @accum.
  */
 static void btree_node_update_credit(const struct m0_btree  *tree,
-				     struct m0_be_tx_credit *accum,
-				     m0_bcount_t             nr)
+				     struct m0_be_tx_credit *accum)
 {
-	struct m0_be_tx_credit cred = {};
  	m0_bcount_t             node_size;
 	int                     shift;
 
 	shift     = node_shift(tree->t_desc->t_root);
 	node_size =  1ULL << shift;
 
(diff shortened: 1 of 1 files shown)


NikhilBirgade (approved):

btree/btree.c
```diff
@@ -2993,6 +2993,41 @@ static void btree_callback_credit(struct m0_be_tx_credit *accum)
...
+	struct m0_be_tx_credit cred = {};
+ 	m0_bcount_t             node_size;
+	int                     shift;
+
+	shift     = node_shift(tree->t_desc->t_root);
+	node_size =  1ULL << shift;
+
+	cred = M0_BE_TX_CREDIT(1, node_size);
+
+	m0_be_tx_credit_mac(accum, &cred, nr);
```
shashank-parulekar, line 3014: The credit_mac is not needed here since the caller anyway does it as the nr passed to this function is just 1 so mac

shashank-parulekar (changes requested):

btree/btree.c
RadhaGulhane13: for calculation of update credit, we are passing nr= 1 , but in case of put or delete, we may need to calculate credits two update more than 1 nodes, Therefore I am using m0_be_tx_credit_mac(accum, &cred, nr);.

shashank-parulekar: But I assume that would come as a parameter to the below routine m0_btree_update_credit() and not this routine btree_node_update_credit()

shashank-parulekar (changes requested):

btree/btree.c
RadhaGulhane13: Yes . But btree_node_update_credit() is routine  to calculate credit require to update **node** and depending on the operation we may need to calculate credits to update more than 1 node. Therefore taking nr as parameter

RadhaGulhane13: Made changes as per the suggestion as it won't make much difference.

shashank-parulekar (approved):

PR #910, opened 2021-08-02
Balloc ut changes for group count and groupsizes

nihar1313:
*) New UT has been add for group sizes and group count tunable.
*) code reformat  for ut/balloc.c


nihar1313:
This is no longer valid, New PR has been raised.

PR #913, opened 2021-08-03
Eos 19527 : Credit calculation for KV Put

RadhaGulhane13:


btree/btree.c
```diff
@@ -2928,6 +2928,91 @@ static void btree_callback_credit(struct m0_be_tx_credit *accum)
...
+	/* Credits for split operation */
+	btree_node_split_credit(tree, &cred);
+	m0_be_tx_credit_mac(accum, &cred, MAX_TREE_HEIGHT);
+
+	/**
+	 * Additional credits required when extranode is used when height needs
+	 * to be inceased.
+	*/
+	btree_node_alloc_credit(tree, accum);
+	btree_node_update_credit(tree, accum, 1);
```
shashank-parulekar, line 2999: If on line 2992 you have taken care of credits for MAX_TREE_HEIGHT then why are we still worrying about extranode ?
Because during extra node the tree height will be just 2 !!!

shashank-parulekar (changes requested):

btree/btree.c
RadhaGulhane13: Yes right. we do not need to allocate credit for extra node, made these changes

shashank-parulekar (approved):

NikhilBirgade (approved):

ydb242 (approved):

PR #914, opened 2021-08-03
"EOS-21909: unit test updates for balloc tunables."

nihar1313:
Signed-off-by: Nihar Nayak <nihar.nayak@seagate.com>

### DIFF
--- a/balloc/ut/balloc.c
+++ b/balloc/ut/balloc.c
@@ -42,10 +42,6 @@
 #include "stob/ad.h"      /* m0_stob_ad_spares_calc */
 
 #define BALLOC_DBNAME "./__balloc_db"
-
-#define GROUP_COUNT (BALLOC_DEF_CONTAINER_SIZE / (BALLOC_DEF_BLOCKS_PER_GROUP * \
-						 (1 << BALLOC_DEF_BLOCK_SHIFT)))
-
 #define BALLOC_DEBUG
 
 static const int    MAX     = 10;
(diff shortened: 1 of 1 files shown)


balloc/ut/balloc.c
```diff
@@ -263,52 +252,130 @@ int test_balloc_ut_ops(struct m0_be_ut_backend *ut_be, struct m0_be_seg *seg,
...
-	m0_be_ut_backend_fini(&ut_be);
+void test_balloc_group_size()
+{
+	struct m0_ad_balloc_format_req  bcfg;
+	int                             iter;
+
+	/* Test various group size */
+	for (iter = 0; iter <= 4; iter++) {
+		prepare_balloc_config(&bcfg, BALLOC_DEF_CONTAINER_SIZE,
+				      BALLOC_DEF_BLOCKS_PER_GROUP * (1 << iter),
```
mukundkanekar, line 331: In /etc/sysconfig/motr we configure MOTR_M0D_BALLOC_GROUP_COUNT where as in this test we are scaling BALLOC_DEF_BLOCKS_PER_GROUP which will also have similar impact, but would you line to update /etc/sysconfig/motr  also with MOTR_M0D_BALLOC_DEF_BLOCKS_PER_GROUP so that the configuration is in sync.

nihar1313: Added new TC  for different group count including TC for group size.  did not disturb prepare_balloc_config() to include group count as parameter as  other TC will be affected..

mukundkanekar: ok.

mukundkanekar (approved):

vidyadhar-pinglikar (approved):

PR #917, opened 2021-08-04
Return error codes instead of record's error flags in get and iter tick.

NikhilBirgade:

### DIFF
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -4499,9 +4499,7 @@ int  btree_sibling_first_key(struct m0_btree_oimpl *oi, struct td *tree,
 			return 0;
 		}
 	}
-	s->s_rec.r_flags = M0_BSC_KEY_NOT_FOUND;
-	return 0;
-
+	return M0_ERR(-ENOENT);
 }
 
 /**
@@ -4696,6 +4694,7 @@ static int64_t btree_get_kv_tick(struct m0_sm_op *smop)
 								     &s);
 					if (rc != 0) {
 						node_op_fini(&oi->i_nop);
+						lock_op_unlock(tree);
 						return fail(bop, rc);
 					}
 				}
@@ -4706,9 +4705,11 @@ static int64_t btree_get_kv_tick(struct m0_sm_op *smop)
 				s.s_idx = bop->bo_opc == M0_BO_MINKEY ? 0 :
 					  count - 1;
 				node_rec(&s);
-			} else
+			} else {
 				/** Only root node is present and is empty. */
-				s.s_rec.r_flags = M0_BSC_KEY_NOT_FOUND;
+				lock_op_unlock(tree);
+				return fail(bop, M0_ERR(-ENOENT));
+			}
 		}
 
 		bop->bo_cb.c_act(&bop->bo_cb, &s.s_rec);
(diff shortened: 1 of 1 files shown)


btree/btree.c
```diff
@@ -4499,9 +4499,7 @@ int  btree_sibling_first_key(struct m0_btree_oimpl *oi, struct td *tree,
 			return 0;
 		}
 	}
-	s->s_rec.r_flags = M0_BSC_KEY_NOT_FOUND;
-	return 0;
-
+	return M0_ERR(-ENOENT);
```
RadhaGulhane13, line 4502: you can return M0_BSC_KEY_NOT_FOUND instead of M0_ERR(-ENOENT) as we need to return linux codes only value for M0_BSC_KEY_NOT_FOUND is assigned equal to ENOENT. so that returning M0_BSC_KEY_NOT_FOUND will provide more readability and we will return linux codes only.

Please do the same changes for rest of places.

btree/btree.c
```diff
@@ -4706,9 +4705,11 @@ static int64_t btree_get_kv_tick(struct m0_sm_op *smop)
...
-			} else
+			} else {
 				/** Only root node is present and is empty. */
-				s.s_rec.r_flags = M0_BSC_KEY_NOT_FOUND;
+				lock_op_unlock(tree);
+				return fail(bop, M0_ERR(-ENOENT));
+			}
 		}
 
 		bop->bo_cb.c_act(&bop->bo_cb, &s.s_rec);
```
RadhaGulhane13, line 4715: set sm_rc flag as M0_BSC_SUCCESS , as in ut's we are adding ASSERT based rc == M0_BSC_SUCCESS , currently M0_BSC_SUCCESS code is 0 even if we do not set sm_rc to M0_BSC_SUCCESS , it won't fail but to avoid future changes you might want to set sm_rc to M0_BSC_SUCCESS if operation is successful.

Same thing need to be done in iter_tick routine.

btree/btree.c
```diff
@@ -7279,22 +7278,23 @@ static void btree_ut_kv_oper_thread_handler(struct btree_ut_thread_info *ti)
...
-					 m0_btree_get(tree,
-						      &rec.r_key, &ut_get_cb,
-						      BOF_EQUAL, &kv_op));
-		M0_ASSERT(get_data.flags == M0_BSC_SUCCESS);
+		rc = M0_BTREE_OP_SYNC_WITH_RC(&kv_op,
+					      m0_btree_get(tree,
+							   &rec.r_key,
+							   &ut_get_cb,
+							   BOF_EQUAL, &kv_op));
+		M0_ASSERT(rc == 0);
```
RadhaGulhane13, line 7286: rc == M0_BSC_SUCCESS instead rc == 0

btree/btree.c
```diff
@@ -7334,21 +7334,20 @@ static void btree_ut_kv_oper_thread_handler(struct btree_ut_thread_info *ti)
...
 		get_key[0] = key_last + 1;
 
-		M0_BTREE_OP_SYNC_WITH_RC(&kv_op,
-					 m0_btree_minkey(tree, &ut_get_cb, 0,
-							 &kv_op));
-		M0_ASSERT(get_data.flags == M0_BSC_SUCCESS);
+		rc = M0_BTREE_OP_SYNC_WITH_RC(&kv_op,
+					      m0_btree_minkey(tree, &ut_get_cb,
+							      0, &kv_op));
+		M0_ASSERT(rc == 0);
```
RadhaGulhane13, line 7340: rc == M0_BSC_SUCCESS instead rc == 0 do same for rest of the places.

shashank-parulekar (approved):

btree/btree.c
RadhaGulhane13: in btree.h value for M0_BSC_KEY_NOT_FOUND  is assigned equal to ENOENT, therefore I suggested to return M0_BSC_KEY_NOT_FOUND. But if M0_BSC_* are purposed for r_flags and we should return ENOENT , above implementation is correct. I will make similar changes in put and delete as in put I am returning M0_BSC_KEY_NOT_FOUND instead of ENOENT.

RadhaGulhane13 (approved):

ydb242 (approved):

PR #920, opened 2021-08-05
Adding interfaces for credits calculation for KV operation 

RadhaGulhane13:


btree/btree.h
```diff
@@ -316,6 +316,24 @@ void m0_btree_op_fini(struct m0_btree_op *bop);
 void m0_btree_op_credit(const struct m0_btree_op *bt,
 			struct m0_be_tx_credit *cr);
```
RadhaGulhane13, line 318: Not using m0_btree_op_credit as  we are calculating credits before calling m0_btree_put/ m0_btree_update etc, and m0_btree_op will get filled in m0_btree_put this routine so operation type will not be available at time of calling credit calculation if we are using m0_btree_op_credit . Therefore, adding separate interface for credit calculation for each operation.

btree/btree.c
```diff
@@ -6521,7 +6525,7 @@ static void ut_basic_kv_oper(void)
 		void                *v_ptr  = &value;
 
 		cred = M0_BE_TX_CB_CREDIT(0, 0, 0);
-		m0_btree_put_credit(tree, &cred, 1);
+		m0_btree_put_credit(tree, 1, -1, -1, &cred);
```
shashank-parulekar, line 6528: You cannot send -1 for key size and value size. You need to send the ksize and vsize values. The caller always knows the size of key and value of the record which he will be putting.

shashank-parulekar (changes requested):

NikhilBirgade (approved):

shashank-parulekar (approved):

PR #921, opened 2021-08-05
EOS-22606: Credit Calculation for tree_create and tree_destroy

ydb242:


btree/btree.c
```diff
@@ -3021,6 +3031,20 @@ static void m0_btree_update_credit(const struct m0_btree  *tree,
...
+{
+	int size = nt->nt_credit_size();
+	m0_be_tx_credit_add(accum, &M0_BE_TX_CREDIT(1, size));
+}
+
+void m0_btree_destroy_credit(struct m0_btree *tree,
+			     struct m0_be_tx_credit *accum)
+{
+	int size = tree->t_desc->t_root->n_type->nt_credit_size();
+	m0_be_tx_credit_add(accum, &M0_BE_TX_CREDIT(1, size));
```
RadhaGulhane13, line 3046: At time of destroy you are also freeing the node, don't you need to calculate credits for node free.

ydb242: We will need to calculate the credits for destroy operation.
This is because we call node_fini() in it and not node_free().
We will go ahead and free the node_descriptor but not the node in the BE segment as this node is not allocated by the tree_ops. We get the node from outside and then we go ahead to initialise and finalise the header for that particular node.

RadhaGulhane13 (approved):

btree/btree.c
```diff
@@ -1067,6 +1067,9 @@ struct node_type {
 	/** Captures node data in segment */
 	void (*nt_capture)(struct slot *slot, struct m0_be_tx *tx);
 
+	/** Returns the header size for credit calculation of tree operations */
+	int  (*nt_credit_size)(void);
```
shashank-parulekar, line 1071: The name should be indicating the operation. In this case it could be nt_create_delete_credit_size.
This is because the same will be used for other credit calculations on this node too like record_add and record_del !!!

btree/btree.c
```diff
@@ -2356,6 +2359,7 @@ static bool ff_verify(const struct nd *node);
 static void ff_opaque_set(const struct segaddr *addr, void *opaque);
 static void *ff_opaque_get(const struct segaddr *addr);
 static void ff_capture(struct slot *slot, struct m0_be_tx *tx);
+static int  ff_header_size(void);
```
shashank-parulekar, line 2362: same comment as above. Change the name to something which identifies the operation for which this function returns the credit count.

shashank-parulekar (changes requested):

btree/btree.c
ydb242: Updated.

btree/btree.c
ydb242: Updated.

NikhilBirgade (approved):

shashank-parulekar (approved):

PR #930, opened 2021-08-06
EOS-22533 : Credits for delete operation

RadhaGulhane13:

### DIFF
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -1070,6 +1070,13 @@ struct node_type {
 	/** Returns the header size for credit calculation of tree operations */
 	int  (*nt_create_delete_credit_size)(void);
 
+	/**
+	 * Calculates credits required to free the node and adds those credits
+	 * to @accum
+	 */
+	void (*nt_node_free_credits)(const struct nd *node,
+				     struct m0_be_tx_credit *accum);
+
 	/** Gets key size from segment. */
 	/* uint16_t (*nt_ksize_get)(const struct segaddr *addr); */
 
@@ -1671,7 +1678,15 @@ static void node_fini(const struct nd *node, struct m0_be_tx *tx)
 {
 	node->n_type->nt_fini(node, tx);
 }
+
+static void node_free_credits(const struct nd *node,
+			      struct m0_be_tx_credit *accum)
+{
+	node->n_type->nt_node_free_credits(node, accum);
+}
+
 #endif
+
 static struct mod *mod_get(void)
 {
 	return m0_get()->i_moddata[M0_MODULE_BTREE];
@@ -2359,6 +2374,8 @@ static bool ff_verify(const struct nd *node);
 static void ff_opaque_set(const struct segaddr *addr, void *opaque);
 static void *ff_opaque_get(const struct segaddr *addr);
 static void ff_capture(struct slot *slot, struct m0_be_tx *tx);
+static void ff_node_free_credits(const struct nd *node,
+				 struct m0_be_tx_credit *accum);
 static int  ff_create_delete_credit_size(void);
 /* uint16_t ff_ksize_get(const struct segaddr *addr); */
 /* uint16_t ff_valsize_get(const struct segaddr *addr);  */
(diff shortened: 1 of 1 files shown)


btree/btree.c
```diff
@@ -6647,6 +6716,7 @@ static void ut_basic_kv_oper(void)
 		void                *k_ptr  = &key;
 
 		cred = M0_BE_TX_CB_CREDIT(0, 0, 0);
+		m0_btree_del_credit(tree, 1, ksize, -1, &cred);
```
RadhaGulhane13, line 6719: Passing -1 as value size is not available for delete operation.

btree/btree.c
```diff
@@ -2921,6 +2939,19 @@ static void ff_capture(struct slot *slot, struct m0_be_tx *tx)
...
 
+static void ff_node_free_credits(const struct nd *node,
+				 struct m0_be_tx_credit *accum)
+{
+	struct ff_head *h           = ff_data(node);
+	int             shift       = h->ff_shift;
+	int             node_size   = 1ULL << shift;
+	int             header_size = sizeof(*h);
+
+	m0_be_allocator_credit(NULL, M0_BAO_FREE_ALIGNED,
```
ydb242, line 2950: ff_node_free_credits() should return only the size that you want to add to your credits.
Actual addition of credits should be done outside of this function, i.e. in btree_del_credit()
Otherwise for every node_type() you will be calling an m0_be_allocator_credit() and m0_be_tx_credit_add() which is not good.
With this change, the function name will also require some modification.

shashank-parulekar: @ydb242 The credit addition happens in the credit functions in the old btree code. They seem to take the last parameter 'accum' and add the credits which they need to the passed accum structure.

btree/btree.c
```diff
@@ -2973,6 +3004,28 @@ static void btree_node_update_credit(const struct m0_btree  *tree,
...
+	/* Credits for freeing the node. */
+	node_free_credits(tree->t_desc->t_root, &cred);
+	m0_be_tx_credit_mac(accum, &cred, MAX_TREE_HEIGHT);
+
+	/**
+	 * Credits for update node:
+	 * After deletetion of record, if node is not underflowed, node will get
+	 * updated.
+	 */
+	btree_node_update_credit(tree, accum);
```
shashank-parulekar, line 3026: This is not needed because if we did not underflow then the credits calculated above for the node free are more than enough to capture the node update if a record was deleted.

btree/btree.c
```diff
@@ -3019,6 +3072,22 @@ void m0_btree_put_credit(const struct m0_btree  *tree,
 	m0_be_tx_credit_mac(accum, &cred, nr);
 }
 
+/**
+ * This function will calculate credits required to perform @nr delete KV
+ * operations and it will add those credits to @accum.
+ */
+void m0_btree_del_credit(const struct m0_btree  *tree,
+			 m0_bcount_t             nr,
```
shashank-parulekar, line 3080: Do we ever have nr > 1 ? As far as the del_kv_tick routine is concerned I think you will be doing one record delete at a time. While having nr is fine the call to m0_be_tx_credit_mac at the end of this function will consume a lot of credits.

btree/btree.c
shashank-parulekar: I agree but this might become a problem for us because we can never be able to get the proper credits since this will work for fixed format but might be an issue for variable value size as we discussed.

shashank-parulekar (changes requested):

btree/btree.c
RadhaGulhane13: Yes, nr > 1 , for /cas/ctg_store.c: ctg_meta_delete_credit(), /be/extmap.c: m0_be_emap_credit , etc.

btree/btree.c
RadhaGulhane13: Currently passing ksize, vsize param to node_free_credit function , in future when we implement node formats which require key, value size , we can use this param, but in delete routine, currently we are taking only key from user and not value, So we might want toc check if user is  going to provide whole record, this way we can get value size, else for credit calculation, we might want to store max value size somewhere.

btree/btree.c
RadhaGulhane13: Right. I have made this change.

NikhilBirgade (approved):

ydb242 (approved):

shashank-parulekar (approved):

PR #939, opened 2021-08-11
Adding node interfaces for credit calculation for kv_operation

RadhaGulhane13:

### DIFF
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -1729,6 +1729,7 @@ static void node_rec_put_credit(const struct nd *node, m0_bcount_t ksize,
 	node->n_type->nt_rec_put_credit(node, ksize, vsize, accum);
 }
 
+#if 0
 static void node_rec_update_credit(const struct nd *node, m0_bcount_t ksize,
 				   m0_bcount_t vsize,
 				   struct m0_be_tx_credit *accum)
@@ -1736,7 +1737,6 @@ static void node_rec_update_credit(const struct nd *node, m0_bcount_t ksize,
 	node->n_type->nt_rec_update_credit(node, ksize, vsize, accum);
 }
 
-#if 0
 static void node_rec_del_credit(const struct nd *node, m0_bcount_t ksize,
 				m0_bcount_t vsize,
 				struct m0_be_tx_credit *accum)
@@ -3629,54 +3629,6 @@ static void btree_callback_credit(struct m0_be_tx_credit *accum)
 {
 	accum->tc_cb_nr += BTREE_CALLBACK_CREDIT;
 }
-#if 0
-/**
- * This function will calculate credits required to allocate node and it will
- * add those credits to @accum.
- */
-static void btree_node_alloc_credit(const struct m0_btree  *tree,
-				    struct m0_be_tx_credit *accum)
-{
-	m0_bcount_t             node_size;
-	int                     shift;
-
-	shift     = node_shift(tree->t_desc->t_root);
-	node_size =  1ULL << shift;
-
-	m0_be_allocator_credit(NULL, M0_BAO_ALLOC_ALIGNED,
-			       node_size, shift, accum);
-}
-
-/**
- * This function will calculate credits required to update node and it will add
- * those credits to @accum.
- */
-static void btree_node_update_credit(const struct m0_btree  *tree,
-				     struct m0_be_tx_credit *accum)
-{
- 	m0_bcount_t             node_size;
-	int                     shift;
-
-	shift     = node_shift(tree->t_desc->t_root);
-	node_size =  1ULL << shift;
-
-	m0_be_tx_credit_add(accum, &M0_BE_TX_CREDIT(1, node_size));
-}
-#endif
-
-/**
- * This function will calculate credits required for the delete KV operation and
- * add those credits to @accum.
- */
-static void btree_del_credit(const struct m0_btree  *tree, m0_bcount_t ksize,
-			     m0_bcount_t vsize, struct m0_be_tx_credit *accum)
-{
-	struct m0_be_tx_credit cred = {};
-
-	/* Credits for freeing the node. */
-	node_free_credit(tree->t_desc->t_root, ksize, vsize, &cred);
-	m0_be_tx_credit_mac(accum, &cred, MAX_TREE_HEIGHT);
-}
 
 /**
  * This function will calculate credits required to split node and it will add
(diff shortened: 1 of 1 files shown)


btree/btree.c
```diff
@@ -1071,11 +1071,25 @@ struct node_type {
...
+
+	/**
+	 * Calculates credit required for the update KV operation and adds those
+	 * credit to @cred.
+	 */
+	void (*nt_update_credit) (const struct nd *node, m0_bcount_t ksize,
+			       m0_bcount_t vsize, struct m0_be_tx_credit *cred);
+
+	/**
+	 * Calculates credit required for the update KV operation and adds those
```
NikhilBirgade, line 1088: /s/update/del

btree/btree.c
```diff
@@ -2939,145 +2953,128 @@ static void ff_capture(struct slot *slot, struct m0_be_tx *tx)
...
 
 /**
- * This function will calculate credits required to update node and it will add
+ * This function will calculate credits required to split node and it will add
  * those credits to @accum.
  */
-static void btree_node_update_credit(const struct m0_btree  *tree,
-				     struct m0_be_tx_credit *accum)
+static void ff_node_split_credit(const struct nd *node,
+				 struct m0_be_tx_credit *cred)
```
shashank-parulekar, line 3006: We do not need to have a ff_node_split_credit() because nodes themselves do not know about splitting. The concept of splitting is done above the node layer. So the higher layer *_tick() functions should do the split credit calculations if needed using the node credit functions already provided.

btree/btree.c
```diff
@@ -2939,145 +2953,128 @@ static void ff_capture(struct slot *slot, struct m0_be_tx *tx)
...
+			  m0_bcount_t vsize, struct m0_be_tx_credit *cred)
 {
-	struct m0_be_tx_credit cred = {};
-
-	/* Credits for freeing the node. */
-	node_free_credits(tree->t_desc->t_root, ksize, vsize, &cred);
-	m0_be_tx_credit_mac(accum, &cred, MAX_TREE_HEIGHT);
+	struct m0_be_tx_credit put_cred = {};
+	/* Credits for split operation */
+	ff_node_split_credit(node, &put_cred);
```
shashank-parulekar, line 3024: ff_put_credit() should only cover the credits needed for adding a record to this node. Splitting is not known at the node layer and hence node functions should not calculate credits for splitting. This should again be done at the higher layers like m0_node_put_credit().

btree/btree.c
```diff
@@ -2939,145 +2953,128 @@ static void ff_capture(struct slot *slot, struct m0_be_tx *tx)
...
+	m0_be_tx_credit_mac(cred, &put_cred, MAX_TREE_HEIGHT);
 }
 
-/**
- * This function will calculate credits required to split node and it will add
- * those credits to @accum.
- */
-static void btree_node_split_credit(const struct m0_btree  *tree,
-				    struct m0_be_tx_credit *accum)
+static void ff_update_credit(const struct nd *node, m0_bcount_t ksize,
```
shashank-parulekar, line 3028: Why is this function needed when you have ff_node_update_credit() ???? Only one of them needs to exist unless they do something different logically and based on their scope. Currently the ff_* routines take care of all the node services for ff formatted node.

btree/btree.c
```diff
@@ -2939,145 +2953,128 @@ static void ff_capture(struct slot *slot, struct m0_be_tx *tx)
...
- */
-static void btree_callback_credit(struct m0_be_tx_credit *accum)
+static void ff_node_update_credit(const struct nd *node,
+				   struct m0_be_tx_credit *cred)
 {
-	accum->tc_cb_nr += BTREE_CALLBACK_CREDIT;
+	int         shift     = ff_shift(node);
+	m0_bcount_t node_size = 1ULL << shift;
+
+	m0_be_tx_credit_add(cred, &M0_BE_TX_CREDIT(1, node_size));
```
shashank-parulekar, line 2980: s/1/2/
If you look at ff_capture then it calls the CAPTURE routine twice based on a condition. So we can keep the same node_size but the region count should be 2.
Also make similar changes for ff_put_credit() and ff_del_credit() if needed.

btree/btree.c
```diff
@@ -2939,145 +2953,128 @@ static void ff_capture(struct slot *slot, struct m0_be_tx *tx)
...
 {
-	struct m0_be_tx_credit cred = {};
-
-	/* Credits for freeing the node. */
-	node_free_credits(tree->t_desc->t_root, ksize, vsize, &cred);
-	m0_be_tx_credit_mac(accum, &cred, MAX_TREE_HEIGHT);
+	struct m0_be_tx_credit put_cred = {};
+	/* Credits for split operation */
+	ff_node_split_credit(node, &put_cred);
+	m0_be_tx_credit_mac(cred, &put_cred, MAX_TREE_HEIGHT);
```
shashank-parulekar, line 3025: MAX_TREE_HEIGHE is not known to ff_* routines. The ff_put_credit() only gives the credit count for this node. The rest of the addition/multiplication needs to be done by the caller routine - m0_put_credit() which knows about the tree height and also any tree related changes happening as a part of put_tick().

shashank-parulekar (changes requested):

btree/btree.c
RadhaGulhane13: I think there is some confusion related to function uses. Need discussion over it.

btree/btree.c
RadhaGulhane13: Made changes in code after this comment ,  this part is outdated now.

btree/btree.c
RadhaGulhane13: Made changes as per the discussion.

btree/btree.c
RadhaGulhane13: Made changes as per the discussion.

btree/btree.c

btree/btree.c
```diff
@@ -3565,12 +3683,14 @@ static void btree_del_credit(const struct m0_btree  *tree, m0_bcount_t ksize,
  * those credits to @accum.
  */
 static void btree_node_split_credit(const struct m0_btree  *tree,
+				    m0_bcount_t             ksize,
+				    m0_bcount_t             vsize,
 				    struct m0_be_tx_credit *accum)
 {
-	btree_node_alloc_credit(tree, accum);
+	node_alloc_credit(tree->t_desc->t_root, ksize, vsize, accum);
```
shashank-parulekar, line 3690: This call should be after variable declaration.

btree/btree.c
```diff
@@ -3634,7 +3756,7 @@ void m0_btree_update_credit(const struct m0_btree  *tree,
 {
 	struct m0_be_tx_credit cred = {};
 
-	btree_node_update_credit(tree, &cred);
+	node_rec_update_credit(tree->t_desc->t_root, ksize, vsize, &cred);
 	m0_be_tx_credit_mac(accum, &cred, nr);
```
shashank-parulekar, line 3760: This implementation is okay for fixed format; for other formats when the updated value is of different size than the original value there might be changes to other nodes in the tree so this call might be the same as m0_btree_put_credit().

shashank-parulekar (changes requested):

ydb242 (approved):

NikhilBirgade (approved):

PR #944, opened 2021-08-12
Btree be seg support

shashank-parulekar:
Provide support for persisting btree data. This is initial code checkin. 
The code seems to work fine though the multi_thread_kv ut's are failing. 
Macro AVOID_BE_SEGMENT is provided to turn ON/OFF the persistence feature 
for running ut's.


btree/btree.c
```diff
@@ -4380,8 +4409,13 @@ int64_t btree_create_tree_tick(struct m0_sm_op *smop)
 		bop->bo_arbor->t_desc->t_height = bop->bo_arbor->t_height;
 		m0_rwlock_write_unlock(&bop->bo_arbor->t_desc->t_lock);
 
+		_slot.s_node      = oi->i_nop.no_node;
+		_slot.s_idx       = 0;
+		node_capture(&_slot, bop->bo_tx);
```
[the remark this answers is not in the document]
shashank-parulekar: yes we do. Good catch.

btree/btree.c
```diff
@@ -4279,6 +4286,7 @@ void m0_btree_del_credit(const struct m0_btree  *tree,
 	node_free_credit(tree->t_desc->t_root, ksize, vsize, &cred);
 	m0_be_tx_credit_mul(&cred, MAX_TREE_HEIGHT);
 
+	btree_callback_credit(&cred);
```
RadhaGulhane13, line 4289: if we call btree_callback_credit here, for nr > 1,for accum,  tc_cb_nr will be still equal to BTREE_CALLBACK_CREDIT where as it should be nr * BTREE_CALLBACK_CREDIT as m0_be_tx_credit_mul is not modifying tc_cb_nr. That is every time tc_cb_nr will b equal to BTREE_CALLBACK_CREDIT irrespective of nr.

btree/btree.c
```diff
@@ -5549,6 +5558,7 @@ int64_t btree_create_tree_tick(struct m0_sm_op *smop)
 		 */
 		if (!addr_is_aligned(data->addr))
 			return M0_ERR(-EFAULT);
+#endif
```
ydb242, line 5561: Since this check has been disabled, the condition in UT basic_tree_op_icp also needs to be commented out. This condition is from lines 7553 to 7556.

btree/btree.c
```diff
@@ -5580,10 +5590,14 @@ int64_t btree_create_tree_tick(struct m0_sm_op *smop)
 
 		m0_rwlock_write_lock(&bop->bo_arbor->t_desc->t_lock);
 		bop->bo_arbor->t_desc->t_height = bop->bo_arbor->t_height;
+		_slot.s_node                    = oi->i_nop.no_node;
+		_slot.s_idx                     = 0;
+		node_capture(&_slot, bop->bo_tx);
```
ydb242, line 5595: If node_capture() is added here, then we need to add node_capture() in btree_create_tree_tick() as node_header is modified there also. The addition is to be done after line 5633.

RadhaGulhane13 (approved):

ydb242 (approved):

PR #956, opened 2021-08-16
Unload free extents[LRU] framework.

nihar1313:
LRU based unload in memory extents has been implemented.
It has been exposed as a callback per ad balloc free_entents.
This framework per AD can be called by thread or a RPC.


cortx-admin:

madhavemuri:
@mukundkanekar : Can you check if this PR is still needed ?

mukundkanekar:
This changes can remain in forked repo for now, needs additional perf analysis before merging to balloc branch.

PR #960, opened 2021-08-16
EOS-23582: update the format of support.yaml file

mehjoshi:
The new format of support.yaml is:
support_bundle:
    - "sh /opt/seagate/cortx/motr/libexec/m0reportbug-bundler"

### DIFF
--- a/scripts/install/opt/seagate/cortx/motr/conf/support.yaml
+++ b/scripts/install/opt/seagate/cortx/motr/conf/support.yaml
@@ -1,4 +1,2 @@
-
-bundle:
-    create:
-      url: "sh /opt/seagate/cortx/motr/libexec/m0reportbug-bundler"
+support_bundle:
+    - "sh /opt/seagate/cortx/motr/libexec/m0reportbug-bundler"


scripts/install/opt/seagate/cortx/motr/conf/support.yaml
```diff
@@ -1,4 +1,2 @@
-
-bundle:
-    create:
-      url: "sh /opt/seagate/cortx/motr/libexec/m0reportbug-bundler"
+support_bundle:
+    - "sh /opt/seagate/cortx/motr/libexec/m0reportbug-bundler"
```
yatin-mahajan, line 2: Do we have any dependency on CSM for this ? i.e. does this has corresponding changes in CSM as well. 
AFAIK this support.yaml is consumed by CSM only as of now.

huanghua78, line 2: Why is it an array?
Why not jus key:val ?

yeshpal-jain-seagate (approved):

scripts/install/opt/seagate/cortx/motr/conf/support.yaml
sachinpunadikar: @huanghua78 , this is not a final version of support.yaml.
This is stopgap arrangements, so that support bundle is getting generated. Currently YAML parsing of setup.yaml is failing due to changes made for "upgrade". Hence we are asking to have the same entry (like setup.yaml) in support.yaml
We will let you know once new format is ready for following.

sachinpunadikar: That will be taken care @yatin-mahajan

yatin-mahajan (approved):
Looks good to me.

rahul-tripathi-git (approved):
Looks good to me.

PR #986, opened 2021-08-22
Changes to m0_btree_create() to deermine node type from m0_btree_type data

shashank-parulekar:

### DIFF
--- a/btree/btree.h
+++ b/btree/btree.h
@@ -154,7 +154,7 @@ enum m0_btree_opflag {
  * Calculates the credit needed to create tree using root node and adds this
  * credit to @accum.
  */
-void m0_btree_create_credit(const struct node_type *nt,
+void m0_btree_create_credit(const struct m0_btree_type *bt,
 			    struct m0_be_tx_credit *accum);
 
 /**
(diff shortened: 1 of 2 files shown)


RadhaGulhane13 (approved):

btree/btree.h
```diff
@@ -154,7 +154,7 @@ enum m0_btree_opflag {
  * Calculates the credit needed to create tree using root node and adds this
  * credit to @accum.
  */
-void m0_btree_create_credit(const struct node_type *nt,
+void m0_btree_create_credit(const struct m0_btree_type *bt,
```
NikhilBirgade, line 157: What was the issue with passing node type as parameter?

shashank-parulekar: The node type is not exported to Motr since this type is internal to btree.c. On the same lines it is not needed for Motr to know what node type is used by the btree as long as it can tell the btree_create routine what is going to be the type of records that will be added to the btree. In this case the btree module is free to use whichever node types it wants to.

NikhilBirgade: If motr is only going to tell the record type and not node_type to btree_create routing, there is no need to pass node_type as parameter to btree_create.

shashank-parulekar: yes the changes are made there too. The only thing is in case I provided a way for btree_create to accept non-NULL nt parameter too. I will remove it.

NikhilBirgade: yes, we can remove passing nt` as parameter if we are sure that `nt` will not be visible to motr.

NikhilBirgade (approved):

ydb242 (approved):

PR #995, opened 2021-08-24
[DTM0] EOS-23886: Watcher for DTM0 log.

ivan-alekhin:
Problem: DTM0 recovery machine needs a "be_op-based" way of
triggering iteration over DTM0 log when the log gets "new" records.
Solution: Add a "one-shot" m0_be_op protected by a lock.

```
The watcher will help to address the following scenario:

C1|           M1(inprogress)
P2|   O  T    |   R(P1)
P1|   |  |    |   |     R(P1)  M1(executed)
  |   |  |    |   |     |      |
--------------------------------------------------------------->
      t1 t2   t3  t4    t5     t6                  realtime, t

Actors:
 C1 is a client (originator)
 P1, P2 are DTM0 processes (participants)

t1: P2 is ONLINE.
t2: P2 enters TRANSIENT state.
t3: Client sends request M1. It is sent only to P1 becaues
    P2 is in TRANSIENT state.
t4: P2 enters RECOVERING state.
t5: P1 learns about RECOVERING state of P2.
t6: M1 gets executed on P1. M1 is added to the log, and it
    has to be replayed to P2 even though it was executed
    (realtime-wise) after recovery of P2 started on P1.

It may happen that, at the moment t6, P2 replayed all
the previous (realtime-wise) records to P1, so that the corresponding
FOM fell asleep. The watcher helps to wake up the FOM, so that
it will continue replay of the log.
```

This patch is a dependency of EOS-23894 (remote recovery FOM).

### DIFF
--- a/be/dtm0_log.c
+++ b/be/dtm0_log.c
@@ -35,6 +35,7 @@
 #include "lib/trace.h"
 #include "dtm0/fop.h"  /* dtm0_req_fop */
 #include "motr/magic.h"
+#include "be/op.h"     /* m0_be_op_init and so on */
 
 
 M0_TL_DESCR_DEFINE(lrec, "DTM0 Log", static, struct m0_dtm0_log_rec,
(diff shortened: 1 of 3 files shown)


r-wambui:
@ivan-alekhin Are the changes on this PR still valid? If not can we please close it, if yes, we can request the reviewers to look at it and merge.

ivan-alekhin:
The patch is not relevant anymore because we are going to use the new log (and new approach where RECOVERING is not needed). So that, I'm closing the PR.
FYI: @madhavemuri , @andriytk

PR #970, opened 2021-08-17
lib: fix bogus warnings about thread migration

andriytk:
After commit 4c27187 the spurious warnings started appearing:

    [lib/locality.c:157:m0_locality_here]  thread migrated to another CPU core

The warnings were meant for the locality threads only, who are confined to a specific CPU cores. But they appear now for any thread created in Motr process.

Solution: add the warning code as a locality chore which runs once in 10 mins.

Also, set the initial value of tls_loci at m0_thread_trampoline() which is called from the new thread already, rather than at m0_thread_init() which is called from the parent-thread. As we use m0_processor_id_get() for the initial value, this seems to be more appropriate.

PR #998, opened 2021-08-26
Revert "lib: fix bogus warnings about thread migration"

mehjoshi:
Reverts Seagate/cortx-motr#970. Need to test this on a feature branch for its performance impact.


andriytk:
Where I can see the performance impact numbers, and how can one reproduce them?

PR #1006, opened 2021-08-26
WIP: An approach to fix EOS-14840.

gshipra:
1. Added a boolean variable available in struct m0_lockers
2. This variable will be set true in m0_lockers_init()
3. This variable will be set false in m0_lockers_fini()
4. This variable will be checked in m0_fom_locality_inc() and  fom_ready() and appropriate actions will be taken.


andriytk:
Please, describe the problem you are trying to fix by this patch in the description section above and the commit msg. I understand it might be described in the Jira ticket, but the Jira tickets are not open to the public.

PR #1010, opened 2021-08-27
EOS-24188: Enable m0crate to read existing populated data

yatin-mahajan:
Added one more op_code (CR_READ_ONLY) to m0crate utility, to allow m0crate to
read existing pupolated data with CR_POPULATE.

Added one more workload added to m0cratr_st.sh (m0crate_st_workloads.yaml),
to test CR_READ_ONLY opcode. This workload is placed after populate and
before cleanup. CR_POPULATE workload will populate the data -> CR_READ_ONLY
wolklaod will read the populated data and CR_CLEANUP wokload will cleanup
the data.

Signed-off-by: Yatin Mahajan <yatin.mahajan@seagate.com>


mehjoshi (approved):
Looks good to me.

motr/m0crate/crate_io.c
```diff
@@ -717,6 +717,12 @@ int cr_task_execute(struct m0_task_io *cti)
 			if (rc == 0)
 				cr_op_namei(cwi, cti, CR_DELETE);
 			break;
+		case CR_READ_ONLY:
+			rc = cr_op_namei(cwi, cti, CR_OPEN);
+			if (rc == 0) {
+				cr_op_io(cwi, cti, CR_READ);
+			}
+			break;
```
yeshpal-jain-seagate, line 725: What about the delete, dont we need delete operation here

yeshpal-jain-seagate: What if read_only was called without write

motr/st/utils/m0crate_st_workloads.yaml
```diff
@@ -35,7 +35,7 @@ WORKLOAD_SPEC:                   # Workload specification section
    WORKLOAD:                     # First Workload
       WORKLOAD_TYPE: 1           # Workload type, 0->Index, 1->IO
       WORKLOAD_SEED: tstamp      # SEED to the random number generator.
-      OPCODE: 2                  # CREATE/OPEN/WRITE/READ/DELETE/POPULATE/CLEANUP=(0,1,2,3,4,5,6).
+      OPCODE: 2                  # CREATE/OPEN/WRITE/READ/DELETE/POPULATE/CLEANUP/READ_ONLY=(0,1,2,3,4,5,6,7).
```
yeshpal-jain-seagate, line 38: Better to keep cleanup as a last parameter for more readability

yatin-mahajan: In enum m0_operations{ }, I added READ_ONLY as last parameter because adding in-between would change opcode of subsequent operations (CLEANUP). and that needs to be change other places as well. To avoid this I added it to last as similarly into workload.yaml file.

motr/m0crate/crate_io.c
yatin-mahajan: > What about the delete, dont we need delete operation here

This opcode is added so that we can read exiting populated data and thus can be done multiple times explicitly or  with NR_ROUND option in workload. To deleted this data CLEANUP opcode can be used.

yatin-mahajan: > What if read_only was called without write

I did not checked this, but it should return ENOENT error if the objects itself are not present. Will check this.

yeshpal-jain-seagate (approved):

motr/m0crate/crate_io.c
mehjoshi: @yatin-mahajan, Please update here if you have checked and verified.

PR #1018, opened 2021-09-01
Check for invalid tree open

ydb242:
This changes are meant to enhance the invalid case of UT where we cannot open a tree after it is destroyed.

### DIFF
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -2195,7 +2195,7 @@ static int64_t node_get(struct node_op *op, struct td *tree,
 	 * restart the tick funcions.
 	 */
 	if (!segaddr_header_isvalid(addr)) {
-		op->no_op.o_sm.sm_rc = M0_ERR(-ECHILD);
+		op->no_op.o_sm.sm_rc = M0_ERR(-EINVAL);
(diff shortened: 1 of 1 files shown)


btree/btree.c
```diff
@@ -9941,6 +9941,13 @@ static void btree_ut_tree_oper_thread_handler(struct btree_ut_thread_info *ti)
 		M0_ASSERT(rc == 0);
 		m0_be_tx_close_sync(tx);
 		m0_be_tx_fini(tx);
+
+		/** Error Case */
+		rc = M0_BTREE_OP_SYNC_WITH_RC(&b_op,
+					      m0_btree_open(rnode,
+							    1024, &tree,
+							    seg, &b_op));
+		M0_ASSERT(rc == -ECHILD);
```
shashank-parulekar, line 9950: In correct error code. Please change.

shashank-parulekar (changes requested):

btree/btree.c
ydb242: Done

NikhilBirgade (approved):

RadhaGulhane13 (approved):

shashank-parulekar (approved):

PR #1019, opened 2021-09-01
calling delete calllback before deleting record

RadhaGulhane13:

### DIFF
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -7084,10 +7084,7 @@ static int64_t btree_del_kv_tick(struct m0_sm_op *smop)
 		 * successful.
 		 */
 	case P_ACT: {
-		m0_bcount_t          ksize;
-		void                *p_key;
-		m0_bcount_t          vsize;
-		void                *p_val;
+
 		struct slot         node_slot;
 		bool                node_underflow;
 		int                 rc;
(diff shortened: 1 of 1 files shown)


btree/btree.c
```diff
@@ -7099,8 +7102,15 @@ static int64_t btree_del_kv_tick(struct m0_sm_op *smop)
...
 		lev = &oi->i_level[oi->i_used];
+
 		node_slot.s_node = lev->l_node;
 		node_slot.s_idx  = lev->l_idx;
+		node_slot.s_rec  = REC_INIT(&p_key, &ksize, &p_val, &vsize);
+
+		node_rec(&node_slot);
+		node_slot.s_rec.r_flags = M0_BSC_SUCCESS;
+		rc = bop->bo_cb.c_act(&bop->bo_cb, &node_slot.s_rec);
+		M0_ASSERT(rc == 0);
```
shashank-parulekar, line 7113: This should not assert. Instead if the return value is non-0 then abort the delete operation and return back to the caller with the same error code as is added in rc by the callback routine.

btree/btree.c
```diff
@@ -7099,8 +7102,15 @@ static int64_t btree_del_kv_tick(struct m0_sm_op *smop)
...
 
 		lev = &oi->i_level[oi->i_used];
+
 		node_slot.s_node = lev->l_node;
 		node_slot.s_idx  = lev->l_idx;
+		node_slot.s_rec  = REC_INIT(&p_key, &ksize, &p_val, &vsize);
+
+		node_rec(&node_slot);
+		node_slot.s_rec.r_flags = M0_BSC_SUCCESS;
+		rc = bop->bo_cb.c_act(&bop->bo_cb, &node_slot.s_rec);
```
shashank-parulekar, line 7112: The callback needs to be called only if (bop->bo_cb.c_act != NULL)

shashank-parulekar (changes requested):

btree/btree.c

btree/btree.c
RadhaGulhane13: okay , updated.

ydb242 (approved):

NikhilBirgade (approved):

PR #1020, opened 2021-09-01
Added alternate credit calculator functions to be used in absence of btree.

shashank-parulekar:
Added functions m0_btree_put_credit2(), m0_btree_update_credit2() and
m0_btree_del_credit2(). These alternate functions are to be used when the btree
pointer is not available. The new alternate functions should never be used to
replace the original credit functions for performance reasons.

### DIFF
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -1183,6 +1183,12 @@ struct nd {
 	struct node_op         *n_op;
 
 	/**
+	 *  Size of the node on BE segment pointed by this nd. Added here for
+	 *  easy reference.
+	 */
+	uint32_t                n_size;
+
+	/**
 	 * flag for indicating if node on BE segment is valid or not, but it
 	 * does not indicated anything about node descriptor validity.
 	 */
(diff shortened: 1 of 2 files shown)


NikhilBirgade (approved):

ydb242 (approved):

btree/btree.c
```diff
@@ -4509,10 +4514,30 @@ void m0_btree_put_credit(const struct m0_btree  *tree,
 	m0_be_tx_credit_mac(accum, &cred, nr);
 }
 
-/**
- * This function will calculate credits required to perform @nr delete KV
- * operations and it will add those credits to @accum.
- */
+void m0_btree_put_credit2(const struct m0_btree_type *type,
```
RadhaGulhane13, line 4517: In current credit calculation, we are using MAX_HEIGHT, but btree user may provide us exact height so we can optimize credit calculations.   In old btree,  there are functions m0_be_btree_insert_credit2, m0_be_btree_delete_credit2, etc. which
used btree height provided by user, We can do the similar thing, as if you are creating dummy tree that means height is not known so we can set height to MAX_height, and call this function if btree pointer is not provided to us.

shashank-parulekar: @RadhaGulhane13  - Our current btree credit calculation functions do not take the parameter to use either MAX_HEIGHT or user provided height. If we ever start to support this then at that time we should change the _credit2() functions - in other words that will be a different task.

RadhaGulhane13: okay.

PR #1023, opened 2021-09-02
Rectify tree ops

ydb242:
Modify tree_create() and tree_open() to accept tree from the user and assigning it to bop.bo_arbor instead of allocating memory for bop.bo_arbor in tree_create()

### DIFF
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -5846,11 +5846,6 @@ int64_t btree_create_tree_tick(struct m0_sm_op *smop)
 		if (oi == NULL)
 			return M0_ERR(-ENOMEM);
 		bop->bo_i = oi;
-		bop->bo_arbor = m0_alloc(sizeof *bop->bo_arbor);
-		if (bop->bo_arbor == NULL) {
-			m0_free(oi);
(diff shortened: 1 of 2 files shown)


btree/btree.c
```diff
@@ -7698,7 +7686,7 @@ static void ut_basic_tree_oper_cp(void)
 	rc = M0_BTREE_OP_SYNC_WITH_RC(&b_op, m0_btree_destroy(b_op.bo_arbor,
 							      &b_op, tx));
 	M0_ASSERT(rc == 0);
-	m0_free(btree);
+	M0_SET0(&btree);
```
shashank-parulekar, line 7689: This call should be moved above to line # 7682. Please make similar change to call the SET0 after every close.

shashank-parulekar (changes requested):

btree/btree.c
ydb242: Updated.

shashank-parulekar (approved):

NikhilBirgade (approved):

RadhaGulhane13 (approved):

PR #1024, opened 2021-09-02
Integrating New Btree with Conf database of Motr.

shashank-parulekar:
This commit captures the changes which will make the conf database use new
BTree.

### DIFF
--- a/conf/db.c
+++ b/conf/db.c
@@ -412,8 +412,9 @@ M0_INTERNAL int m0_confdb_create(struct m0_be_seg      *seg,
 		put_cb.c_act       = confd_btree_kv_put_cb;
 		put_cb.c_datum     = &rec;
 		rc = M0_BTREE_OP_SYNC_WITH_RC(&kv_op,
-					      m0_btree_put(&btree, &rec, &put_cb,
-							   &kv_op, tx));
+					      m0_btree_put(&btree, &rec,
+							   &put_cb, &kv_op,
+							   tx));
 		if (rc != 0) {
 			/** Delete all objects added to the btree. */
 			while (i--) {
@@ -424,11 +425,11 @@ M0_INTERNAL int m0_confdb_create(struct m0_be_seg      *seg,
 				k_ptr = (struct m0_fid *)
 					     m0_conf_objx_fid(obj_ctx.oc_obj);
 				M0_BTREE_OP_SYNC_WITH_RC(&kv_op,
-							 m0_btree_del(&btree,
-								      &rec.r_key,
-								      NULL,
-								      &kv_op,
-								      tx));
+							m0_btree_del(&btree,
+								     &rec.r_key,
+								     NULL,
+								     &kv_op,
+								     tx));
 			}
 		}
 	}
--- a/conf/db.c
+++ b/conf/db.c
@@ -1,6 +1,6 @@
 /* -*- C -*- */
 /*
- * Copyright (c) 2013-2020 Seagate Technology LLC and/or its Affiliates
+ * Copyright (c) 2013-2021 Seagate Technology LLC and/or its Affiliates
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -29,14 +29,14 @@
 #include "conf/onwire_xc.h"
 #include "conf/obj.h"
 #include "xcode/xcode.h"
-#include "be/btree.h"
+#include "btree/btree.h"
 #include "be/tx.h"
 #include "be/op.h"
 #include "lib/memory.h"      /* m0_alloc, m0_free */
 #include "lib/errno.h"       /* EINVAL */
 #include "lib/misc.h"        /* M0_SET0 */
 
-static int confdb_objs_count(struct m0_be_btree *btree, size_t *result);
+static int confdb_objs_count(struct m0_btree *btree, size_t *result);
 static void confdb_table_fini(struct m0_be_seg *seg);
 
 struct __pack {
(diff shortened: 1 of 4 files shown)


conf/db.c
```diff
@@ -142,63 +142,55 @@ struct confx_ctx {
...
-
-	rc = m0_be_seg_dict_insert(seg, tx, btree_name, *btree);
-	if (rc == 0 && M0_FI_ENABLED("ut_confdb_create_failure"))
-		rc = -EINVAL;
-	if (rc != 0) {
-		confdb_table_fini(seg);
-		m0_confdb_destroy(seg, tx);
+	M0_ASSERT(rnode_sz > 0 && m0_is_po2(rnode_sz));
+	rnode_sz_shift = __builtin_ffsl(rnode_sz) - 1;
+	M0_BE_ALLOC_ALIGN_ARR_SYNC(rnode, rnode_sz, rnode_sz_shift, seg, tx);
```
NikhilBirgade, line 167: Add an assert to check rnode !=NULL.

conf/db.c
```diff
@@ -142,63 +142,55 @@ struct confx_ctx {
...
+							     seg, btree_fid,
+							     tx));
+	if (rc != 0)
+		M0_BE_FREE_ALIGN_ARR_SYNC(rnode, seg, tx);
+	else {
+		rc = m0_be_seg_dict_insert(seg, tx, btree_name, rnode);
+		if (rc == 0 && M0_FI_ENABLED("ut_confdb_create_failure"))
+			rc = -EINVAL;
+		if (rc != 0) {
+			M0_BTREE_OP_SYNC_WITH_RC(&b_op,
```
NikhilBirgade, line 183: Add check for return code.

conf/db.c
```diff
@@ -142,63 +142,55 @@ struct confx_ctx {
...
+						 m0_btree_close(btree, &b_op));
+			m0_confdb_destroy(seg, tx);
+		}
 	}
 
 	return M0_RC(rc);
 }
 
 static void confdb_table_fini(struct m0_be_seg *seg)
```
NikhilBirgade, line 192: Empty function, it can be removed.

conf/db.c
```diff
@@ -328,15 +333,44 @@ static int confx_allocator_init(struct confx_allocator *alloc,
...
+{
+	struct m0_btree_rec     *put_data = cb->c_datum;
+	struct m0_bufvec_cursor  scur;
+	struct m0_bufvec_cursor  dcur;
+	m0_bcount_t              ksize;
+	m0_bcount_t              vsize;
+
+	ksize = m0_vec_count(&put_data->r_key.k_data.ov_vec);
+	vsize = m0_vec_count(&put_data->r_val.ov_vec);
+	if (ksize > m0_vec_count(&rec->r_key.k_data.ov_vec) ||
```
NikhilBirgade, line 347: As per btree type, it is fixed key variable value size type btree. Is there any case where `ksize` can be greater than `m0_vec_count(&rec->r_key.k_data.ov_vec)` ?

conf/db.c
```diff
@@ -362,15 +399,41 @@ M0_INTERNAL int m0_confdb_create(struct m0_be_seg      *seg,
...
+				M0_BTREE_OP_SYNC_WITH_RC(&kv_op,
+							 m0_btree_del(&btree,
+								      &rec.r_key,
+								      NULL,
+								      &kv_op,
+								      tx));
+			}
+		}
 	}
+	M0_BTREE_OP_SYNC_WITH_RC(&b_op, m0_btree_close(&btree, &b_op));
```
NikhilBirgade, line 435: shouldn't this be under `rc !=0`

conf/db.c
```diff
@@ -380,43 +443,53 @@ M0_INTERNAL int m0_confdb_create(struct m0_be_seg      *seg,
...
+				     m0_btree_open(rnode, rnode_sz,
+						   &btree, seg, &b_op)) == 0) {
+		m0_btree_destroy_credit(&btree, accum);
+	} else
+		/** Use the same credit count as btree_create. */
+		m0_btree_create_credit(&bt, accum);
 	m0_be_seg_dict_delete_credit(seg, btree_name, accum);
-	M0_BE_FREE_CREDIT_PTR(&btree, seg, accum);
+
+	M0_ASSERT(rnode_sz > 0 && m0_is_po2(rnode_sz));
```
NikhilBirgade, line 469: It will be better to have this assert at start of function as m0_btree_open is also using these variables.

conf/db.c
shashank-parulekar: This is done on purpose since because we need to return the previous rc value to the caller. If have a failure for closing the tree that is secondary.

conf/db.c
shashank-parulekar: I did not want to make changes in the calling functions and hence kept these empty functions.

conf/db.c
shashank-parulekar: While your assumption is correct I am just trying to make sure that the User is not trying to put lengthier key because of a bug in the code. If we allow this then it will corrupt our tree, hence we are returning an error instead of ASSERTing it.

conf/db.c
shashank-parulekar: No this tree needs to be closed because the btree pointer is not available to the callers of this routine. Other routines which need to operate on this btree will need to re-open the btree.

conf/db.c
```diff
@@ -362,15 +400,41 @@ M0_INTERNAL int m0_confdb_create(struct m0_be_seg      *seg,
...
+							 m0_btree_del(&btree,
+								      &rec.r_key,
+								      NULL,
+								      &kv_op,
+								      tx));
+			}
+		}
 	}
+	M0_BTREE_OP_SYNC_WITH_RC(&b_op, m0_btree_close(&btree, &b_op));
 	if (rc != 0) {
```
NikhilBirgade, line 437: `{}` can be removed.

shashank-parulekar: done

NikhilBirgade (approved):

conf/db.c
```diff
@@ -362,15 +399,41 @@ M0_INTERNAL int m0_confdb_create(struct m0_be_seg      *seg,
...
+		v_ptr = &obj_ctx;
+		vsize = m0_confx_sizeof() + sizeof(struct confx_allocator);
+
+		rec.r_key.k_data   = M0_BUFVEC_INIT_BUF(&k_ptr, &ksize);
+		rec.r_val          = M0_BUFVEC_INIT_BUF(&v_ptr, &vsize);
+
+		put_cb.c_act       = confd_btree_kv_put_cb;
+		put_cb.c_datum     = &rec;
+		rc = M0_BTREE_OP_SYNC_WITH_RC(&kv_op,
+					      m0_btree_put(&btree, &rec, &put_cb,
```
RadhaGulhane13, line 415: exceed 80 char limit

conf/db.c
```diff
@@ -362,15 +399,41 @@ M0_INTERNAL int m0_confdb_create(struct m0_be_seg      *seg,
...
+			while (i--) {
+				if (confx_obj_dup(&alloc, &obj_ctx.oc_obj,
+						  M0_CONFX_AT(conf, i)) != 0)
+					continue;
+
+				k_ptr = (struct m0_fid *)
+					     m0_conf_objx_fid(obj_ctx.oc_obj);
+				M0_BTREE_OP_SYNC_WITH_RC(&kv_op,
+							 m0_btree_del(&btree,
+								      &rec.r_key,
```
RadhaGulhane13, line 428: exceed 80 char limit

RadhaGulhane13 (approved):

RadhaGulhane13 (approved):

PR #1027, opened 2021-09-05
EOS-23641: CRC support for fixed key variable value size node format.

NikhilBirgade:

### DIFF
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -576,13 +576,6 @@
  *  --------------------------------------------
  */
 
-struct td;
-struct m0_btree {
-	const struct m0_btree_type *t_type;
-	unsigned                    t_height;
-	struct td                  *t_desc;
-};
-
 enum base_phase {
 	P_INIT = M0_SOS_INIT,
 	P_DONE = M0_SOS_DONE,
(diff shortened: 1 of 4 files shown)


btree/btree.c
```diff
@@ -3112,6 +3112,46 @@ static void ff_rec_del_credit(const struct nd *node, m0_bcount_t ksize,
...
+ *
+ * In btree context, CRC will be used to verify the btree-node data read from
+ * storage. It can be used to verify keys and/or values stored in the node. An
+ * additional CRC(known as CRC of CRCs) will be computed from all the
+ * individual CRCs and stored in the node header.
+ *
+ * There are two proposal for supporting CRCs in the node:
+ * 1. Discrete CRC : Calculate CRC for each key and/or value and store it after
+ * the respective key in the node. In case CRC is provided by
+ * user as part of key and/or value, there is no need of calculating CRCs. The
```
RadhaGulhane13, line 3126: on't we want to calculate CRC irrespective of user-provided CRC. As if we do downgrade or upgrade, the user provided CRC length may get differ, so we may not be able to calculate cCRC accurately.

PR #1039, opened 2021-09-13
lib: fix panic at m0_locality_chores_run() on NUMA nodes

andriytk:
On NUMA nodes, with numad service enabled, CPU affinity of
locality threads can be changed (especially, for compute-
intensive applications which use a lot of CPU and/or memory).
As a result, this leads to the following panic:

    Motr panic: (locality == m0_locality_here()) at m0_locality_chores_run() lib/locality.c:310 (errno: 0) (last failed: none) [git: sage-base-1.0-389-g09ff618]

    #0  0x00002b25fbcb6387 in raise () from /usr/lib64/libc.so.6
    #1  0x00002b25fbcb7a78 in abort () from /usr/lib64/libc.so.6
    #2  0x00002b25fc5e7acd in m0_arch_panic (c=c@entry=0x2b25fca74400 <__pctx.15545>, ap=ap@entry=0x2b26032c7a18) at lib/user_space/uassert.c:131
    #3  0x00002b25fc5d7e44 in m0_panic (ctx=ctx@entry=0x2b25fca74400 <__pctx.15545>) at lib/assert.c:52
    #4  0x00002b25fc5dbf6a in m0_locality_chores_run (locality=locality@entry=0x106f778) at lib/locality.c:310
    #5  0x00002b25fc5abcaa in loc_handler_thread (th=0xf5d160) at fop/fom.c:926
    #6  0x00002b25fc5de2ee in m0_thread_trampoline (arg=arg@entry=0xf5d168) at lib/thread.c:117
    #7  0x00002b25fc5e882d in uthread_trampoline (arg=0xf5d168) at lib/user_space/uthread.c:98
    #8  0x00002b25fba6bea5 in start_thread () from /usr/lib64/libpthread.so.0
    #9  0x00002b25fbd7e96d in clone () from /usr/lib64/libc.so.6

The reason is that we link our localities to the CPUs in the
very beginning, when the application is just started and Motr
is initialised. And we use the CPU id to figure out the current
locality at m0_locality_here(). Of course, when the affinity
is changed later and the thread is moved to another CPU, this
gives the wrong locality result.

Solution: stash the initial CPU id in TLS and use it at
m0_locality_here().

Kudos to Nikita Danilov for the idea.

Reviewed-by: Nikita Danilov <nikita.danilov@seagate.com>
Reviewed-by: Huang Hua <hua.huang@seagate.com>
Signed-off-by: Andriy Tkachuk <andriy.tkachuk@seagate.com>

### DIFF
--- a/fop/fom.c
+++ b/fop/fom.c
@@ -135,6 +135,7 @@
 
 enum {
 	LOC_IDLE_NR = 1,
+	LOCM_CHECK_PERIOD_SEC = 10*60,
 	HUNG_FOP_SEC_PERIOD   = 5,
 	HUNG_FOP_TIME_SEC_MAX = 2*60,
 	HUNG_FOP_TIME_SEC_IEM = 5*60,
@@ -197,6 +198,8 @@ static int loc_thr_create(struct m0_fom_locality *loc);
 
 static void hung_foms_notify(struct m0_locality_chore *chore,
 			     struct m0_locality *loc, void *place);
+static void locm_warn_notify(struct m0_locality_chore *chore,
+			     struct m0_locality *loc, void *place);
 
 static struct m0_sm_conf fom_states_conf0;
 M0_INTERNAL struct m0_sm_conf fom_states_conf;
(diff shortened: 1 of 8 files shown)


mehjoshi:
@andriytk, Is this the same patch that was checked in earlier or are there some changes in it? The earlier patch showed regressions during performance testing.

andriytk:
@mehjoshi, yes - it's the same patch. It was agreed to re-test it in order to investigate the performance issues.

cortx-admin:

PR #1045, opened 2021-09-15
go/mio: update after m0_obj_op() API change

andriytk:
Motr API was changed at commit 3a429ad57: attr[] parameter to
m0_obj_op() has to be NULL now if there are no checksums in it.

### DIFF
--- a/bindings/go/mio/mio.go
+++ b/bindings/go/mio/mio.go
@@ -114,7 +114,6 @@ type slot struct {
 type iov struct {
     buf     []C.struct_m0_bufvec
     ext     []C.struct_m0_indexvec
-    attr    []C.struct_m0_bufvec
     minBuf  []byte
     ch      chan slot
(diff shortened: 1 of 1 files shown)


madhavemuri:
@yeshpal-jain-seagate : Can you check if a flag can be added instead of checking for attribute to be NULL, so that old API compatibility can be maintained ?

nikitadanilov (approved):

mehjoshi:
@yeshpal-jain-seagate , @andriytk , Is an alternative approach, as suggested by @madhavemuri, using a flag instead of NULL values being tried?

andriytk:
@mehjoshi I hope so. Ask @truptiatseagate or @madhavemuri .

andriytk:
This patch is not needed anymore, the compatibility issue has been fixed by #1345.

PR #1050, opened 2021-09-17
Create pull_request_template.md

mehjoshi:
PR for adding a pull_request_template.md

-----
[View rendered .github/pull_request_template.md](https://github.com/Seagate/cortx-motr/blob/mehjoshi-pull-request-template/.github/pull_request_template.md)

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,32 @@
+# Problem Statement
+- Problem statement
+
+# Design
+-  For Bug, Describe the fix here.
+-  For Feature, Post the link for design
+
+# Coding
+-  [ ] Coding conventions are followed and code is consistent
+
+# Testing 
+- [ ] Unit and System Tests are added
+- [ ] Test Cases cover Happy Path, Non-Happy Path and Scalability
+- [ ] Testing was performed with RPM
+
+# Review Checklist 
+- [ ] PR is self reviewed
+- [ ] JIRA number/GitHub Issue added to PR
+- [ ] Jira and state/status is updated and JIRA is updated with PR link
+- [ ] Check if the description is clear and explained
+- [ ] Is there a change in filename/package/module or signature? [Y/N]: 
+- [ ] If yes for above point, is a notification sent to all other cortx components? [Y/N]
+- [ ] Side effects on other features (deployment/upgrade)? [Y/N]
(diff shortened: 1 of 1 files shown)


ivan-alekhin:
_Disclaimer: please take this message with a gain of salt. Time-to-time I look through patches across cortx projects, but I do not know about ongoing problems that the motr gatekeepers face everyday. So that, I may be wrong/out-of-the-loop._

A PR template is a good first step for a project to formalize its review process. The next step is a separate document that describes code review process.
Motr has a quite long history already, so I wonder if it would be better to write down a document that describes our code review process instead of adding a PR template. It is nearly impossible to create a PR template that would be good for every kind of situation (sometimes patches do not require testing, do not have Jira tickets, do not have HLD/DLDs, and so on). But it is possible to describe code review process: "do this and that", "run perfline using this link", "add this person to the reviewers", "here is an example of well-written PR description", and so on. You, as a gatekeeper, could point to this document anytime when someone does not follow the guidelines (in the same way as we are doing it with Motr coding style).

From this perspective, this PR template would have only one checkbox: "[] I have read Motr code review process document[link]".

<details>
  <summary>Btw, we already have examples where a well defined process works better than templates. </summary>
  
Take a look at hare (https://github.com/Seagate/cortx-hare/commit/d73c934257089eb7afe9e6f4d0de1c66f7809bd0) and utils (https://github.com/Seagate/cortx-utils/commit/2e6f46a909fda2d96166e677396467d742461139). Hare has no PR templates but they have rigorous reviewers/gatekeepers and they strictly follow the process and the corresponding documents. Because of that, Hare commit history looks great (it is readable and usable). I cannot tell the same about utils: the project has templates but its commit history does not look as good as Hare's history.
</details>

mehjoshi:
Thank you, Ivan. I whole heartedly agree with what you suggest here.
Yes, we do have a PR review process and it is documented here:
https://seagate-systems.atlassian.net/wiki/spaces/PRIVATECOR/pages/512033231/Guidelines+for+submitting+feature+patches+to+Motr

I will see how I can improve it using the suggestions you provided. Thank you for that and keep them coming :thumbsup:

truptiatseagate (approved):

mukul-seagate11 (approved):
looks good

PR #1052, opened 2021-09-17
Dtm0 main dtm recovered

just-now:
EOS-24483: Added CONF_HA_RECOVERED() message and corresponding HA states
     - Works in conjunction with HARE related patch. Reference: EOS-24483


cortx-admin:

PR #1053, opened 2021-09-17
mio: drop duplicate CFLAGS from mkv.go

andriytk:
Drop duplicate CFLAGS settings from mio/mkv.go - they are
already set at mio/mio.go. Also, give priority to the local
headers from the sources tree first over the installed ones,
this allows to build the utilities from the more recent
sources tree than the installed motr packages.

### DIFF
--- a/bindings/go/mio/mio.go
+++ b/bindings/go/mio/mio.go
@@ -39,8 +39,8 @@
 // For the usage example, refer to mcp utility.
 package mio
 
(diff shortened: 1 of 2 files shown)


andriytk:
cc @chumakd

sanjognaik (approved):
Looks good.

siningwuseagate (approved):
looks fine.

PR #1054, opened 2021-09-20
EOS-24521: [POC} create SEG0 and SEG1 on the same device

imvenkip:
Analysis:
Presently Seg0 is created on the file system and Seg1 is created on the
device. LC will not support Seg0 on the file system. So create Seg0 and
Seg1 on the same device.

Solution:
Created Seg0 and Seg1 on the same device.

Co-authored-by: Vidyadhar Pinglikar <Vidyadhar.Pinglikar@seagate.com>
Signed-off-by: Venkateswarlu Payidimarry <venkateswarlu.payidimarry@seagate.com>

### DIFF
--- a/be/seg.c
+++ b/be/seg.c
@@ -225,17 +225,23 @@ M0_INTERNAL int m0_be_seg_create(struct m0_be_seg *seg,
 				 m0_bcount_t size,
 				 void *addr)
 {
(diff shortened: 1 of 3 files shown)


imvenkip:
It is relevant to PoC code.

PR #1071, opened 2021-09-23
Clean up of btree ut's

shashank-parulekar:
Removed ut's which are not used.
Fixed ut basic_tree_op_icp which used to crash.
Added code to check for node validity when it is on BE segment.

Signed-off-by: Shashank Parulekar <Shashank.Parulekar@seagate.com>

# Problem Statement
- An issue is seen during initialization of balloc trees. Some incorrect data is
- seen on the root node of the btree when it is on BE segment.

# Design
-  An enhancement in the ut to check for node validity when the nodes
- are offline.

# Coding
-  [Yes ] Coding conventions are followed and code is consistent

# Testing 
- [Yes ] Unit and System Tests are added
- [Yes ] Test Cases cover Happy Path, Non-Happy Path and Scalability
- [No ] Testing was performed with RPM

### DIFF
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -10791,29 +10791,31 @@ static void ut_mt_tree_oper(void)
 static bool validate_nodes_on_be_segment(struct segaddr *rnode_segaddr)
 {
 	const struct node_type     *nt;
-	struct node_data {
-		struct segaddr      node_segaddr;
-		uint16_t            rec_idx;
+	struct {
+		struct segaddr node_segaddr;
+		uint16_t       rec_idx;
 		}                   stack[MAX_TREE_HEIGHT + 1];
 	uint32_t                    stack_level = 0;
-	struct nd                   _node;
-	struct slot                 _slot = { .s_node = &_node };
+	struct nd                   n;
+	struct slot                 s       = { .s_node = &n };
 	uint16_t                    rec_idx = 0;
 
 	nt = btree_node_format[segaddr_ntype_get(rnode_segaddr)];
-	_node.n_addr = *rnode_segaddr;
+	n.n_addr = *rnode_segaddr;
 
 	while (true) {
-		if (nt->nt_level(&_node) > 0 &&
-		    rec_idx < nt->nt_count_rec(&_node)) {
-			stack[stack_level] = (struct node_data) {
-						.node_segaddr = _node.n_addr,
-						.rec_idx      = rec_idx,
-						};
+		/**
+		 * Move down towards leaf node only if we have child nodes still
+		 * to traverse.
+		 */
+		if (nt->nt_level(&n) > 0 &&
+		    rec_idx < nt->nt_count_rec(&n)) {
+			stack[stack_level].node_segaddr = n.n_addr,
+			stack[stack_level].rec_idx      = rec_idx,
 			stack_level++;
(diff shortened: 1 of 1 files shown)


btree/btree.c
```diff
@@ -8445,199 +8445,6 @@ static void btree_ut_fini(void)
 	btree_ut_initialised = false;
 }
 
-#if 0
-static void get_rec_at_index(struct nd *node, int idx, uint64_t *key,
```
NikhilBirgade, line 8449: `get_rec_at_index and get_key_at_index` are called from `ut_traversal()`.  Currently ut_traversal () is disabled but it is useful for debuging. Re-include these functions.

btree/btree.c
```diff
@@ -11091,6 +10788,57 @@ static void ut_mt_tree_oper(void)
 	btree_ut_num_threads_tree_oper(0);
 }
 
+static bool validate_nodes_on_be_segment(struct segaddr *rnode_segaddr)
+{
+	const struct node_type     *nt;
+	struct node_data {
```
NikhilBirgade, line 10794: Declare structure outside of function body and use structure array in the function.

btree/btree.c
```diff
@@ -11091,6 +10788,57 @@ static void ut_mt_tree_oper(void)
...
+		struct segaddr      node_segaddr;
+		uint16_t            rec_idx;
+		}                   stack[MAX_TREE_HEIGHT + 1];
+	uint32_t                    stack_level = 0;
+	struct nd                   _node;
+	struct slot                 _slot = { .s_node = &_node };
```
NikhilBirgade, line 10800: Please use other name for _node and _slot, which do not start from _ .

btree/btree.c
```diff
@@ -11091,6 +10788,57 @@ static void ut_mt_tree_oper(void)
...
+		if (nt->nt_level(&_node) > 0 &&
+		    rec_idx < nt->nt_count_rec(&_node)) {
+			stack[stack_level] = (struct node_data) {
+						.node_segaddr = _node.n_addr,
+						.rec_idx      = rec_idx,
+						};
+			stack_level++;
+			_slot.s_idx = rec_idx;
+			nt->nt_child(&_slot, &_node.n_addr);
+			_node.n_addr.as_core = (uint64_t)segaddr_addr(&_node.n_addr);
```
NikhilBirgade, line 10816: 80 char limit.

RadhaGulhane13 (approved):

btree/btree.c
shashank-parulekar: I do not see anyone invoking ut_traversal. If this function is called from some other ut then we need to have had the call at least commented in the code. I will have a look at the traversal function and see if we need it. If we need it then we need to make a call to it from some function.

btree/btree.c
shashank-parulekar: This has been done in Motr code (especially in ut's) so that the structure is a part of only this function and is not used anywhere else.

btree/btree.c
shashank-parulekar: done.

btree/btree.c
shashank-parulekar: 👍

NikhilBirgade (approved):

PR #1116, opened 2021-10-06
Btree truncate implementation changes.

NikhilBirgade:

### DIFF
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -6670,7 +6670,7 @@ static struct m0_sm_state_descr btree_states[P_NR] = {
 	[P_ACT] = {
 		.sd_flags   = 0,
 		.sd_name    = "P_ACT",
-		.sd_allowed = M0_BITS(P_CAPTURE, P_CLEANUP, P_DOWN, P_DONE),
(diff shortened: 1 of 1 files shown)


btree/btree.c
```diff
@@ -8229,132 +8229,69 @@ static int64_t btree_truncate_tick(struct m0_sm_op *smop)
...
 		 * Case: Tree has more than one level.
-		 * Tree traversal hase reached the leaf node.Take parent node of
-		 * leaf and delete all its child nodes.
-		 * Before deleting all the nodes in parent, delete the last
-		 * accessed node as we do not want to call node_get again for
-		 * this node.
+		 * Tree traversal has reached the leaf node. Delete the leaf
+		 * node.
 		 */
+		bnode_capture(&node_slot, bop->bo_tx);
```
shashank-parulekar, line 8261: Call bnode_capture() after calling bnode_fini() on the next line so that the header magic change is also captured.

shashank-parulekar (changes requested):

btree/btree.c
NikhilBirgade: Oh yes. Thanks.

PR #1121, opened 2021-10-07
Update QSG with Arm64 platform info

andriytk:
Update Quick Start Guide with Arm64 platform information.

### DIFF
--- a/cas/service.c
+++ b/cas/service.c
@@ -1140,10 +1140,19 @@ static int op_sync_wait(struct m0_fom *fom)
 	struct m0_be_tx     *tx;
 
 	tx = m0_fom_tx(fom);
-	if (m0_be_tx_state(tx) < M0_BTS_LOGGED) {
-		M0_LOG(M0_DEBUG, "fom wait for tx to be logged");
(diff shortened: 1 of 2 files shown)


doc/Quick-Start-Guide.rst
```diff
@@ -8,7 +8,7 @@ Prerequisites
 *************
 The prerequisite that is necessary to install the Motr component is mentioned below.
 
-- CentOS-7 for x86_64 platform (ARM64 platform support work is in progress).
+- CentOS 7 or 8 on x86_64 or ARM64 (AArch64) platform.
```
huanghua78, line 11: I know that CORTX-Motr may run on CentOS 8. But this is not officially fully tested/verified, and not announced publicly.

andriytk: Well, let's announce it this silent way, then. :)

It seems, the important question here is whether we will support and fix the reported bugs on CentOS 8. And I think the answer is - yes. Isn't it?

Let me know your opinion. If you don't agree - I will drop it.

andriytk: Ok, dropped for now.

huanghua78 (approved):
Looks good.

PR #1125, opened 2021-10-08
EOS-24933: s3server service failing intermittently while executing s3 tests.

gshipra:

# Problem Statement
s3server service failing intermittently while executing s3 tests.

# Design
1.There was an assumption taken in code that m0_ivec_cursor_move returns false after reaching end of cursor. This assumption is wrong and generated defect.  Changed the code as per correct return value of  m0_ivec_cursor_move. 
2. rep_ivec may contain very less data and it may not be eligible to fetch checksums. In this case simply return, do not assert.
3. Use the correct unit size to move the cursors.

# Testing 
 Tested the fix for different IO sizes on 3 node cluster setup.
Sanity tests performed.

### DIFF
--- a/motr/io_req_fop.c
+++ b/motr/io_req_fop.c
@@ -136,7 +136,9 @@ static void application_attribute_copy(struct m0_indexvec *rep_ivec,
 				       struct m0_op_io *ioo,
 				       struct m0_buf *buf)
 {
-	uint32_t                unit_size, off, cs_sz;
+	uint32_t                unit_size;
+	uint32_t                off;
+	uint32_t                cs_sz;
 	m0_bindex_t             rep_index;
 	m0_bindex_t             ti_cob_index;
 	m0_bindex_t             ti_goff_index;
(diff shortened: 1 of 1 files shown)


gshipra:
> @gshipra , please also check the m0t1fs/linux_kernel implementation. Maybe it also needs a fix.

@huanghua78 could you please elaborate on this. What fix does m0t1fs/linux_kernel needs. This is a new function which is added.

gshipra:
> This fix and the earlier DI code that it fixes are violating motr coding style. Please fix this and re-submit. Adhering to the coding style is important for code readability.

@nikitadanilov - I  joined in motr some months back, I would be more careful about formatting in the future.

motr/io_req_fop.c
```diff
@@ -163,28 +165,32 @@ static void application_attribute_copy(struct m0_indexvec *rep_ivec,
...
 	off = rep_index % unit_size;
 	if (off) {
-		if (m0_ivec_cursor_move(&rep_cursor, unit_size - off)) {
+		if (!m0_ivec_cursor_move(&rep_cursor, unit_size - off))
 			rep_index = m0_ivec_cursor_index(&rep_cursor);
+		else
+			return;
+	}
+	off = ti_cob_index % unit_size;
+	if (off) {
```
huanghua78, line 181: coding style issue.

```
137   * use difference between NULL, 0 and "false" to emphasize whether
138     an expression is used as a pointer, integer (including success
139     or failure code) or boolean:
140
141           if (p == NULL) { /* assumes that p is a pointer */
142           } else if (q != 0) { /* q is an integer */
143           } else if (r) { /* r is a boolean */
144           }
145

```

Please write it as:
if (off != 0) {

yeshpal-jain-seagate: Fixed

motr/io_req_fop.c
```diff
@@ -197,31 +203,33 @@ static void application_attribute_copy(struct m0_indexvec *rep_ivec,
...
+			  (ioo->ioo_ext.iv_index[ioo->ioo_ext.iv_vec.v_nr-1] +
+			  ioo->ioo_ext.iv_vec.v_count[ioo->ioo_ext.iv_vec.v_nr-1]));
 
 		dst = m0_extent_vec_get_checksum_addr(&ioo->ioo_attr,
 						      ti_goff_index,
 						      &ioo->ioo_ext,
-						      unit_size, cs_sz );
-
+						      unit_size, cs_sz);
 		memcpy(dst, src, cs_sz);
```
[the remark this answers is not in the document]
yeshpal-jain-seagate: Fixed

vidyadhar-pinglikar (approved):

sg-shankar (approved):

PR #1131, opened 2021-10-12
go/mio: fix build after arm64 port

andriytk:
Fix build after commit bc844a8 (ARM64 platform support) by including config.h file.

### DIFF
--- a/bindings/go/mio/mio.go
+++ b/bindings/go/mio/mio.go
@@ -45,6 +45,7 @@ package mio
 // #cgo CFLAGS: -Wno-attributes
 // #cgo LDFLAGS: -L../../../motr/.libs -Wl,-rpath=../../../motr/.libs -lmotr
 // #include <stdlib.h>
+// #include "config.h"
 // #include "lib/types.h"
 // #include "lib/trace.h"   /* m0_trace_set_mmapped_buffer */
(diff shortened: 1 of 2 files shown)


bindings/go/mio/mio.go
```diff
@@ -45,6 +45,7 @@ package mio
 // #cgo CFLAGS: -Wno-attributes
 // #cgo LDFLAGS: -L../../../motr/.libs -Wl,-rpath=../../../motr/.libs -lmotr
 // #include <stdlib.h>
+// #include "config.h"
```
huanghua78, line 48: Can you take a look at this file:
https://github.com/Seagate/cortx-motr/blob/main/motr/examples/example1.c
```
/*
 * Please change the dir according to you development environment.
 *
 * How to build:
 * gcc -I/work/cortx-motr -I/work/cortx-motr/extra-libs/galois/include \
 *     -DM0_EXTERN=extern -DM0_INTERNAL= -Wno-attributes               \
 *     -L/work/cortx-motr/motr/.libs -lmotr                            \
 *     example1.c -o example1
```

Maybe it also needs to be updated?
We can do it in another PR.

siningwuseagate (approved):

PR #1285, opened 2021-12-02
EOS-26618 dtm0: skip execution of REDO UT if DTM0 is disabled

sergey-shilov:
The DTM0 service is required for REDO UT. So this UT should not
be executed if DTM0 is disabled.


ivan-alekhin (approved):
It is not a ideal solution but it is the best we can do at the moment. REDOs are part of DTM0, so hiding this test case behind the feature flag seems to be the right way to go.

mehjoshi (approved):

PR #1327, opened 2021-12-17
lib: fix m0_uint128_sscanf() for large values

andriytk:
# Problem Statement

Currently, if any of the u64 values is bigger than LLONG_MAX, say, `0x8f9381cfacb7576a` - m0_uint128_sscanf() returns just LLONG_MAX for it (`0x7fffffffffffffff`), which is, obviously, wrong.

# Design
Solution: fix m0_uint128_sscanf() in such a way that if any of the u64 values are resulted in LLONG_MAX - it will try to scan the numbers in unsigned hex format.

Also, make 0x prefix for hex format optional. If 0x prefix is not specified, decimal format is tried. If it fails to
parse the values, hex format is tried then. This way we try our best to parse the numbers in various formats.

Relates to https://github.com/Seagate/cortx-motr/pull/486.

# Coding
   Checklist for Author

# Testing 
  Checklist for Author

# Impact Analysis
- [x] Interface change (if any) are documented
- [x] Side effects on other features (deployment/upgrade)
- [x] Dependencies on other component(s)

# Review Checklist 
  Checklist for Author
- [x] PR is self reviewed
- [x] Check if the description is clear and explained

# Documentation
  Checklist for Author

### DIFF
--- a/lib/misc.c
+++ b/lib/misc.c
@@ -51,6 +51,11 @@ M0_INTERNAL int m0_uint128_cmp(const struct m0_uint128 *u0,
 M0_INTERNAL int m0_uint128_sscanf(const char *s, struct m0_uint128 *u128)
 {
 	int rc = sscanf(s, U128I_F, U128_S(u128));
+
+	if (rc < 2 || u128->u_hi == LLONG_MAX || u128->u_lo == LLONG_MAX)
+		/* fallback to hex format */
+		rc = sscanf(s, U128X_F, U128_S(u128));
+
 	return rc == 2 ? 0 : -EINVAL;
 }
 
--- a/lib/types.h
+++ b/lib/types.h
@@ -66,6 +66,11 @@ M0_INTERNAL void m0_uint128_add(struct m0_uint128 *res,
 /** res = a * b; */
 M0_INTERNAL void m0_uint128_mul64(struct m0_uint128 *res, uint64_t a,
 				  uint64_t b);
+
+/**
+ * Scans uint128 in num:num format, where num can be in hex, octal
+ * or decimal format (depending on prefix).
+ */
 M0_INTERNAL int m0_uint128_sscanf(const char *s, struct m0_uint128 *u128);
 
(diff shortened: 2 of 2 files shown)


lib/misc.c
```diff
@@ -51,6 +51,9 @@ M0_INTERNAL int m0_uint128_cmp(const struct m0_uint128 *u0,
 M0_INTERNAL int m0_uint128_sscanf(const char *s, struct m0_uint128 *u128)
 {
 	int rc = sscanf(s, U128I_F, U128_S(u128));
+	if (rc == 2 && (u128->u_hi == LLONG_MAX ||
+	                u128->u_lo == LLONG_MAX))
+		rc = sscanf(s, U128X_F, U128_S(u128));
```
nikitadanilov, line 56: Won't it be better to use PRIu64 instead of PRIi64 in U128I_F? After all, it is _unsigned_ 128-bit integer, so why do we allow something like "-1:-2"?

nikitadanilov (changes requested):

lib/misc.c
andriytk: `PRIu64` allows only decimal format, but we want to be able to parse in hex format also (if 0x prefix is present).

nikitadanilov (approved):

lib/misc.c
nikitadanilov: Agree. Approved.

siningwuseagate (approved):

andriytk:
I've made a small change, so that we can parse the values in hex format even when `0x` prefix is not specified. For example, values like `a:b` can be parsed now. This way, we try our best to parse the numbers in various formats. @nikitadanilov, @siningwuseagate please review again. Thanks!

huanghua78 (approved):

nikitadanilov (approved):

siningwuseagate (approved):

mehjoshi (approved):

PR #1348, opened 2021-12-27
EOS-27125:  Errors in UT io-nw-xfer-ut and m0singlenode script

VinothVijayakumar-dev:
- Allocate ti_goff_ivec in UT testcase before free call
        - m0singlenode update default xprt fetch function

Signed-off-by: Vinoth.V <vinoth.v@seagate.com>

### DIFF
--- a/motr/ut/io_nw_xfer.c
+++ b/motr/ut/io_nw_xfer.c
@@ -386,6 +386,9 @@ static void ut_test_target_ioreq_fini(void)
 	M0_ALLOC_PTR(ti->ti_pageattrs);
 	rc = m0_indexvec_alloc(&ti->ti_ivec, 1);
 	M0_UT_ASSERT(rc == 0);
+	rc = m0_indexvec_alloc(&ti->ti_goff_ivec,1);
+	M0_UT_ASSERT(rc == 0);
+
 	target_ioreq_fini(ti);
 	ut_dummy_ioo_delete(ioo, dummy_instance);
 }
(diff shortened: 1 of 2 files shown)


utils/m0singlenode
```diff
@@ -32,8 +32,8 @@ readonly base_dir=$(dirname $self)
 M0_SRC_DIR=`readlink -f $0`
 M0_SRC_DIR=${M0_SRC_DIR%/*/*}
 
-. $M0_SRC_DIR/utils/functions # m0_default_xprt
-XPRT=$(m0_default_xprt)
+source /usr/libexec/cortx-motr/motr-service.functions
```
huanghua78, line 35: This script is supposed to work without installing Motr scripts/binaries into system?
We must make sure it can run from source code.

motr/ut/io_nw_xfer.c
```diff
@@ -386,6 +386,9 @@ static void ut_test_target_ioreq_fini(void)
 	M0_ALLOC_PTR(ti->ti_pageattrs);
 	rc = m0_indexvec_alloc(&ti->ti_ivec, 1);
 	M0_UT_ASSERT(rc == 0);
+	rc = m0_indexvec_alloc(&ti->ti_goff_ivec,1);
```
huanghua78, line 389: Can you please also check, where this "ti->ti_goff_ivec" is de-allocated?

utils/m0singlenode
madhavemuri: Agree.
Let's wait until https://jts.seagate.com/browse/EOS-26866 Selection of network transport based on configure option is ready.

PR #1394, opened 2022-01-17
Updated link in client.h overview

### LINKED ISSUE #1392: incorrect link for motr/client.h overview section
I found the link for  `exascale E10 stack (see eiow.org). ` in **overview** section no longer valid, eiow.org had link to a spam site.

`https://opensfs.org` probably is the latest/correct link.

hessio:
Signed-off-by: Patrick Hession <patrick.hession@seagate.com>

# Problem Statement
- Previous link was linked to spam site
- Updated to new link which is what was supposed to be linked to
Fixes #1392


mehjoshi:
@hessio, I see that https://opensfs.org takes me to Lustre filesystem page. It does not appear to be the same as the original E10 exascale content. I don't know if change is correct.
@andriytk , @huanghua78 , @madhavemuri any suggestions?

motr/client.h
```diff
@@ -50,7 +50,7 @@
  *
  *     - Motr-based block device (part of BOMO);
  *
- *     - exascale E10 stack (see eiow.org).
+ *     - exascale E10 stack (see https://opensfs.org).
```
madhavemuri, line 53: opensfs is not related to E10 work, domain of eiow is no longer there, let's remove it instead of changing it opensfs.

mehjoshi:
@hessio , As Madhav mentioned, can you remove the link and update the patch? Thanks.

motr/client.h
johnbent: @sainarasim and @nikitadanilov , is there any relevant E10 stack webpages that might still be useful to link? If not, then @madhavemuri is right and we should just remove this line altogether.

johnbent (changes requested):

motr/client.h
nikitadanilov: Yes, let's remove it.

mehjoshi (approved):

PR #1395, opened 2022-01-17
Update Quick-Start-Guide.rst

mengwanguc:
Add link on how to uninstall Libfabric. Otherwise users don't know how to uninstall it.

Signed-off-by: Meng Wang <mengwanguc@gmail.com>


cortx-admin:

andriytk (commented):
Users should usually know how to install / uninstall packages in their Linux distributions. Let's not clog the document with needless information.

Also, the document should be self-sufficient as much as possible, without needless links.

mengwanguc:
The problem was that, my OS (tried CentOS 7.8/8.3) didn't have `libfabric` installed.
It's installed when I run `sudo ./scripts/install-build-deps`
And then I'm asked to uninstall it myself
And I had to either google how to do it, or refer to the link above.

I guess we need to improve user experience somehow, because Motr is asking them to learn to uninstall something that was installed by Motr.

Is it possible to choose whether to use lnet or libfabic when running "sudo ./scripts/install-build-deps"?

andriytk:
The user might have libfabric already installed in his system even before `install-build-deps` is run. But your point is right about the user experience, thanks for sharing it.

What if we add the following text: "If it is installed, uninstall it manually using `sudo yum remove libfabric` or `sudo apt purge libfabric`, depending on your Linux distribution." Would that be helpful?

mehjoshi:
> The user might have libfabric already installed in his system even before `install-build-deps` is run. But your point is right about the user experience, thanks for sharing it.
> 
> What if we add the following text: "If it is installed, uninstall it manually using `sudo yum remove libfabric` or `sudo apt purge libfabric`, depending on your Linux distribution." Would that be helpful?

@mengwanguc , Can you respond to @andriytk's comment above?

mengwanguc:
> > The user might have libfabric already installed in his system even before `install-build-deps` is run. But your point is right about the user experience, thanks for sharing it.
> > What if we add the following text: "If it is installed, uninstall it manually using `sudo yum remove libfabric` or `sudo apt purge libfabric`, depending on your Linux distribution." Would that be helpful?
> 
> @mengwanguc , Can you respond to @andriytk's comment above?

Sure! Sorry about the late reply.

Hi @andriytk , I think your suggestion is good. I created a new PR #1440 here with the changes..

I'm closing this PR now.

PR #1500, opened 2022-02-21
Fix compilation with --enable-debug on CentOS 7.9

andriytk:
# Problem Statement

The compilation of kernel module on CentOS 7.9 with gcc-4.8 fails with --enable-debug configure option:

    conf/cache.c:226:31: error: ‘obj’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
            FID_P(&obj->co_id), obj->co_status);
                                 ^
    cc1: all warnings being treated as errors

The problem started appearing lately after we changed `-O0` flag to `-Og` for kernel modules compilation in the debug mode at commit 1fc7b4ba5.

# Design
Solution: initialise obj pointer at conf_cache_encode().

# Coding
   Checklist for Author

# Testing 
  Checklist for Author

# Impact Analysis

# Review Checklist 
  Checklist for Author
- [ ] PR is self reviewed

# Documentation
  Checklist for Author

### DIFF
--- a/conf/cache.c
+++ b/conf/cache.c
@@ -209,7 +209,7 @@ conf_encode(struct m0_confx *enc, const struct m0_conf_obj *obj, bool debug)
 static int conf_cache_encode(const struct m0_conf_cache *cache,
 			     struct m0_confx *dest, bool debug)
 {
-	struct m0_conf_obj *obj;
+	struct m0_conf_obj *obj = NULL;
 	int                 rc = 0;
 	size_t              nr;
 	char               *data;


madhavemuri (approved):

mehjoshi:
@andriytk , In Motr code, we do not explicitly initialize the variable upon declaration to avoid possible masking of a real problem. So, this change does not look correct to me.
Also, in this case the m0_tl_forall() macro will initialize the scan variable correctly.
@madhavemuri , Can you please review the change again?

madhavemuri:
> @andriytk , In Motr code, we do not explicitly initialize the variable upon declaration to avoid possible masking of a real problem. So, this change does not look correct to me. Also, in this case the m0_tl_forall() macro will initialize the scan variable correctly. @madhavemuri , Can you please review the change again?

I think there is a path where variable object is not modified, so it's fine to have initialization for such cases.

andriytk:
@mehjoshi, note: the problem appears only with the old version of gcc from CentOS 7.9 which we still officially support. Newer versions of gcc (8+) do not complain, as they probably can correctly recognise that the code is correct.

If you have any idea how to handle this particular case in a better way, I would be glad to hear. Thanks.

PR #1536, opened 2022-03-14
go: add high level go client

kiwionly2:
Add high level go client so can use by go code.

Signed-off-by: kiwionly2 <gheewooi.ong@seagate.com>

# Problem Statement
- Problem statement

# Design
-  For Bug, Describe the fix here.
-  For Feature, Post the link for design

# Coding
   Checklist for Author

# Testing 
  Checklist for Author

# Impact Analysis

# Review Checklist 
  Checklist for Author
- [ ] PR is self reviewed

# Documentation
  Checklist for Author

-----
[View rendered bindings/go/client/README.md](https://github.com/kiwionly2/cortx-motr/blob/high_level_go_client/bindings/go/client/README.md)

### DIFF
--- a/bindings/go/client/README.md
+++ b/bindings/go/client/README.md
@@ -1,0 +1,55 @@
+High level Client for Motr
+==========================
+
+To run this sample, first start standalone motr by running command below:
+
+```sh
+> sudo su root
+> cd cortx-motr/motr/examples/
+>./setup_a_running_motr_system.sh
+```
+
+Go to directory: `cortx-motr/bindings/go`
+
+Modify `mio/mio.go` find the variable and adding value as below. 
+Because these variables are package private, that was no way to modify in `client.go`.
+
+```golang
+var verbose bool = true
+var threadsN int = 1
+```
+
+Then go to directory: `cortx-motr/bindings/go/client`, open `client_test.go`
+and change which motr server to point to ( which the value you could obtain when start motr ) :
+
+```golang
+func TestInit() client option:
+
+	client, err = NewClient(&Options{
+		HAAddr:     "192.168.63.52@tcp:12345:34:1",
+		LocalAddr:  "192.168.63.52@tcp:12345:33:1000",
+		Profile:    "0x7000000000000001:0",
+		ProcessFId: "0x7200000000000001:64",
+		TraceOn:    false,
+	})
+```
+
+Then run command below to load test dependency library
+```sh
+> go mod tidy
+```
+
+lastly, run
+```sh
+> go test
+```
+
+You should see the result and the time it take to run the operation.
+
+If you received this error :
+```sh
+ERROR  [dix/req.c:801:dix_idxop_meta_update_ast_cb]  All items are failed
+```
+It mean index already exist, you might want to comment TestCreate().
+
+If the test case is hang, you probably not setting threadsN in mio.go.
--- a/bindings/go/client/client.go
+++ b/bindings/go/client/client.go
@@ -1,0 +1,436 @@
+/*
+ * Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * For any questions about this software or licensing,
+ * please email opensource@seagate.com or cortx-questions@seagate.com.
+ *
+ */
+
+package client
+
+/*
+#cgo CFLAGS: -I../../.. -I/usr/include/motr
+#cgo CFLAGS: -DM0_EXTERN=extern -DM0_INTERNAL=
+#cgo CFLAGS: -Wno-attributes
+#cgo LDFLAGS: -L../../../motr/.libs -Wl,-rpath=../../../motr/.libs -lmotr
+
+#include <stdlib.h>
+#include "motr/client.h"
+#include "lib/trace.h"
+
+struct m0_config    			conf;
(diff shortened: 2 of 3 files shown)


cortx-admin:

bindings/go/client/README.md
```diff
@@ -0,0 +1,54 @@
...
+```
+
+Go to directory: `cortx-motr/bindings/go`
+
+Modify `mio/mio.go` find the variable and adding value as below. 
+Because these variables are package private, that was no way to modify in `client.go`.
+
+```
+var verbose bool = true
+var threadsN int = 1
```
andriytk, line 20: These are not private variables. See how they are changed in mcp example.

johnbent:
@andriytk , I also notice there are multiple Codacy warnings. Should those be fixed or ignored?

bindings/go/client/README.md
kiwionly2: @andriytk Are you mean `mio/mio.go` ? I do see those variable value is set in `mio.go`  using command-line.

Because of how the way of code from `client` invoke code from `mio`, instead of get from command-line, it need a way to initial those variable in order to make it work.

I do think couple of way how to do it:

1. Make the variable capital case. ( so it is able public access outside of `mio` package )
2. Set a default value for those variable. ( but less flexible in code perspective, current implementation ) 
3. Make public setter for those variable. ( this look like better option, we could set those value at runtime ).

Any thought ? ( or I had misunderstanding you question )

andriytk: Yes, `mio/mio.go` automatically adds `-verbose` and `-threads` command-line options to manipulate these variables. `mcp/mcp.go` uses that. Why doesn't it work for you?

andriytk:
> @andriytk , I also notice there are multiple Codacy warnings. Should those be fixed or ignored?

They should be fixed, of course.

bindings/go/client/README.md
andriytk: Sorry, I don't understand this. Why does it work for mcp and doesn't work for your code?

kiwionly2: This is because `client` package do invoke method implement` io.Writer.Write()` from `mio.go`.

In `mio.go`:
```
// code implement io.Writer
func (mio *Mio) Write(p []byte) (n int, err error) {
    return mio.write(p, &mio.off)
}
```

```
func (mio *Mio) write(p []byte, off *int64) (n int, err error) {
    
    // some code omit here
    // alloc() is using threadsN
    if err = v.alloc(); err != nil {
        return 0, err
    }

```

```
func (v *iov) alloc() error {
    // threadsN may had no value, e.g value is not passing from command line,
    v.buf = make([]C.struct_m0_bufvec, threadsN)
```

Since this method is not call from command-line,  `threadsN`  default to zero and cause `client` code hang at runtime.

kiwionly2: Oh, I mean code in `client` package is indirectly calling io.Writer.Write() and eventually call `alloc()` and stuck there because `threadsN` suppose to be set to non zero before the `alloc()` invoke.

andriytk: `threadsN` default value is 1, as you can see from the code.

kiwionly2: Sorry to confuse, the original [mio.go](https://github.com/seagate/cortx-motr/blob/main/bindings/go/mio/mio.go#L132) is no set any value.

I was thinking maybe we could have a public method ( also for verbose variable ) to set the variable value ? so it could set any desired value.

kiwionly2: Ah.. ok, I had saw the code. 

Because the `client` code is not pass value through cmdline, I need to set the value some where. ( in this case, is in `mio.go`) 

Perhaps we could have a default value ? or public setter method ?

chandradharraval:
@kiwionly2 - Can you please fix codacy issues reported here?

kiwionly2:
@chandradharraval  codacy issues fixed.

bindings/go/client/README.md
kiwionly2: @andriytk I remove the change for mio.go and let user modify the file instead if they want to run client.go

madhavemuri:
@andriytk  and @siningwuseagate : Can you review this PR ?

andriytk:
The justification of such a high level go client is not convincing. The existing go/mio interface is high enough to write any motr client apps in Golang (see mcp and mkv examples).

johnbent:
Hello @kiwionly2 , thanks for this contribution! @andriytk thinks maybe this is redundant with the existing go bindings. Have you tried those? If you have tried those and you think this new client is still useful, can you please help us understand your thinking here? 

We want to make sure we include everything that makes the community experience better! But we also want to make sure we aren't adding redundant things.

@r-wambui , can you also please get involved here and evaluate whether the newly proposed client is redundant with what already exists?

kiwionly2:
The initial goal of `high level go client` serve slightly difference scenario then `mio`. One major difference is it it persist connection to `motr` and provide higher level API ( PUT, GET etc. ) without exposure much lower level implementation. It is a library instead of an application. ( e.g. like libmysqlclient for mysql )

Meanwhile `mio` interact with motr using command, to integrate into application, understanding underlying architecture is need and and also code modification.

Since these codes more like an experiment, reviewers could decide whether to merge into master :)

*The reason reuse `mio` code is because `mio` had provided all available features, it is unnecessary to rewrite code with same function again, so code will easy to maintain.

andriytk:
> higher level API ( PUT, GET etc. ) without exposure much lower level implementation. It is a library instead of an application.

`go/mio` is also library. `mcp` and `mkv` are just application examples of using this library.

And PUT/GET API is available there, see the usage example here - https://github.com/Seagate/cortx-motr/blob/3b70300f63ebd55fb477126b71b01f43326aaa97/bindings/go/mkv/mkv.go#L77

kiwionly2:
I read the code and aware that  `go\mio` using `flag` to initial connection parameters, 

Say If I using  `go\mio`  as a library for an REST API server, is that a way to initial the connection parameters programmatically ?

andriytk:
@kiwionly2, see https://github.com/Seagate/cortx-motr/pull/1740. This will allow passing motr connection parameters programmatically to `mio.Init()`.

madhavemuri:
@andriytk  @kiwionly2 : PR-1740 is merged in main, is this PR is still needed ?

kiwionly2:
@madhavemuri 

Yup, this longer need, I will close this PR. Thx !

PR #1602, opened 2022-04-11
CORTX-29857: GITHUB ISSUE-list unit benchmarks failed.

papan-singh:
Fix: make m0_ub_set struct object m0_timer_ub to non const.

Signed-off-by: root <root@ssc-vm-rhev4-2209.colo.seagate.com>

# Problem Statement
- Problem statement

# Design
-  For Bug, Describe the fix here.
-  For Feature, Post the link for design

# Coding
   Checklist for Author
-  [ ] Coding conventions are followed and code is consistent

# Testing 
  Checklist for Author
- [ ] Unit and System Tests are added
- [ ] Test Cases cover Happy Path, Non-Happy Path and Scalability
- [ ] Testing was performed with RPM

# Impact Analysis
  Checklist for Author/Reviewer/GateKeeper
- [ ] Interface change (if any) are documented
- [ ] Side effects on other features (deployment/upgrade)
- [ ] Dependencies on other component(s)

# Review Checklist 
  Checklist for Author
- [x] JIRA number/GitHub Issue added to PR
- [x] PR is self reviewed
- [x] Jira and state/status is updated and JIRA is updated with PR link
- [x] Check if the description is clear and explained

# Documentation
  Checklist for Author
- [ ] Changes done to WIKI / Confluence page / Quick Start Guide

PR #1622, opened 2022-04-13
m0_ub_set-s cannot be const: m0ub.c modifies them.

nikitadanilov:


mehjoshi:
@nikitadanilov, This change looks same as in https://github.com/Seagate/cortx-motr/pull/1602

huanghua78 (approved):

nikitadanilov:
> @nikitadanilov, This change looks same as in #1602

@mehjoshi, you are right. Thank you for spotting!

PR #1627, opened 2022-04-14
global_philter(): fix inherently racy UT.

nikitadanilov:
Additional addb records can be generated by
the system while the UT is running. This should
not cause a failure.

Signed-off-by: Nikita Danilov nikita.danilov@seagate.com


nikitadanilov:
@andriytk, @huanghua78: could you please review this?

madhavemuri (approved):

PR #1632, opened 2022-04-15
lib/user_space: use sigsetjmp() and siglongjmp().

nikitadanilov:
Cookie code (lib/user_space/ucookie.c) uses long jumps
to exit the signal handler. longjmp() restores the contents
of processor registers, including instruction and stack
pointers, to the state saved by setjmp(). longjmp() does
not restore the thread signal mask, which indicates how the
thread reacts to signals. But the signal mask is different
when the signal handler is executed (the signal being
handled is blocked), so longjmp() out of a signal handler
resumes execution with a wrong mask. Specifically, when the
cookie module jumps out of sigsegv(), the thread continues
with SIGSEGV blocked and the next receipt of SIGSEGV will
unconditionally terminate the process.

Fix this by using sigsetjmp() and siglogjmp().

### DIFF
--- a/lib/user_space/thread.h
+++ b/lib/user_space/thread.h
@@ -52,9 +52,9 @@ struct m0_thread_handle {
 /** User space thread-local storage. */
 struct m0_thread_arch_tls {
 	/** Non-zero iff the thread is in awkward context. */
-	uint32_t   tat_awkward;
+	uint32_t    tat_awkward;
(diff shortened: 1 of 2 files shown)


mehjoshi (approved):

andriytk (approved):

andriytk:
I wonder - how did we live with this issue for so long...

PR #1633, opened 2022-04-15
fdmi/ut: use m0_strdup().

nikitadanilov:
Results of strdup() cannot be used in motr code that
assumes that strings are freed by m0_free().

### DIFF
--- a/fdmi/ut/sd_apply_filter.c
+++ b/fdmi/ut/sd_apply_filter.c
@@ -185,7 +185,7 @@ static void fdmi_sd_apply_filter_internal(const struct m0_filterc_ops *ops)
 
 	M0_ENTRY();
 
-	g_var_str = strdup("test");
+	g_var_str = m0_strdup("test");
 	M0_SET0(&g_conf_filter);
 
 	fdmi_serv_start_ut(ops);
--- a/fdmi/ut/sd_send_not.c
+++ b/fdmi/ut/sd_send_not.c
@@ -253,7 +253,7 @@ void fdmi_sd_send_notif(void)
 	int                           rc;
 
 	M0_SET0(&g_conf_filter);
-	g_var_str = strdup("test");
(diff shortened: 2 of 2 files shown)


fdmi/ut/sd_apply_filter.c
```diff
@@ -185,7 +185,7 @@ static void fdmi_sd_apply_filter_internal(const struct m0_filterc_ops *ops)
 
 	M0_ENTRY();
 
-	g_var_str = strdup("test");
+	g_var_str = m0_strdup("test");
```
huanghua78, line 188: Is this `g_var_str` freed?

Let me check.

andriytk (approved):

fdmi/ut/sd_apply_filter.c
nikitadanilov: Yes, it is copied into a buffer in filterc_send_notif_get_next(), then passed down to fdmi and finally freed within fdmi. I know this, because with alloc-prof enabled, an attempt to m0_free() memory allocated with malloc() or strdup() crashes.

huanghua78 (approved):

fdmi/ut/sd_apply_filter.c
andriytk: > I know this, because with alloc-prof enabled, an attempt to m0_free() memory allocated with malloc() or strdup() crashes.

I wonder - how does it work? Because, according to the following code, m0_strdup is just an alias to strdup:
https://github.com/Seagate/cortx-motr/blob/9456562c6c86e2ea67f79c071005484770c0f8fd/lib/string.h#L43

nikitadanilov: Because m0_free() is not necessarily identical to free(). With alloc-profiler enabled it assumes allocated blocks has a certain header.

PR #1635, opened 2022-04-15
cas/ut/client_ut.c: use aligned free for aligned allocations.

nikitadanilov:
Memory allocated with m0_alloc_aligned() should be
freed with m0_free_aligned(). Existing m0_free()
happens to work correctly with memory allocated
by m0_alloc_aligned(), but this is against the
specification and will complicate the upcoming
allocation profiler.

### DIFF
--- a/cas/ut/client_ut.c
+++ b/cas/ut/client_ut.c
@@ -2063,8 +2063,8 @@ static void next_bulk(void)
 	vals_create(COUNT, COUNT_VAL_BYTES, &keys);
 	vals_create(COUNT, COUNT_VAL_BYTES, &values);
 	next_common(&keys, &values, 0);
-	m0_bufvec_free(&keys);
-	m0_bufvec_free(&values);
+	m0_bufvec_free_aligned(&keys, m0_pageshift_get());
+	m0_bufvec_free_aligned(&values, m0_pageshift_get());
 
 	/* Bulk values. */
 	rc = m0_bufvec_alloc(&keys, COUNT, sizeof(uint64_t));
@@ -2073,7 +2073,7 @@ static void next_bulk(void)
 	vals_create(COUNT, COUNT_VAL_BYTES, &values);
 	next_common(&keys, &values, 0);
 	m0_bufvec_free(&keys);
-	m0_bufvec_free(&values);
+	m0_bufvec_free_aligned(&values, m0_pageshift_get());
 
 	/* Bulk keys. */
 	vals_create(COUNT, COUNT_VAL_BYTES, &keys);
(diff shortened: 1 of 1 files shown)


andriytk:
Why free() should be aligned at all, what's the semantic of it?

nikitadanilov:
> Why free() should be aligned at all, what's the semantic of it? It does not make any sense.

@andriytk, it's a standard motr interface. Memory allocated with m0_alloc_aligned() must be freed with m0_free_aligned(), because m0_alloc() and m0_alloc_aligned() can be different allocators (and they are, in kernel).

andriytk:
> Memory allocated with m0_alloc_aligned() must be freed with m0_free_aligned(), because m0_alloc() and m0_alloc_aligned() can be different allocators (and they are, in kernel).

I'd argue that these are the details of the implementation, of which the user even should not know.

Anyway, why the alignment argument is needed at `m0_free_aligned()`? First, it's very inconvenient user interface (to remember the alignment) and error-prone. Second - why allocator cannot remember the alignment used on allocation, if it needs it when freeing?

nikitadanilov:
> Second - why allocator cannot remember the alignment used on allocation, if it needs it when freeing?

The allocator is an external module. Linux kernel does not remember alignment, this is why our interface looks like this. If you think about it a bit, remembering the alignment and size within allocator is tricky: where do you store this information? If you store it in some form of a header or footer adjacent to the allocated block, then alignment is difficult to maintain.

andriytk (approved):
Thanks for the explanations, @nikitadanilov. Although, I'm not sure if I can agree. The convenience of storing the alignment in the allocator implementation should not impact or prevail over the convenience of the user interface. I think.

Anyway, surely we are not in the position to change the kernel API, so +1.

nikitadanilov:
@huanghua78: could you please review this?

cas/ut/client_ut.c
```diff
@@ -2063,8 +2063,8 @@ static void next_bulk(void)
 	vals_create(COUNT, COUNT_VAL_BYTES, &keys);
```
huanghua78, line 2063: How about renaming `vals_create()` to `vals_create_aligned()`?

The patch looks fine. If we can rename this or add comments for `vals_create()`, it would be better.

```
diff --git a/cas/ut/client_ut.c b/cas/ut/client_ut.c
index 3194c4c3..c204340a 100644
--- a/cas/ut/client_ut.c
+++ b/cas/ut/client_ut.c
@@ -140,6 +140,10 @@ static void value_create(int size, int num, char *buf)
        }
 }

+/**
+ * Allocates bufvec(s) aligned.
+ * The result must be freed by m0_bufvec_free_aligned().
+ */
 static void vals_create(int count, int size, struct m0_bufvec *vals)
 {
        int i;

```

madhavemuri: @nikitadanilov : Can you reply to the comment from @huanghua78  ?

cortx-admin:

shailesh-vaidya:
Closing as an obsolete

PR #1671, opened 2022-04-24
m0_nanosleep(): nanosleep() returns -1, use errno.

nikitadanilov:


huanghua78 (approved):

andriytk (commented):
In a multithreaded process, errno might be reset by another call very quickly. Are you sure about this change?

ivan-alekhin:
> In a multithreaded process, errno might be reset by another call very quickly (in which case the error will be missed, as 0 will be returned). Are you sure about this change?

@andriytk , usually, `errno` is a thread-local variable (or rather a macro that lets us access a thread-local variable). So that, the only concern here is that we need to ensure that there are no calls that modify errno between the point where nanosleep() done its work and the point where we read out this variable inside M0_ERR. I don't see anything problem like that in M0_ERR.

andriytk (approved):

PR #1760, opened 2022-05-11
Fix a race between flip fom processing and asts

nikitadanilov:
To process confd flip fom, conf cache is destroyed and re-initialised. And
concurrent activity that tries to access the cache will see the cache in
half-destructed state.

Fix: introduce a new per-reqh lock and use it for synchronisation.


cortx-admin:

shailesh-vaidya:
Closing as an obsolete

PR #1545, opened 2022-03-16
CORTX-27608 dix/dtm0: add support of transient failures for PUT

sergey-shilov:
Problem:
Currently there is no support for TRANSIENT failures on the DIX
and DTX levels for DIX PUT operations, only the happy path is
supported (with and without DTM0 enabled). In case of TRANSIENT
failure the DIX PUT operation hangs in awaiting of response from
the OFFLINE process as currently the requests are sent to OFFLINE
devices (as TRANSIENT notification is mapped to OFFLINE devices
state).

Solution:
In order to support transient failures for DIX PUT operations the
following changes have been introduced:
 - skip sending of CAS requests to OFFLINE devices;
 - add services that correspond to OFFLINE devices into
   participants list as they are still the cluster members, when
   they are back we need to restore missing info, that's why we
   need to add them also;
 - move transient participants into EXECUTED state right after
   CAS requests is sent (see comment below regarding EXECUTED-ALL
   logic);
 - move participants for those CAS requests that are failed to be
   sent into EXECUTED state (see comment below regarding EXECUTED-ALL
   logic);
 - move DTX into STABLE state when at least 1 PERSISTENT message
   received on the client side;
 - remove the log record from the volatile DTM0 log when DTX reaches
   the STABLE state, we can clean this up as currently the client
   (volatile log lives on the client side only) does not participate
   in recovery process, no need to keep stable transactions.

Moving of transient participants and those for which CAS requests
are failed to be sent into EXECUTED state is made to unblock the
EXECUTED-ALL logic. It allows to move transaction to the stable
state once the persistent message received (EXECUTED state
required for all participants). So EXECUTED participant state is
reused in case of failures.

Signed-off-by: Sergey Shilov <sergey.shilov@seagate.com>

PR #1804, opened 2022-05-20
CORTX-30872 dix: add support of transient failures for GET

sergey-shilov:
Problem:
Before merging of degraded DIX PUT support for transient
failures (PR https://github.com/Seagate/cortx-motr/pull/1545)
the degraded GET worked by a mistake: sending of requests
to offline targets + resend logic.
The PR mentioned above prevents sending of requests to
offline devices, so the offline unit is skipped. Other alive
targets are not selected due to dix_online_unit_choose()
function (limits the number of units to be selected to 1
to send a single request for DIX GET, will be removed by
current commit). So if the first unit is offline then whole
DIX operation wails with -EIO immediately and no resend logic
is triggered.
So need a proper way to handle transient failures and another
logic of target selection for DIX GET.

Solution:
Make logic of targets selection for DIX GET as close to logic
for DIX PUT as possible, i.e. no special marking of targets
as "failed" to limit the number of requests to 1, now this
limitation is done by breaking a CAS requests creation loop
when a single CAS request was successfully created. Finally
the CAS GET request will be sent to a first alive unit of
the parity group.
Also this approach is the first step to versioned CAS usage:
sending of a bunch of CAS GET requests can be done by a very
little change.

### DIFF
--- a/dix/req.c
+++ b/dix/req.c
@@ -1977,7 +1977,7 @@ static int dix__spare_target(struct m0_dix_rec_op         *rec_op,
 		if (rc != 0)
 			return M0_ERR(rc);
 		spare = &rec_op->dgp_units[spare_offset + slot];
-		if (!spare->dpu_failed) {
+		if (!spare->dpu_unavail) {
 			/* Found non-failed spare unit, exit the loop. */
 			*spare_unit = spare;
 			*spare_slot = slot;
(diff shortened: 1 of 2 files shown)


nikitadanilov (changes requested):
As discussed elsewhere #1803 is a better approach.

PR #1833, opened 2022-05-27
motr/idx_dix: Add conditional dtx clink init/fini.

ivan-alekhin:
Problem: initialization/finalization of the dtx clink inside
dix_req is not aligned with the conditional nature of this
clink. It may cause a panic.

Solution: initialize/finalize the dtx clink under a condition.

This PR is for an internal review of dtm/recovery feature branch. Hua reported an issue in the code, and this PR is an attempt to fix it. Once @huanghua78  and @sergey-shilov  are ok with this patch (here or in the chat), I'll merge it into dtm/recovery (probably, it would be better to squash it into  "dtm0/dtx: Remove EXECUTED from DTX. " commit).

### DIFF
--- a/motr/idx_dix.c
+++ b/motr/idx_dix.c
@@ -72,6 +72,11 @@ struct dix_req {
 	struct m0_sm_ast         idr_ast;
 	struct m0_clink          idr_clink;
 
+	/**
+	 * Clink that connects STABLE/FAILURE states of DTX with this dix_req.
+	 * Note, the clink is always initialised but not always attached
+	 * to a dtx.
+	 */
 	struct m0_clink          idr_dtx_clink;
 
 	/**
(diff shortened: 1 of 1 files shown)


sergey-shilov (approved):

huanghua78:
This patch looks OK and I have verified that it works.
But I have the following suggestions.
In the patch, we initialize the "req->idr_dtx_clink" when the dtx is not NULL, and finalize it with the same condition.
Can we init and fini it un-conditionally (but it is not used)?
My reason is:
If we only initialize the "req->idr_dtx_clink" when the dtx is not NULL, if we are doing some debugging,
you will see this "req->idr_dtx_clink" is all zero (which is invalid). This will be confusing to developers.
How do you think?

ivan-alekhin:
>  But I have the following suggestions. ...

Thank you @huanghua78, it is a good idea, I did not think about debugging. I changed the code accordingly. Please take a look.

huanghua78 (approved):

ivan-alekhin:
merged as 46cf899ccbf0b24feb59fd729ce805c984d69e60

PR #1834, opened 2022-05-27
Fix "Invalid entity fid detected" issue when boundary condition obj_id is passed …

wb-droid:
…to Motr Object Application.

Signed-off-by: wb-droid <bo.b.wei@seagate.com>

# Problem Statement
- When obj_id of 1048576 is passed as input to the example below, the "Invalid entity fid detected" error is encountered. 
./example1 192.168.53.101@tcp:12345:34:1 192.168.53.101@tcp:12345:33:1000 0x7000000000000001:0 0x7200000000000001:64 1048576 
motr[04290]:  f0c0  ERROR  [client.c:360:entity_id_is_valid]  Invalid entity fid detected: <0:100000>.Entity id should larger than M0_ID_APP. 

-  The fix is to treat this boundary case as invalid input.

### DIFF
--- a/motr/examples/example1.c
+++ b/motr/examples/example1.c
@@ -401,7 +401,7 @@ int main(int argc, char *argv[])
 		exit(-1);
 	}
 	obj_id.u_lo = atoll(argv[5]);
-	if (obj_id.u_lo < M0_ID_APP.u_lo) {
+	if (obj_id.u_lo <= M0_ID_APP.u_lo) {
 		printf("obj_id invalid. Please refer to M0_ID_APP "
 		       "in motr/client.c\n");
 		exit(-EINVAL);


cortx-admin:

mehjoshi (changes requested):
The change is not correct. Here is the comment from motr/client.c:

```
/**
 * First identifier the application is allowed to use.
 * The first 0x100000 ids are reserved for use by client.
 */
const struct m0_uint128 M0_ID_APP = { 0ULL, 0x100000ULL };
```

wb-droid:
@mehjoshi , the comment is a bit misleading. The actual logic to decide valid entity id is entity_id_is_valid() in motr/client.c below. So in actual implementation, M0_ID_APP is an invalid entity id. 

https://github.com/Seagate/cortx-motr/blob/b3606f2075db556099446449e285be16ede6ff18/motr/client.c#L351-L365

mehjoshi:
I searched looking for other places where M0_ID_APP is checked against. I found the following comparison check in 

https://github.com/Seagate/cortx-motr/blob/b3606f2075db556099446449e285be16ede6ff18/motr/io.c#L234-L245

It looks like entity id needs to be strictly greater than M0_ID_APP. So, I think the change is correct.

mehjoshi (approved):

mehjoshi:
ok to test

huanghua78 (approved):

PR #1869, opened 2022-06-09
CORTX-31180 cas: fix sdev ID if it does not match local sdev

sergey-shilov:
Problem:
REDO messages sent by the DTX participants contain CAS operation
with a component catalogue FID containing device ID that is local
for each participant.
Handling of the REDO message resulting in a processing of CAS
operation. During processing of any CAS operation we didn't check
that device encoded into component catalogue FID is served by
the target CAS service using conf cache (rely on the client).
But in case of REDO device ID is incorrect, that results into
a new component catalogue creation using CROW mechanism that is
incorrect.

Solution:
Check the device ID of the incoming request and fix it if it does
not match the device that is attached to the CAS service that is
handling the request.

Assumption: CAS service has a single attached device that is the
right configuration currently. For some unit tests (DIX client UTs)
we have a configuration with a several attached devices to emulate
several CAS services. In that case the device ID fixing logic is
disabled.

Issues:
For the recovering process the state of attached device is OFFLINE,
but we need to handle REDO messages that result into processing of
CAS operation. Device state is checked in this case and now we need
to treat OFFLINE device as ok for operation that is logically
incorrect. Probably need to add a special device state that will
be exposing RECOVERING state.

### DIFF
--- a/cas/service.c
+++ b/cas/service.c
@@ -311,6 +311,14 @@ enum {
 struct cas_service {
 	struct m0_reqh_service  c_service;
 	struct m0_be_domain    *c_be_domain;
+
+	/*
+	 * sdev used by an instance of CAS service. This should be just one,
+	 * but current unit tests use more than one count and to make
+	 * sure those continue to run we skip the sdev replacement if more than
+	 * one sdev is discovered to be serviced by this cas service.
+	 */
+	uint32_t                c_sdev_id;
 };
 
 struct cas_kv {
@@ -477,8 +485,8 @@ static int  cas_buf_cid_decode(struct m0_buf    *enc_buf,
 			       struct m0_cas_id *cid);
 static bool cas_fid_is_cctg(const struct m0_fid *fid);
 static int  cas_id_check(const struct m0_cas_id *cid);
-static int  cas_device_check(const struct cas_fom   *fom,
-			     const struct m0_cas_id *cid);
+static int  cas_device_check(const struct cas_fom *fom,
+			     struct m0_cas_id     *cid);
 static int cas_op_check(struct m0_cas_op *op,
 			struct cas_fom   *fom,
 			bool              is_index_drop);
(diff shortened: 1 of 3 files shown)


cas/service.c
```diff
@@ -311,6 +311,14 @@ enum {
 struct cas_service {
 	struct m0_reqh_service  c_service;
 	struct m0_be_domain    *c_be_domain;
+
+	/*
```
nikitadanilov, line 315: Doxygen, please.

cas/service.c
```diff
@@ -556,6 +564,60 @@ m0_cas__ut_svc_be_get(struct m0_reqh_service *svc)
...
+	service = M0_CONF_CAST(obj, m0_conf_service);
+	M0_ASSERT(service != NULL);
+
+	sdevs_dir = &service->cs_sdevs->cd_obj;
+	rc = m0_confc_open_sync(&sdevs_dir, sdevs_dir, M0_FID0);
+	if (rc != 0)
+		return M0_RC(rc);
+
+	obj = NULL;
+	while ((rc = m0_confc_readdir_sync(sdevs_dir, &obj)) > 0) {
```
nikitadanilov, line 600: What happens when no device is configured and m0_confc_readdir_sync() returns 0? We return 0 without assigning anything to ->c_sdev_id. Is this right?

cas/service.c
```diff
@@ -556,6 +564,60 @@ m0_cas__ut_svc_be_get(struct m0_reqh_service *svc)
...
+			cas_svc->c_sdev_id = sdev->sd_dev_idx;
+			found = true;
+		} else {
+			/*
+			 * Several devices attached to a single CAS service are
+			 * possible if we are run in ut. In this case we
+			 * do not replace the device ID and leave the ut alone.
+			 * Also we can not break the loop until whole directory
+			 * is iterated since it leads to crash during request
+			 * handler finalisation.
```
nikitadanilov, line 612: Hmm... why? Do we understand what is the problem here?

cas/service.c
```diff
@@ -1826,12 +1890,25 @@ static int cas_device_check(const struct cas_fom   *fom,
...
+		if (svc->c_sdev_id != UINT32_MAX &&
+		    svc->c_sdev_id != device_id &&
+		    cas_type(&fom->cf_fom) != CT_META) {
+			M0_LOG(M0_DEBUG,
+			       "Incorrect device ID (%d) found in a component "
+			       "catalogue FID, possibly because of REDO msg. "
+			       "Replacing it with correct device ID (%d).",
+			       device_id, svc->c_sdev_id);
+			m0_dix_fid__device_id_set(&cid->ci_fid, svc->c_sdev_id);
+			device_id = svc->c_sdev_id;
```
nikitadanilov, line 1902: That's fishy and potentially masks errors. Can the correct device identifier be set on the sender side when the REDO message is formed?

nikitadanilov (changes requested):

cas/service.c
nikitadanilov: As discussed with @ivan-alekhin, this is fine as a temporary solution.

andriytk:
> But in case of REDO device ID is incorrect,..

Why it can be incorrect? Is this even a legal scenario?

sergey-shilov:
> > But in case of REDO device ID is incorrect,..
> 
> Why it can be incorrect? Is this even a legal scenario?

@andriytk Because each CAS service receives a request with a CAS id that contains device fid attached to that particular service, this CAS id is written into DTM0 log as a payload, so each service stores its own copy of CAS id. This DTM0 log record is sent as a part of REDO message, so CAS id will be different from each service participating in recovery process.

cas/service.c
sergey-shilov: Not completely, need to root cause.

cas/service.c
sergey-shilov: @nikitadanilov 
When I saw the design proposed in corresponding ticket
https://jts.seagate.com/browse/CORTX-31180
I also mentioned that the solution may be risky and proposed to think about two possible approaches:
* do the same but on a REDO sender side;
* extend participants info with corresponding devices id info.
Reference:
https://jts.seagate.com/browse/CORTX-31180?focusedCommentId=2370635&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-2370635

There was no answer on that, I looked through the code and answered my self:
https://jts.seagate.com/browse/CORTX-31180?focusedCommentId=2371345&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-2371345

So the final decision was as the initial proposal.
If you @nikitadanilov thinks that we need to review the approach ones again - let's do it.

BTW, didn't understand about the temporary solution... What's stopping us to do the final solution now? Do we expect some changes (DTM0 log data structures and so on) that will help us in a more safe solution?

cas/service.c
sergey-shilov: Yes, ->c_sdev_id will be invalid (UINT32_MAX) and no substitution logic will be applied. The same is for non-standard configurations with multiple attached devices.

cas/service.c
sergey-shilov: Sending of REDO messages is done by the common code of recovery machine, so we don't distinguish the payload (may be IOs in future and so on). May be we can register in recovery machine some pre-sending callback to make required changes in a payload? In this case we need some service type ID embedded in a log record, or can call CAS pre-sending callback directly in a first version (since only KVS is supported currently). @nikitadanilov @ivan-alekhin What do you think?

sergey-shilov: I think I can quickly prepare a PoC with a callback on a sender side.

cas/service.c
shashank-parulekar: Done

sergey-shilov:
An alternative solution that is treated as primary for now
https://github.com/Seagate/cortx-motr/pull/1882/

sergey-shilov:
Merged the alternative solution:
https://github.com/Seagate/cortx-motr/pull/1882/
Closing this PR.

PR #1882, opened 2022-06-14
CORTX-31180 dtm0/cas: fix sdev ID for incoming REDO messages

sergey-shilov:
Problem:
REDO messages sent by the DTX participants contain CAS operation
with a component catalogue FID containing device ID that is local
for each participant.
Handling of the REDO message resulting in a processing of CAS
operation. During processing of any CAS operation we didn't check
that device encoded into component catalogue FID is served by
the target CAS service using conf cache (rely on the client).
But in case of REDO device ID is incorrect, that results into
a new component catalogue creation using CROW mechanism that is
incorrect.

Solution:
Find attached device in conf and store it inside of CAS service
structure when CAS service starts.
When the REDO message arrives the REDO message handler related to
CAS service gets device id attached to a local CAS service and
sets it as a part of operation CAS id.

Assumption:
CAS service has a single attached device that is the standard
configuration currently. For some unit tests (DIX client UTs)
we have a configuration with a several attached devices to emulate
several CAS services. In that case the device ID stored in a CAS
service structure is set to a special invalid value.

Issues:
For the recovering process the state of attached device is OFFLINE,
but we need to handle REDO messages that result into processing of
CAS operation. Device state is checked in this case and now we need
to treat OFFLINE device as ok for operation that is logically
incorrect. Probably need to add a special device state that will
be exposing RECOVERING state.

Signed-off-by: Sergey Shilov <sergey.shilov@seagate.com>

PR #1906, opened 2022-06-20
CORTX-31180: Validate sdev in the incoming DIX request.

shashank-parulekar:
# Problem Statement
- In case of code bugs a CAS request could be received with an incorrect sdev field in the DIX FID which could end up creating new indices in case of CROW requests.

# Design
-  The committed code will confirm the sdev ID is correct and if not it will either ASSERT (for debug builds) or return error to the caller thus aborting this operation.

# Coding
   Checklist for Author

P.S.: This PR is a follow-up to [PR 1882](https://github.com/Seagate/cortx-motr/pull/1882)


sergey-shilov:
Since that change is not related to DTM iteslf I propose to wait until dtm/recovery is merged into main and then re-created the PR against main.

PR #1907, opened 2022-06-21
Changes in BE allocator to support no-clear mode during allocation.

shashank-parulekar:
# Problem Statement
- Current BE allocator will clear memory contents when m0_alloc*() routines are
called. This is not always desirable since the Tx credits requirements increase
because the cleared memory needs to be captured in the transaction.

# Coding
   Checklist for Author

# Testing 
  Checklist for Author
- [x] Unit and System Tests are added

# Impact Analysis
- [x] Interface change (if any) are documented

# Review Checklist 
  Checklist for Author
- [x] JIRA number/GitHub Issue added to PR
- [x] PR is self reviewed

This commit contains:
1) Changes to skip memory clearing during allocation.
2) Revised credit needs when no-clear version of the allocation is called.
3) UT to verify memory allocation contents.
4) Btree changes to use he no-clear memory routines.

Signed-off-by: Shashank Parulekar <Shashank.Parulekar@seagate.com>

### DIFF
--- a/be/alloc.c
+++ b/be/alloc.c
@@ -924,11 +924,12 @@ M0_INTERNAL void m0_be_allocator_destroy(struct m0_be_allocator *a,
 	M0_LEAVE();
 }
 
-M0_INTERNAL void m0_be_allocator_credit(struct m0_be_allocator *a,
-					enum m0_be_allocator_op optype,
-					m0_bcount_t             size,
-					unsigned                shift,
-					struct m0_be_tx_credit *accum)
+M0_INTERNAL void m0_be_allocator_credit_common(struct m0_be_allocator *a,
+					       enum m0_be_allocator_op optype,
+					       m0_bcount_t             size,
+					       unsigned                shift,
+					       struct m0_be_tx_credit *accum,
+					       bool                    mem_zero)
 {
 	/*
 	 * XXX `a' can be NULL, &(struct m0_be_seg)(0)->bs_allocator or
@@ -965,7 +966,8 @@ M0_INTERNAL void m0_be_allocator_credit(struct m0_be_allocator *a,
 			    &M0_BE_TX_CREDIT_PTR(&h->bah_addr));
 
 	shift = max_check(shift, (unsigned) M0_BE_ALLOC_SHIFT_MIN);
-	mem_zero_credit = M0_BE_TX_CREDIT(1, size);
+	mem_zero_credit = (mem_zero) ? M0_BE_TX_CREDIT(1, size)
+				     : M0_BE_TX_CREDIT(0, 0);
 
 	chunks_all_be_list_credit(M0_BLO_CREATE,  1, &cred_list_create);
 	chunks_all_be_list_credit(M0_BLO_DESTROY, 1, &cred_list_destroy);
@@ -1044,14 +1046,33 @@ M0_INTERNAL void m0_be_allocator_credit(struct m0_be_allocator *a,
 	m0_be_tx_credit_add(accum, &M0_BE_TX_CREDIT(40, 640));
 }
 
-M0_INTERNAL void m0_be_alloc_aligned(struct m0_be_allocator *a,
-				     struct m0_be_tx *tx,
-				     struct m0_be_op *op,
-				     void **ptr,
-				     m0_bcount_t size,
-				     unsigned shift,
-				     uint64_t zonemask,
-				     bool chunk_align)
+M0_INTERNAL void m0_be_allocator_credit(struct m0_be_allocator *a,
+					enum m0_be_allocator_op optype,
+					m0_bcount_t             size,
+					unsigned                shift,
+					struct m0_be_tx_credit *accum)
+{
+	m0_be_allocator_credit_common(a, optype, size, shift, accum, true);
+}
+
+M0_INTERNAL void m0_be_allocator_credit_no_clear(struct m0_be_allocator *a,
+					enum m0_be_allocator_op optype,
+					m0_bcount_t             size,
+					unsigned                shift,
+					struct m0_be_tx_credit *accum)
+{
+	m0_be_allocator_credit_common(a, optype, size, shift, accum, false);
+}
+
+M0_INTERNAL void m0_be_alloc_aligned_common(struct m0_be_allocator *a,
+				     	    struct m0_be_tx *tx,
+				     	    struct m0_be_op *op,
+				     	    void **ptr,
+				     	    m0_bcount_t size,
+				     	    unsigned shift,
+				     	    uint64_t zonemask,
+				    	    bool chunk_align,
+					    bool clear_space)
 {
 	enum  m0_be_alloc_zone_type  ztype;
 	struct be_alloc_chunk       *c = NULL;
@@ -1084,8 +1105,11 @@ M0_INTERNAL void m0_be_alloc_aligned(struct m0_be_allocator *a,
 					    chunk_align);
 		M0_ASSERT(c != NULL);
 		M0_ASSERT(c->bac_zone == ztype);
-		memset(&c->bac_mem, 0, size);
-		m0_be_tx_capture(tx, &M0_BE_REG(a->ba_seg, size, &c->bac_mem));
+		if (clear_space) {
+			memset(&c->bac_mem, 0, size);
+			m0_be_tx_capture(tx, &M0_BE_REG(a->ba_seg, size,
+					 &c->bac_mem));
+		}
 	}
 	*ptr = c == NULL ? NULL : &c->bac_mem;
 	be_allocator_stats_update(&a->ba_h[ztype]->bah_stats,
@@ -1120,6 +1144,32 @@ M0_INTERNAL void m0_be_alloc_aligned(struct m0_be_allocator *a,
 	m0_be_op_done(op);
 }
 
+M0_INTERNAL void m0_be_alloc_aligned(struct m0_be_allocator *a,
+				     struct m0_be_tx *tx,
+				     struct m0_be_op *op,
+				     void **ptr,
+				     m0_bcount_t size,
+				     unsigned shift,
+				     uint64_t zonemask,
+				     bool chunk_align)
+{
+	m0_be_alloc_aligned_common(a, tx, op, ptr, size, shift, zonemask,
+				   chunk_align, true);
+}
+
+M0_INTERNAL void m0_be_alloc_aligned_no_clear(struct m0_be_allocator *a,
+					      struct m0_be_tx *tx,
+					      struct m0_be_op *op,
+					      void **ptr,
+					      m0_bcount_t size,
+					      unsigned shift,
+					      uint64_t zonemask,
+					      bool chunk_align)
+{
+	m0_be_alloc_aligned_common(a, tx, op, ptr, size, shift, zonemask,
+				   chunk_align, false);
+}
+
 M0_INTERNAL void m0_be_alloc(struct m0_be_allocator *a,
 			     struct m0_be_tx *tx,
 			     struct m0_be_op *op,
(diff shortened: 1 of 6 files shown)


be/ut/alloc.c
```diff
@@ -192,6 +192,145 @@ M0_INTERNAL void m0_be_ut_alloc_concurrent(void)
...
+
+	m0_be_tx_close_sync(&tx);
+	m0_be_tx_fini(&tx);
+	m0_be_ut_seg_allocator_fini(ut_seg, ut_be);
+	m0_be_ut_seg_fini(ut_seg);
+	m0_be_ut_backend_fini(ut_be);
+	M0_SET0(ut_be);
+
+
+	tx = tx;
```
upendrapatwardhan, line 326: Is this required?

be/ut/alloc.c
```diff
@@ -192,6 +192,145 @@ M0_INTERNAL void m0_be_ut_alloc_concurrent(void)
...
+	M0_BE_OP_SYNC(op, m0_be_alloc_aligned(a, &tx, &op, &p, size,
+					      shift, M0_BITS(M0_BAP_NORMAL),
+					      false));
+	/**
+	 * We know the allocator uses the LIFO scheme when dispersing memory to
+	 * its callers. Based on this we can assume to get the recently released
+	 * memory back with the allocation call above.
+	 */
+	M0_ASSERT(p != NULL && p == saved_p);
+	for (i=0; i < size/sizeof(uint64_t); i++) {
```
ydb242, line 290: space before and after =

be/ut/alloc.c
```diff
@@ -192,6 +192,145 @@ M0_INTERNAL void m0_be_ut_alloc_concurrent(void)
...
+	}
+	M0_BE_OP_SYNC(op, m0_be_free_aligned(a, &tx, &op, p));
+
+
+	M0_BE_OP_SYNC(op, m0_be_alloc_aligned_no_clear(a, &tx, &op, &p, size,
+						       shift,
+						       M0_BITS(M0_BAP_NORMAL),
+						       false));
+	M0_ASSERT(p != NULL);
+	for (i=0; i < size/sizeof(uint64_t); i++) {
```
ydb242, line 301: space before and after =

be/ut/alloc.c
```diff
@@ -192,6 +192,145 @@ M0_INTERNAL void m0_be_ut_alloc_concurrent(void)
...
+	}
+
+	M0_BE_OP_SYNC(op, m0_be_free_aligned(a, &tx, &op, p));
+
+	M0_BE_OP_SYNC(op, m0_be_alloc_aligned_no_clear(a, &tx, &op, &p, size,
+						       shift,
+						       M0_BITS(M0_BAP_NORMAL),
+						       false));
+	M0_ASSERT(p != NULL && p == saved_p);
+	for (i=0; i < size/sizeof(uint64_t); i++) {
```
ydb242, line 313: space before and after =

be/ut/alloc.c
```diff
@@ -192,6 +192,145 @@ M0_INTERNAL void m0_be_ut_alloc_concurrent(void)
...
+	m0_be_ut_tx_init(&tx, ut_be);
+	m0_be_tx_prep(&tx, &cred);
+	rc = m0_be_tx_open_sync(&tx);
+
+	M0_BE_OP_SYNC(op, m0_be_alloc_aligned(a, &tx, &op, &p, size,
+					      shift, M0_BITS(M0_BAP_NORMAL),
+					      false));
+	M0_ASSERT(p != NULL);
+	saved_p = p;
+	for (i=0; i < size/sizeof(uint64_t); i++) {
```
ydb242, line 274: space before and after =

be/alloc.c
```diff
@@ -1084,8 +1105,11 @@ M0_INTERNAL void m0_be_alloc_aligned(struct m0_be_allocator *a,
 					    chunk_align);
 		M0_ASSERT(c != NULL);
 		M0_ASSERT(c->bac_zone == ztype);
-		memset(&c->bac_mem, 0, size);
-		m0_be_tx_capture(tx, &M0_BE_REG(a->ba_seg, size, &c->bac_mem));
+		if (clear_space) {
+			memset(&c->bac_mem, 0, size);
+			m0_be_tx_capture(tx, &M0_BE_REG(a->ba_seg, size,
+					 &c->bac_mem));
```
ydb242, line 1111: alignment

btree/btree.c
```diff
@@ -1748,7 +1754,8 @@ static void bnode_move(struct nd *src, struct nd *tgt, enum direction dir,
 
 static void bnode_capture(struct slot *slot, struct m0_be_tx *tx)
 {
-	slot->s_node->n_type->nt_capture(slot, tx);
+	struct nd* node = (struct nd *)slot->s_node;
```
kanchan-chaudhari, line 1757: Why Do we need this change?? Also why to pass `struct node *` to `nt_capture()`, we are not using it inside those functions.

be/ut/alloc.c
shashank-parulekar: This is to suppress compiler warning about variable not used.

btree/btree.c
```diff
@@ -3514,6 +3527,7 @@ static void fkvv_init(const struct segaddr *addr, int ksize, int vsize,
 	 * other place should the opaque data get captured and written to BE
 	 * segment.
 	 */
+	M0_BTREE_TX_CAPTURE(tx, seg, &h->fkvv_opaque, sizeof(h->fkvv_opaque));
```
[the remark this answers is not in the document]
shashank-parulekar: I think these might be from before as I have not changed these lines !!!

btree/btree.c
shashank-parulekar: We need this change because as a part of optimization we will keep the information of what is to be captured in the node descriptor which will be accessible to the capture routines. Also the slot information is not needed because the captures only deal with capturing of the data within the node and not specific to slots. If the record index is ever needed to help the capture then that information can be kept in some node variables to be used by the capture routines.

btree/btree.c
```diff
@@ -2684,6 +2696,7 @@ static void ff_init(const struct segaddr *addr, int ksize, int vsize, int nsize,
 	 * other place should the opaque data get captured and written to BE
 	 * segment.
 	 */
+	M0_BTREE_TX_CAPTURE(tx, seg, &h->ff_opaque, sizeof(h->ff_opaque));
```
RadhaGulhane13, line 2699: I am not sure if this will be the right place to capture opaque pointer, because of following two reasons:
1. We perform node initializaion even during the bnode_alloc(), but sometimes we might not use the allocated node by bnode_alloc() [in cases where we allocated nodes by seeing possibility of overflow and then don't use it]. So, in such cases we are capturing the region unnecessary.

2. In current implementation, we are calling bnode_capture() explicitly for root nodes dureing tree create, so capturing opaque pointer can be handled there.

btree/btree.c
```diff
@@ -1021,7 +1022,8 @@ struct node_type {
 	void* (*nt_opaque_get)(const struct segaddr *addr);
 
 	/** Captures node data in segment */
-	void (*nt_capture)(struct slot *slot, struct m0_be_tx *tx);
+	void (*nt_capture)(struct nd *node, struct slot *slot,
```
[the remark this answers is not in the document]
shashank-parulekar: Yes it will be present in slot but for a node routine nt_* it makes more sense to take node_descriptor as parameter for getting data within the node_descriptor instead of via the slot pointer. Also in future we might even remove the slot pointer as the node descriptor will contain enough information to indicate what needs to be captured within the node.

be/alloc.h
```diff
@@ -224,7 +224,28 @@ M0_INTERNAL void m0_be_allocator_credit(struct m0_be_allocator *a,
...
 
 /**
- * Allocate memory.
+ * Accumulate credits for optype in accum with no memory clear.
+ *
+ * @param a Allocator
+ * @param optype Allocator operation type
+ * @param size Size of allocation. Makes sense only for
+ *	       M0_IN(optype, (M0_BAO_ALLOC_ALIGNED, M0_BAO_ALLOC)).
+ *	       It is ignored for other optypes.
```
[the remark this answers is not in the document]
shashank-parulekar: Sorry I do not follow as I see the same tabs present in the other functions of this file.

be/alloc.h
```diff
@@ -259,7 +280,42 @@ M0_INTERNAL void m0_be_alloc_aligned(struct m0_be_allocator *a,
...
+ * @param tx Allocation will be done in this transaction
+ * @param op See m0_be_op.
+ * @param ptr Pointer to allocated memory will be stored to *ptr.
+ *	      *ptr shouldn't be used before completion of the operation.
+ *	      Can be NULL.
+ * @param size Memory size
+ * @param shift Memory will be aligned on (shift^2)-byte boundary.
+ *		It can be less than M0_BE_ALLOC_SHIFT_MIN - in this case
+ *		allocation will be done as if it is equal to
+ *		M0_BE_ALLOC_SHIFT_MIN.
```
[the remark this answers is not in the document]
shashank-parulekar: this one too.

be/alloc.h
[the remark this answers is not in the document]
ydb242: So as per our discussion on the documentation part, we did say that we use tabs only till the point we hit the first character. After the first character, we use spaces only.
In the diff that i see, i am seeing that after the '*' there are tabs till the next character.

image

Hence, I request you to convert these tabs from * to the next character to spaces.
Same comment for the rest of the places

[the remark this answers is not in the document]
ydb242: Yes, even i see that the use of tabs and spaces is not consistent in the file. However, at least for our changes we can try to follow one particular rule.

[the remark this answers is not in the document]
shashank-parulekar: > Yes, even i see that the use of tabs and spaces is not consistent in the file. However, at least for our changes we can try to follow one particular rule.

We cannot do this just for our changes as it does not go with the theme of the complete file. I will need to either maintain the current way the file is decorated or else change the whole file to stick to the right formatting and I do not want to do the later option so I will have to continue keeping the file format consistent.

btree/btree.c
RadhaGulhane13: > This will only take care of the opaque pointer capture for root nodes but it does not take care of the capture for the rest of the nodes allocated for this tree.
For rest of the nodes, currently we are capturring the opaque pointer when we are actually using that node. For example, We allocate new node for spliting operation, allocate -> initialize -> insert new records -> call captured for capturing node header and rest of the things.
But for root node, user allocate -> btree does node initialization -> capture node header.

RadhaGulhane13: However, I agree with your first point of capturing extra 8 bytes in some cases.

be/alloc.h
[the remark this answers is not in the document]
ydb242: Ok. I will resolve the comments.

ydb242 (approved):

kanchan-chaudhari (approved):

btree/btree.c
shashank-parulekar: > For rest of the nodes, currently we are capturring the opaque pointer when we are actually using that node. For example, We allocate new node for spliting operation, allocate -> initialize -> insert new records -> call captured for capturing node header and rest of the things. But for root node, user allocate -> btree does node initialization -> capture node header.

No we are not capturing opaque for the rest of the nodes. There was a bug in the code where the opaque pointer was never captured in the capture routines and that is because the opaque pointer would never be NULL. It would always contain pointer to the node descriptor by the time the capture routine is called. The opaque pointer happened to be NULL simply because the BE allocator sets all the bytes to NULL as a part of the allocation. With this new code where the BE allocator skips initializing the allocated memory we have non-NULL values in the opaque pointer and hence an explicit call to capture the opaque pointer is done.

RadhaGulhane13: Got the answer

RadhaGulhane13 (approved):

upendrapatwardhan (approved):

PR #1908, opened 2022-06-21
CORTX-31527 Return errno instead of -1 in timer_posix_init

JugalPatil:
Return errno instead of -1 in timer_posix_init
Signed-off-by: Jugal Patil <jugal.patil@seagate.com>


JugalPatil:
retest this please - rerunning vm-deployment check

PR #1986, opened 2022-07-16
CORTX-31893 : Optimize the node Capture for FF node format

upendrapatwardhan:
Optimize the node Capture for FF node format

# Problem Statement
- Problem statement

# Design
-  For Bug, Describe the fix here.
-  For Feature, Post the link for design

# Coding
   Checklist for Author

# Testing 
  Checklist for Author

# Impact Analysis

# Review Checklist 
  Checklist for Author
- [ ] PR is self reviewed

# Documentation
  Checklist for Author

### DIFF
--- a/btree/btree.c
+++ b/btree/btree.c
@@ -862,6 +862,35 @@ enum {
 enum {
 	CR_ALL = -1,
 	CR_NONE = -2,
+	/**
+	 * Flags to indicate which regions are to be captured in the node.
+	 *     Bits 0-1 : HDR flags
+	 *     Bits 2-3 : DIR flags
+	 *     Bits 4-5 : KV flags
+	 *     Bit    6 : Not used
+	 *     Bit    7 : Op flag to indicate ins/del operation
+	 * 
+	 * Value of flag bits:
+	 * 	0x00 - capture nothing.
+	 * 	0x01 - capture only modified part.
+	 * 	0x10 - Invalid value.
+	 * 	0x11 - capture full.
+	 */
+	CR_OP_DEL    = 0x80,
+	CR_HDR_SHIFT = 0,
+	CR_DIR_SHIFT = 2,
+	CR_KV_SHIFT  = 4,
+	CR_MOD       = 0x01,
+	CR_FULL      = 0x03,
+	CR_HDR_MOD   = CR_MOD  << CR_HDR_SHIFT,
+	CR_HDR_FULL  = CR_FULL << CR_HDR_SHIFT,
+	CR_HDR_MASK  = CR_HDR_FULL,
+	CR_DIR_MOD   = CR_MOD  << CR_DIR_SHIFT,
+	CR_DIR_FULL  = CR_FULL << CR_DIR_SHIFT,
+	CR_DIR_MASK  = CR_DIR_FULL,
+	CR_KV_MOD    = CR_MOD  << CR_KV_SHIFT,
+	CR_KV_FULL   = CR_FULL << CR_KV_SHIFT,
+	CR_KV_MASK   = CR_KV_FULL,
 };
 
 /** Direction of move in bnode_move(). */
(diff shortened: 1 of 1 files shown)


btree/btree.c
```diff
@@ -3132,52 +3191,135 @@ static void ff_done(struct slot *slot, bool modified)
...
+
+	if (h->ff_var.ff_level == 0 &&
+	    ff_crctype_get(tgt) == M0_BCT_BTREE_ENC_RAW_HASH)
+		crc_size = CRC_VALUE_SIZE;
+
+	if ((flags & CR_HDR_MASK) == CR_HDR_FULL) {
+		tmp.ncr_hdr_beg = h;
+		tmp.ncr_hdr_end = h->ff_opaque == NULL ?
+				  tmp.ncr_hdr_beg + sizeof(*h) :
+				  (void *)(&h->ff_opaque) - 1;
```
shashank-parulekar, line 3218: You can safely remove the check for h->ff_opaque == NULL  as it will never be NULL. So you can change the code to 
```suggestion
		tmp.ncr_hdr_end = tmp.ncr_hdr_beg + sizeof(*h);
```

btree/btree.c
```diff
@@ -3132,52 +3191,135 @@ static void ff_done(struct slot *slot, bool modified)
...
+	int                      crc_size = 0;
+
+	if (h->ff_var.ff_level == 0 &&
+	    ff_crctype_get(tgt) == M0_BCT_BTREE_ENC_RAW_HASH)
+		crc_size = CRC_VALUE_SIZE;
+
+	if ((flags & CR_HDR_MASK) == CR_HDR_FULL) {
+		tmp.ncr_hdr_beg = h;
+		tmp.ncr_hdr_end = h->ff_opaque == NULL ?
+				  tmp.ncr_hdr_beg + sizeof(*h) :
```
shashank-parulekar, line 3217: If ncr_hdr_end is supposed to contain the address of the last byte which should be persisted then you need to do sizeof(*h) - 1.

btree/btree.c
```diff
@@ -3132,52 +3191,135 @@ static void ff_done(struct slot *slot, bool modified)
...
+				  (void *)(&h->ff_opaque) - 1;
+	} else {
+		tmp.ncr_hdr_beg = &h->ff_var;
+		tmp.ncr_hdr_end = tmp.ncr_hdr_beg + sizeof(struct ff_hdr_var);
+	}
```
shashank-parulekar, line 3222: You are skipping capturing the footer. The footer should change because change to h->ff_var should cause a different CRC to be calculated.
In short any changes to ff_head should also cause the footer to be captured.

shashank-parulekar (changes requested):

btree/btree.c
upendrapatwardhan: Done.

btree/btree.c
upendrapatwardhan: Done.

btree/btree.c
upendrapatwardhan: Done.

kanchan-chaudhari (approved):
Looks good.

btree/btree.c
```diff
@@ -3213,12 +3213,12 @@ static void ff_cap_reg_define(struct nd *tgt, int idx, uint8_t flags)
...
-		tmp.ncr_hdr_end = h->ff_opaque == NULL ?
-				  tmp.ncr_hdr_beg + sizeof(*h) :
-				  (void *)(&h->ff_opaque) - 1;
+		tmp.ncr_hdr_end = (void *)(&h->ff_opaque);
 	} else {
+		/* Capture footer along with ff_var. */
 		tmp.ncr_hdr_beg = &h->ff_var;
-		tmp.ncr_hdr_end = tmp.ncr_hdr_beg + sizeof(struct ff_hdr_var);
+		tmp.ncr_hdr_end = tmp.ncr_hdr_beg + sizeof(h->ff_var) +
+				  sizeof(h->ff_foot);
```
shashank-parulekar, line 3221: This is a bit dangerous since you are assuming that there will be no variables between ff_var and ff_foot. In future this might not be true. A safer approach would be
```suggestion
		tmp.ncr_hdr_end = tmp.ncr_hdr_beg + offsetof(h->ff_var) +
				  offsetof(h->ff_foot) + sizeof(h->ff_foot);
```

I assume there is a macro/compiler keyword offsetof. Please confirm.

shashank-parulekar (changes requested):

btree/btree.c
upendrapatwardhan: This is by design. We plan to keep the frequently changing members together in ff_head.
If any new member is to be added which needs to be frequently captured, it should be added in struct ff_var.
Added a comment in the structure definition.

shashank-parulekar: I mean there could be a case where a non-frequently changing member could be listed between ff_var and ff_foot. There is no hard-coded rule which dictates that the ff_var should always be right before the ff_foot member. If that is supposed to be the case then please add a compile-time assert to make sure that ff_var is always right before ff_foot. You can add a compile-time assert such as the one below right after the location where struct ff_head is declared in the code.
M0_BASSERT(offsetof(struct ff_head, ff_var) + sizeof(((struct ff_head *)0)->ff_var) == offsetof(struct ff_head, ff_foot));

shashank-parulekar (approved):

PR #1992, opened 2022-07-19
dtm0/it: execute index operations in a single transaction

sergey-shilov:
This is the experimental change in all2all DTM0 integration test
that uses a combination of flags M0_ENF_META and M0_OIF_SKIP_LAYOUT
to execute index operations in a single transaction.

Problem:
Currently DTM0 is only applicable for ordinary DIX operations like
key/value PUT and DEL since these operations are executed in a
single local transaction. In case of index create/delete operations
we have two local transactions for layout operation and component
catalogues creation (with and without CROW). Since DTX is not applicable
for multiple local transactions we can not cover index operations with
DTM0 and recover such operations during DTM0 recovery process.

Solution:
Recently two flags of interest were introduced: M0_ENF_META and
M0_OIF_SKIP_LAYOUT. A combination of these two flags with disabled
CROW allows to skip layout operations and use a pool version for
the further ordinary operations stored after index create is executed.
That means we have a single local transaction to create component
catalogues only and DTX for index operations can be used in this case.
Recently the changes were introduced that allow to set these flags
from the command line for m0kv. Since m0kv is used as a client for
all2all DTM0 integration test now we can test these flags to ensure
that index operation is done in a single transaction and following
ordinary key/value operations work. So this is the first step to
start working on DTX support for index operations.

Changes:
In comparison to original all2all the changes in calling of m0kv are:
 * specify -L (enable M0_OIF_SKIP_LAYOUT) and -C (disable M0_OIF_CROW)
   options for all m0kv calls (index and ordinary operations)
 * specify -e (enable M0_ENF_META) m0kv option for index create operation,
   the pool version is printed out in this case and stored into bash variable
 * specify m0kv option -v <pool_version> that was stored after index create
   for key/value PUT operations

### DIFF
--- a/dtm0/it/all2all/all2all
+++ b/dtm0/it/all2all/all2all
@@ -346,7 +346,7 @@ function m0kv_async()
     args=$(python3 -c "import yaml; \
         conf = yaml.load(open('$M0CRATE_CFG'), Loader=yaml.FullLoader); \
         m = conf['MOTR_CONFIG']; \
-        print('-l %s -h %s -p %s -f %s' % (m['MOTR_LOCAL_ADDR'], m['MOTR_HA_ADDR'], m['PROF'], m['PROCESS_FID']));")
+        print('-L -C -l %s -h %s -p %s -f %s' % (m['MOTR_LOCAL_ADDR'], m['MOTR_HA_ADDR'], m['PROF'], m['PROCESS_FID']));")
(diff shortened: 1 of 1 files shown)


cortx-admin:

shailesh-vaidya:
Closing as an obsolete

PR #2003, opened 2022-07-21
cas/service: forbid operating on sdevs in OFFLINE state

sergey-shilov:
Problem:
Some time ago Hare was not updating devices state during
DTM0 recovery and they were still in OFFLINE state. In order
to process incoming REDO msgs and unblock working on DTM0
recovery we allowed operating on CAS OFFLINE devices, that
was a temporary workaround.
But logically it's not correct to operate with OFFLINE devices.

Solution:
Now Hare updates the state of devices that belong to the
recovering process. So that we can remove OFFLINE CAS device
state from allowed states for operating.


shashank-parulekar (approved):

huanghua78 (approved):
Looks good.

